PR #224, opened 2014-01-13
Implemented the image_path option

steveluscher:
The render methods now accept an `imagePath` option for use with the `image-url()` SASS function.


tapsboy:
Based on the recent developments at libsass https://github.com/sass/libsass/pull/834 & https://github.com/sass/libsass/issues/489, should this option be removed in subsequent releases?

am11:
@tapsboy yes! See https://github.com/sass/node-sass/issues/702.

Issue #566, opened 2014-12-09
Self-Signed Certificates and install.js


cuongvo:
Is it possible to add `rejectUnauthorized: false` to the options passed to request in [install.js](https://github.com/sass/node-sass/blob/master/scripts/install.js#L18)? Being behind a corporate firewall, our HTTPS traffic is proxied and certificates are self-signed, causing an exception.

PR #567, opened 2014-12-10
Added 'rejectUnauthorized' to support self-signed TLS certificates

cuongvo:
Per https://github.com/sass/node-sass/issues/566

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -15,7 +15,7 @@ var fs = require('fs'),
 
 function download(url, dest, cb) {
   var file = fs.createWriteStream(dest);
-  var options = { proxy: getProxy() };
+  var options = { proxy: getProxy(), rejectUnauthorized: false };
   var returnError = function(err) {
     fs.unlink(dest);
     cb(typeof err.message === 'string' ? err.message : err);
@@ -34,7 +34,7 @@ function download(url, dest, cb) {
 
   req.end();
   req.on('error', returnError);
-};
+}
 
 /**
  * Get proxy settings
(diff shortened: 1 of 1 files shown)


phosphore:
> At least it could be optional (eg. use an env var like for proxy settings).

Applying this insecure flag by default only to make optional [binary configuration parameters](https://github.com/sass/node-sass#binary-configuration-parameters) work is not ideal. Since the `getBinaryUrl` function was designed to fallback to `https://github.com/sass/node-sass/releases/download`, the certificate validation will be disabled when requesting the binaries from `github.com` even if the user is not using an alternative download path.

https://github.com/sass/node-sass/blob/5a4a48a4a45f41d66d55e2103496e886ec7ee5f1/lib/extensions.js#L242-L250 

> we are opting for most lenient settings, so our install scrip can grab libsass binary.
I understand this, but you should at least consider letting node-sass users know the risk in the README section.

saper:
Shall we revert this?

cuongvo:
Agreed this should be optional and clearly documented. This is a path for folks with issues with the install and are knowingly bypassing certificate verification. Normal users should have the secure path by default.

6 year ago me shouldn't have made this suggestion :)

phosphore:
> I don't have the context on the impact of reverting this change. It may be worth considering this more thoroughly. I can see this breaking users who depend on this being insecure for their build servers and releases in internal environments that will need to modify code.

Wouldn't it be enough to check if the user has set either:
https://github.com/sass/node-sass/blob/5a4a48a4a45f41d66d55e2103496e886ec7ee5f1/lib/extensions.js#L243-L246

If that's the case apply the `rejectUnauthorized`, else don't?

phosphore:
Any updates on this?

OrenSavichWS:
Hi, is there a fix for CVE-2020-24025?
if so, in what commit?

thanks in advance!

LiuJinghao:
Version 5.0.0 was released in August, but through reading the source code of 5.0.0, we found that this issue is still unresolved.

Is there a plan to fix this issue?

Issue #586, opened 2014-12-27
Defining importer function gives Abort trap: 6 error


browniefed:
By simply defining an importer function an Abort trap: 6 gets thrown whenever I attempt to run anything with an import.
```
@import "testimport";
.test {
  color: #000;
}
```

```js
var sass = require('node-sass'),
    path = require('path');

sass.render({
    file: path.resolve(__dirname, './test.scss'),
    outputStyle: 'compressed',
    success: function (css) {
        // the result string
        console.log(css.css);
    },
    error: function(err) {
        console.log(err);
    },
    importer: function(url) {
        return {file: url};
    }
});
```

mgreter:
@am11: First let me be clear: I really think the problem is not in libsass but with (the usage of) libuv. Next: I don't know shit about libuv but tried to hunt down your nasty race condition anyway. So here my uneducated guesses:

The error can also be avoided if this line in `sass_free_context_wrapper` gets commented out: `free(ctx_w);`. But I also don't see why this should error out (it seems valid to deallocate the memory here). Next observation was that commenting out `uv_async_init ...` in `binding.cpp` also "avoids" the crash. I don't know any details about libuv, but I think I know what it should be doing to get this working in an async manner. So I started to wonder where you actually "terminate" the "intermediate" thread. The google search on this lead me to the conclusion that you may miss some `uv_close`calls ...

I guess that's as much as I want to dive into the whole libuv stuff. But I'm pretty convinced that the cluprit is somewhere in there. But I also could be totally wrong here!

As a side note: You should be aware that `uv_work_t`, `uv_mutex_t`, `uv_cond_t` and `uv_async_t` are all direct members to `(sass_context_wrapper) ctx_w`. You seem to pass a lot of pointers to these structs (which is valid) to the internal functions of libuv; ie. `&ctx_w->async`. Once you free `ctx_w`, these pointers will become wild pointers!

am11:
@mgreter, thanks for looking into it.

You are right about `uv_close`, and this was the first thing which @txdv pointed out. I had that fixed with this commit: https://github.com/am11/node-sass/commit/35a82abc79dfd69b265da1751ed8d99e0359c4c9#diff-672b939c4bdf28c42fc7ba5da56105dbR240.

Now with the tip of my master branch, I generated two different sets of logs on Ubuntu:

Valgrind: https://gist.github.com/am11/6cfcebdef05e3751f16d#file-valgrind-log-L151
[...]

PR #589, opened 2014-12-28
Importer: General overhauling

am11:
* Fixes the async crashing by explicitly
  closing async.
* Changes the behavior of renderSync importer
* Updates index accordingly.
* Removes obsolete tests accordingly.

BIG THANKS to @txdv and this addresses #586 ! :tada::+1:

**Need to update the docs accordingly.**

---

Code: Adds NanScope() to guard nan handlers.
* Fixes for Persistent.
* Credit @kkoopa (thanks for pointing these out!).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -144,7 +144,6 @@ function getOptions(options) {
 
   var error = options.error;
   var success = options.success;
-  var importer = options.importer;
 
   options.error = function(err, code) {
     try {
@@ -172,23 +171,6 @@ function getOptions(options) {
       });
     }
   };
-
-  if (importer) {
-    options.importer = function(file, prev, key) {
-      var done = function(data) {
-        binding.importedCallback({
-          index: key,
-          objectLiteral: data
-        });
-      };
-
-      var result = importer(file, prev, done);
-
-      if (result) {
-        done(result);
-      }
-    };
-  }
 
   delete options.image_path;
   delete options.include_paths;
@@ -219,6 +201,25 @@ var binding = require(getBinding());
 module.exports.render = function(options) {
   options = getOptions(options);
 
+  var importer = options.importer;
+
+  if (importer) {
+    options.importer = function(file, prev, key) {
+      function done(data) {
+        binding.importedCallback({
+          index: key,
+          objectLiteral: data
+        });
+      }
+
+      var result = importer(file, prev, done);
+
+      if (result) {
+        done(result);
+      }
+    };
+  }
+
   options.data ? binding.render(options) : binding.renderFile(options);
 };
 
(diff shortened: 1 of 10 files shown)


lib/index.js
```diff
@@ -219,6 +201,25 @@ var binding = require(getBinding());
 module.exports.render = function(options) {
   options = getOptions(options);
 
+  var importer = options.importer;
+
+  if (importer) {
+    options.importer = function(file, prev, key) {
+      var done = function(data) {
```
kevva: `function done(data)`

am11:
@kevva, ignoring the fact that we are getting segmentation fault, there is another issue with cli importer tests.

So I separated this test case. With my master branch (and libsass submodule pointing to their current master), in node-sass clone root dir, run the following in bash:

```
node bin/node-sass /temp/node-sass/test/fixtures/include-files/index.scss --output /temp/node-sass/test/fixtures/include-files --importer /temp/node-sass/test/fixtures/extras/my_custom_importer_file_and_data_cb.js
```

It doesn't produces any output. But when I add `console.log('blah')` statement in lib/index.js#render() function, it generates the output and print on the screen. Could this be due to any dependency, emitter and so forth?

am11:
Turned out this CLI issue is ultimately related to binding. Only happening on Linux. On Win, it is passing all the tests.

Commenting out `free()` statement passes all tests on Linux and TravisCI too. Nothing new there.

Now, lets wait for libsass to release 3.1.0, and I will update the submodules again.

am11:
TravisCI says its all green! :palm_tree: 

Finally! yay!!
:tada:

mgreter:
:+1: Without known memory leaks now? If yes, what was the problem?

am11:
One memory concern (I don't know if it is really a leak): in case of [RenderSync](https://github.com/am11/node-sass/blob/fc1d4e6e72cbf0079c51d5c238ea4b8da0d0bff7/src/binding.cpp#L271-L298) or [RenderFileSync](https://github.com/am11/node-sass/blob/fc1d4e6e72cbf0079c51d5c238ea4b8da0d0bff7/src/binding.cpp#L319-L343), if I call `free(ctx_w)` it throws SIGSEGV. So I disposed everything else inside ctx_w and otherwise, except for the ctx_w itself. Why I don't think if it is a leak, because NanScope() macro seems to do some magic there and (it *feels* like -- not sure) it tries to dispose ctx_w. Maybe I am totally wrong and may be there still exist a way to free the ctx_w without segfaut (something like running `uv_walk` on all handles over and over until we are definitely sure all handles are released then free). At this point it seems pretty trivial. This was the major show stopper.

Then earlier I forgot to `uv_close` the `async`. Also, the invalid instruction exception I mentioned on sassc commit (happening occasionaly) was I think due to the reason that we were setting file_path in both file can data context, so I put the condition to prevent it from setting in case of file context (since we are already passing file path in `sass_make_file_context`). After that, I didn't get that in a while. Need to do some stress testings to be 100% sure:  like running `require("./").renderSync({data:"a{foo:1}"})` in `while(true)` loop etc.)

Since there is always room for improvement, for now I guess its fair enough.

I can still reproduce that CLI issue that I mentioned above, but @equalsraf via IRC told me that he is unable to reproduce it and TravisCI confirms that! (don't know if it is Ubuntu or just my bad luck!)

Thanks to everybody who helped us getting here. :)

am11:
(`while(true)` was a bad idea after all, had to terminate VM :stuck_out_tongue_winking_eye: )

PR #606, opened 2015-01-06
Add clarity to sourceMap option

jenius:
Very small PR that adds the fact that you must specify `outFile` as well in order to generate a source map.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-﻿# node-sass
+# node-sass
 
 <table>
   <tr>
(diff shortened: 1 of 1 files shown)


jenius:
The fact that you had to write that lengthy of a comment to explain what it meant is exactly my point. Here's how I might reword it:

> **sourceMap:** You must define this option as well as `outFile` in order to generate a source map. If it is set to `true`, the source map will be generated at the path provided in the `outFile` option. If set to a path (string), the source map will be generated at the provided path.

Does that seem correct? It makes more sense to me on initial reading.

am11:
Yeah it looks fine. Just to be consistent with rest of the text:

> ... If set to a path (\`String\`)...

jenius:
Ok just rebased to include the modified text. :tada:

am11:
Thanks @jenius! :+1:

jenius:
Whoo! :tada:

Issue #332, opened 2014-06-05
Custom Functions


mgreter:
Hi all

Just wanted to give you a heads up that it should be possible to implement custom functions again. Custom functions are defined inside the dynamic scripting language and then registered with libsass. Then they can be called/referenced from scss/sass files, similar to internal functions. 

Not sure if node-sass had that feature before the libsass refactoring. I fixed [some bits](https://github.com/hcatlin/libsass/pull/181) on libsass so the Perl Bindings, which already had custom functions, would work again. So maybe you can profit off of [that work](https://github.com/mgreter/CSS-Sass/blob/master/lib/CSS/Sass.xs#L215), since both implementations are bindings to a dynamic scripting language. Maybe you also want to expose [sass2scss](https://github.com/mgreter/CSS-Sass/blob/master/lib/CSS/Sass.xs#L344) to convert indented syntax to scss syntax.

There is a new [custom precision](https://github.com/hcatlin/libsass/issues/287) feature, which unfortunately has a regression. I also added the offical spec-tests to the [Perl-Binding test-suite](https://github.com/mgreter/CSS-Sass/blob/master/t/99-sass_specs.t), IMO it would be usefull to include them in node-sass too.

Kind regards
Marcel Greter

chriseppstein:
Hi, this is a critical feature to the extensibility of Sass as users are accustomed to in ruby. I am planning to port a large portion of Compass to node and we need this capability. I have an engineer at my work (linkedin) who is willing to work on this, but he are not accustomed to libsass or node.js's internals so we could really use some pointers about how to get started implementing this feature. As I understand it there are three pieces that are needed here:

* libsass <--> js function binding (in C++)
* Javascript implementation of Sass value objects.
* execution environment and helper functions (exposed as methods and properties of `this`)

Since I wrote a lot of the similar API for ruby, I'd be happy to help define or review the javascript interfaces.

xzyfer:
@mgreter is certainly the person to talk to in regards to the libsass side of things. We have some documentation on the libsass wiki for custom functions that should be of use https://github.com/sass/libsass/wiki/Custom-Functions-internal

chriseppstein:
@xzyfer Ok, I think I get it. So the string passed to `sass_make_function` is a fully qualified sass function argument declaration supporting defaults, etc?

xzyfer:
[...]

PR #615, opened 2015-01-10
 Importer: Drops uv_mutex in favor of std::mutex 

am11:
* Importer: Drops `uv_mutex` in favor of `std::mutex`. (0720fa8)
    * It is not safe to use `uv_cond_wait`
      outside of v8 thread (`uv loop`).

* Importer: Hack to make CLI importer usage work. (ff53926)
  * If anyone could ever explain this!
  * This is related to the changes made in 0720fa8.
  * **Tidbit:** it was also failing on windows with node.js v0.10.32
     updating to v0.10.35 (current stable), it worked without this hack on Windows
     but when tested with v0.10.35 on Ubuntu via nvm, it still fails without this hack.
     The only downside of this hack is extra nuisance on console
     (I tried to dodge it by overriding `console.log` function, but that didn't work either, so we would have to live with this *little annoyance*<b>**</b>, until someone can figure out why it is happening).

@browniefed, @xzyfer, can you guys please confirm if this patch makes custom importer useable on Mac (and addresses https://github.com/sass/node-sass/issues/586)? 

---
\*\* without the hack running the following in cli hangs indefinitely:

```bash
node node_module/.bin/node-sass /path/to/input.scss --importer /path/to/importer.js
```

with the hack, it executes properly, but prints whatever custom importer returns. Tested on Ubuntu and Windows.

**Note 1:** If the importer option is not set (in CLI or `render()` usage, leaving `renderSync()` alone), no unwanted / extra stuff will not be printed. 
**Note 2:** This means we would have to apply same hack to `custom functions` feature (#332), whenever we will implement it. :)
**Note 3:** We use `render` (as opposed to `renderSync`) in case of CLI usage, because we [route the errors to emitter](https://github.com/sass/node-sass/blob/ff539261cb2848b65cfc45843ac8cc97aca59eb7/lib/render.js#L86). In case of `renderSync`, the errors are *thrown* to stderr. <sup>[[1]](https://github.com/sass/node-sass/blob/ff539261cb2848b65cfc45843ac8cc97aca59eb7/src/binding.cpp#L295) &nbsp; [[2]](https://github.com/sass/node-sass/blob/ff539261cb2848b65cfc45843ac8cc97aca59eb7/src/binding.cpp#L340)</sup>

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -206,6 +206,7 @@ module.exports.render = function(options) {
   if (importer) {
     options.importer = function(file, prev, key) {
       function done(data) {
+        console.log(data); // ugly hack
         binding.importedCallback({
           index: key,
           objectLiteral: data
--- a/src/binding.cpp
+++ b/src/binding.cpp
@@ -81,12 +81,14 @@ struct Sass_Import** sass_importer(const char* file, const char* prev, void* coo
      *  can run uv_async_send without a push.
      */
 
+    std::unique_lock<std::mutex> lock(*ctx_w->importer_mutex);
+
     ctx_w->file = file ? strdup(file) : 0;
     ctx_w->prev = prev ? strdup(prev) : 0;
     ctx_w->async.data = (void*)ctx_w;
 
     uv_async_send(&ctx_w->async);
-    uv_cond_wait(&ctx_w->importer_condition_variable, &ctx_w->importer_mutex);
+    ctx_w->importer_condition_variable->wait(lock);
   }
   else {
     NanScope();
@@ -357,7 +359,7 @@ NAN_METHOD(ImportedCallback) {
   sass_context_wrapper* ctx_w = imports_collection[index];
 
   prepare_import_results(returned_value, ctx_w);
-  uv_cond_signal(&ctx_w->importer_condition_variable);
+  ctx_w->importer_condition_variable->notify_all();
 
   if (try_catch.HasCaught()) {
     node::FatalException(try_catch);
--- a/src/sass_context_wrapper.cpp
+++ b/src/sass_context_wrapper.cpp
@@ -24,8 +24,9 @@ extern "C" {
 
   sass_context_wrapper* sass_make_context_wrapper() {
     sass_context_wrapper* ctx_w = (sass_context_wrapper*)calloc(1, sizeof(sass_context_wrapper));
-    uv_mutex_init(&ctx_w->importer_mutex);
-    uv_cond_init(&ctx_w->importer_condition_variable);
+
+    ctx_w->importer_mutex = new std::mutex();
+    ctx_w->importer_condition_variable = new std::condition_variable();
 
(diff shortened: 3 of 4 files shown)


browniefed:
@am11 hot domain, checked out the std-mutex branch on your fork and it seems to work. Followed the same process on the #586 . Did around 30-40 copy and pastes of
```
require("./").render({
    data: "@import \"non-existing-file\"; \n.test {\ncolor: #000;\n}",
    outputStyle: 'compressed',
    success: function(result) {
        console.log(result);
    },
    error: function(error) {
        console.log(error);
    },
        done({contents: 'div { color: #000; }'});
    }
});
```
and they all seemed to work, no errors. Nice!

am11:
Yay! And thanks for testing. :tada: 
Could you please test `npm test` as well?

This will confirm that it's working on all accounts (render, renderSync and CLI usage). :)

browniefed:
@am11 ` 664 passing (5s)`, nothing seems to be failing :+1:

am11:
Awesomeness! So its a show time for v2! :)

Now we need to make the binaries. I have done the windows binaries (need to update [the PR](https://github.com/sass/node-sass-binaries/pull/28)). Could you please send a PR for Mac binary on [node-sass-binaries](https://github.com/sass/node-sass-binaries/) repo? Here is a good [example](https://github.com/sass/node-sass-binaries/pull/26).

I can walk you through the steps:

```bash
# after this is merged, clone node-sass ... start fresh
git clone https://github.com/sass/node-sass --recursive; cd node-sass
git submodule update --init --recursive; npm install
node-gyp rebuild        # without -d, 
                        # so it should create a folder build/Release (with binding.node)

# on github, fork https://github.com/sass/node-sass-binaries
# then clone your fork, perferablly outside node-sass directory:
cd ..; git clone htttps://github.com/browniefed/node-sass-binaries; cd node-sass-binaries

ls -lhS

rm darwin-x64/*   # just so you can 
cp ../node-sass/bin/Release/binding.node darwin-x64/

ls -lhS   # see the difference in file sizes, so you know you are using latest one,
          # its lame I know just for fun sakes :)

git add .;git commit -am "Binary: Adds Mac binaries for v2.0.0."; git push
```

And finally the PR. :)

Thanks in anticipation. :zap: :100:

am11:
Merged! ;)

browniefed:
Thanks for the instructions only one small modification `node-sass/build` instead of `node-sass/bin` :100:

callumlocke:
Curious, does anyone know why the console.log hack works? And does it actually need to print out the whole data object, or could it just print out an empty line?

am11:
@callumlocke, it was explained by @matryo here: https://github.com/sass/node-sass/pull/644#discussion_r23903134. It will be removed when custom functions will arrive to node-sass.

davidgtonge:
The console.log hack is rendering this project unusable

PR #616, opened 2015-01-10
Install: Use npmconf to get proxy config

am11:
//cc @kevva

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "mkdirp": "^0.5.0",
     "mocha": "^2.0.1",
     "nan": "^1.3.0",
+    "npmconf": "^2.1.1",
     "object-assign": "^2.0.0",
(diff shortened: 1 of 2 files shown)


am11:
@kevva, so does it LGTY? :)

kevva:
:+1:

am11:
@kevva is `@api` a legal jsdoc block type (identifier)? I could not find that one on http://usejsdoc.org/.

On that note, I think we need `@returns` across the board. :)

kevva:
@am11, I think it's http://usejsdoc.org/tags-access.html but I've always used `@api` for some reason.

am11:
Perhaps we can submit a proposal somewhere for `@access [private / public / protected / internal]` :)

Or even better:  `@scope [private / public / protected / internal]`

Issue #157, opened 2013-09-26
Imports don't seem to get recompiled


achalddave:
If I have a file `index.scss` which has `@import "common.scss"`, and change `common.scss`, `index.scss` doesn't seem to get recompiled.

The relevant code seems to be in the following locations:

`checkImports` is called [here](https://github.com/andrew/node-sass/blob/a253ed1f945e15b3a1a830a2df561b5915c4f26d/lib/middleware.js#L152), and defined [here](https://github.com/andrew/node-sass/blob/a253ed1f945e15b3a1a830a2df561b5915c4f26d/lib/middleware.js#L178). 

It relies on `imports[path]`, which only seems to be assigned [here](https://github.com/andrew/node-sass/blob/a253ed1f945e15b3a1a830a2df561b5915c4f26d/lib/middleware.js#L117) to an empty list. I can't seem to find any code that detects imported files -- am I missing something?

am11:
`npm install node-sass@2.0.0` is released!
Checkout the release notes: https://github.com/sass/node-sass/releases/tag/v2.0.0

jhnns:
Yay! :checkered_flag:

Issue #590, opened 2014-12-29
[FR] output file dependencies for incremental compilation support


amitport:
if given a sass file, node-sass could output the files that it depends on (imports - optionally recursive) then it would be relatively easy to implement incremental compilation via tools like [gulp](https://github.com/gulpjs/gulp)+[gulp-changed](https://github.com/sindresorhus/gulp-changed)

Any chance that node-sass will support such an function?

am11:
Hi @amitport, take a look at https://github.com/xzyfer/sass-graph! It does what you are looking for. :)

amitport:
@am11 Thanks! I think your previous suggestion was already good enough for this purpose, but I'm not sure. Generally, having the result after only a full node-sass compilation is OK, since I'm doing an incremental compilation not just analysis. In any case, I haven't got around to making this experiment yet, hopefully, I'll have a chance next weekend.

xzyfer:
You can use sass-graph in your gulp watcher or gulp changed. It'll give you
an array of sass files that your changed file in imported by. Then you can
just pass those files into your normal gulp compilation task.

amitport:
@xzyfer, but don't already know which files I need to watch from `stats.includedFiles`? remember I run an initial compilation. Can't I just keep a mapping (processed.stats.includesFiles)->sourceFile?

xzyfer:
I think it might be a little more difficult that it seems. FWIW here is my current gulp setup for incremental sass compilation https://gist.github.com/xzyfer/1225f52c370b511c1fed

PR #629, opened 2015-01-15
Use sass-graph for accurate sass watching

xzyfer:
## Problem 

The watcher is broken. So very broken. Two major problems with it

- it compiles files starting with underscores (`_`) which is a big no-no
- it has no awareness of the the import graph

Assuming the following directory structure

```
src
|--foo.scss
|--bar.scss
|-- components
   |-- _baz.scss
```

where 'foo.scss` and `bar.scss`, and running the following command `node-sass -w -r --output dist src`.

A modification to `src/components/_baz.scss` **should not** compile `_baz.scss`. It **should** traverse the import graph and compile `dist/foo.css` and `dist/bar.css`.

This is how the native sass watcher works.

## Solution

Use [sass-graph](https://github.com/xzyfer/sass-graph) to build an in memory graph of the `@imports` to determine the correct compilation targets for a modified file.

I believe this is the issue at the heart of https://github.com/sass/node-sass/issues/157.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -5,6 +5,7 @@ var Emitter = require('events').EventEmitter,
     meow = require('meow'),
     replaceExt = require('replace-ext'),
     stdin = require('get-stdin'),
+    grapher = require('sass-graph'),
     render = require('../lib/render');
 
 /**
@@ -183,10 +184,20 @@ function watch(options, emitter) {
   gaze.add(dir);
   gaze.on('error', emitter.emit.bind(emitter, 'error'));
 
+  var graph = grapher.parseDir(options.src, { loadPaths: options.includePath });
+
   gaze.on('changed', function(file) {
-    options = getOptions([file], options);
-    emitter.emit('warn', '=> changed: ' + file);
-    render(options, emitter);
+    var files = [file];
+    graph.visitAncestors(file, function(parent) {
+      files.push(parent);
+    });
+
+    files.forEach(function(file) {
+      if (path.basename(file)[0] === '_') return;
+      options = getOptions([path.resolve(file)], options);
+      emitter.emit('warn', '=> changed: ' + file);
+      render(options, emitter);
+    });
   });
 }
 
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "object-assign": "^2.0.0",
     "replace-ext": "0.0.1",
     "request": "^2.48.0",
+    "sass-graph": "^1.0.1",
     "shelljs": "^0.3.0"
   },
   "devDependencies": {


am11:
This is super awesome. :+1: 

Considering this request: https://github.com/sass/node-sass/issues/590,  I didn't knew something like sass-graph existed! :smile:  I think node-sass should not expose this functionality (or expose the show-graph feature) to its API. That would be against the laws of DRY (especially now where have package referenced here).

xzyfer:
> I didn't knew something like sass-graph existed 

It didn't really til today...

browniefed:
@xzyfer any idea how it will play with custom importers in 2.0?
I can imagine that people will write some compass stuff to process Sprites, or other arbitrary things that aren't true file paths.

xzyfer:
> I think node-sass should not expose this functionality (or expose the show-graph feature) to its API.

I've isolated this functionality explicitly to the watch function in the cli. The graph doesn't even exist until you being watch so there shouldn't be any leakage.

am11:
oOo.. Did you release it today for the first time? I saw old PRs in your repo, thought it out there for a while.. :)

xzyfer:
@browniefed sass-graph currently follows the native sass model. It'll walk imports, if the file exists on disk in an include path it's be added to an in memory graph. That graph is then used to find the collection or top-most ancestors which should be re-compiled.

ATM it has no knowledge of custom importers.

browniefed:
@xzyfer I guess I was implying it'd be cool if it could accept an importer, likely the same one that would be used for node-sass, so both things could pair up in terms of custom importer logic.

am11:
Can we take the `*cookie` approach here as well, to save the `require('path/to/my-importer.js')` object in graph? 

Or if it is not possible to retain the JS object, then perhaps function-to-string approach: `require('path/to/my-importer.js').toString()` and eval it before usage.

xzyfer:
@am11 It's been in various stages of development for a while, but I released 1.0.0 a couple hours ago in preparation for getting it into node-sass for my dev team :)

xzyfer:
Custom importers may very well be possible in future versions.

am11:
On second thought, we may only want to store the path to custom importer in graph and then re-require it in watch even handler. (whichever approach is elegant).

Yes the importers can certainly wait, since you might like to consider providing room for the future "custom functions" as well (which will behave the same way, except there will by a multiplicity factor -- spoiler alert!!). :)

xzyfer:
aha! CI is passing :D :tada:

test/cli.js
```diff
@@ -139,6 +139,8 @@ describe('cli', function() {
 
       bin.stderr.setEncoding('utf8');
       bin.stderr.on('data', function(data) {
+        console.log('=> changed: ' + src);
+        console.log(data);
```
am11: This can be rebased out. :D

xzyfer: done in a35326a

am11:
:tada: 

Feel free to merge it whenever you think this is ready.

Maybe we can set sass-graph to `^1.0.1` in package.json.

xzyfer:
> Maybe we can set sass-graph to ^1.0.1 in package.json.

I generally give npm the benefit of the doubt but better safe than sorry. I've update this. When all the builds pass I'll give it a run on my local Libsass 3.2.0 fork then ship it.

xzyfer:
Works amazingly! Are those AppVeyor failures expected?

am11:
Yup! :)

xzyfer:
:tada: :rocket:

am11:
Yay! :)

Issue #530, opened 2014-11-10
Provide api for custom importers


jhnns:
With sass/libsass#21 libsass will provide the possibility to specify a custom importer. This is useful for build tools like [webpack](https://github.com/webpack/webpack) that provide a custom resolving algorithm.

We basically just have to wrap the js-callback so it can be called by libsass:

``` javascript
sass.render({
    ...
    importer: function (filename) {
        // rewrite filename
        return filename;
    }
});
```

The libsass-api is a bit more complex to support more advanced use-cases as described [here](https://github.com/sass/libsass/issues/21#issuecomment-62318381). I'm planing to make the js-api a bit more pleasant to work with.

For example you could do:

``` javascript
    importer: function (filename) {
        return filename;
    }
```

``` javascript
    importer: function (filename) {
        return {
            filename: "..."
            filesource: "..."
        };
    }
```

``` javascript
    importer: function (filename) {
        return [
            {
                filename: "..."
                filesource: "..."
            },
            {
                filename: "..."
                filesource: "..."
            }
        ];
    }
```

One problem I currently see is that if a `filesource` is provided, this needs to be done **synchronously** which might be impossible in certain situations.

What do you think about it? If it's ok for you I could start a pull-request.

PR #644, opened 2015-01-31
Custom functions

matryo:
This is a refactor of #530 (custom importers) to also support #332 (custom functions). 

The mechanism to synchronize the worker thread that runs libsass and the main v8 loop has been centralized and abstracted into an object called `CallbackBridge` that has two specializations: `CustomImporterBridge` and `CustomFunctionBridge`. The former supports the existing functionality (#530) while the other makes it possible to register custom functions that are backed up by v8 functions and takes care of translating `Sass_Value`s to something that makes sense to the JS runtime and vice versa.

The function arguments are relayed to the JS callback by means of wrapping `Sass_Value*` pointers into v8 objects. Those objects mimic the API offered by libsass and as such could be a bit cumbersome to work with on the javascript side. This problem could however be largely mitigated by translating those wrapper objects into JS primitives upon entering the JS callback and back into wrapper objects upon return. The only edge cases that I can think of is that of `SASS_NUMBER` values when they carry units and `SASS_COLOR` where we could parse strings and detect patterns like `\d+(px|em|%)` and `rgba?\(...)\` or resolve to use more JS friendly facades.

This is a quick example demonstrating how compass' `headings` function may be implemented:

```javascript
sass.render({
  data: '#{headings(2,5)} { color: #08c; }',
  functions: {
    'headings($from: 0, $to: 6)': function(from, to) {
      var i, f = from.getValue(), t = to.getValue(),
          list = new sass.types.List(t - f + 1);

      for (i = f; i <= t; i++) {
        list.setValue(i - f, new sass.types.String('h' + i));
      }

      return list;
    }
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,68 @@ When returning or calling `done()` with `{ contents: "String" }`, the string val
 
 `this` refers to a contextual scope for the immediate run of `sass.render` or `sass.renderSync`
 
+### functions
+`functions` is an `Object` that holds a collection of custom functions that may be invoked by the sass files being compiled. They may take zero or more input parameters and must return a value either synchronously (`return ...;`) or asynchronously (`done();`). Those parameters will be instances of one of the constructors contained in the `require('node-sass').types` hash. The return value must be of one of these types as well. See the list of available types below:
+
+#### types.Number(value [, unit = ""])
+* `getValue()`/ `setValue(value)` : gets / sets the numerical portion of the number
+* `getUnit()` / `setUnit(unit)` : gets / sets the unit portion of the number
+
+#### types.String(value)
+* `getValue()` / `setValue(value)` : gets / sets the enclosed string
+
+#### types.Color(r, g, b [, a = 1.0]) or types.Color(argb)
+* `getR()` / `setR(value)` : red component (integer from `0` to `255`)
+* `getG()` / `setG(value)` : green component (integer from `0` to `255`)
+* `getB()` / `setB(value)` : blue component (integer from `0` to `255`)
+* `getA()` / `setA(value)` : alpha component (number from `0` to `1.0`)
+
+Example:
+
+```javascript
+var Color = require('node-sass').types.Color,
+  c1 = new Color(255, 0, 0),
+  c2 = new Color(0xff0088cc);
+```
+
+#### types.Boolean(value)
+* `getValue()` : gets the enclosed boolean
+* `types.Boolean.TRUE` : Singleton instance of `types.Boolean` that holds "true"
+* `types.Boolean.FALSE` : Singleton instance of `types.Boolean` that holds "false"
+
+#### types.List(length [, commaSeparator = true])
+* `getValue(index)` / `setValue(index, value)` : `value` must itself be an instance of one of the constructors in `sass.types`.
+* `getSeparator()` / `setSeparator(isComma)` : whether to use commas as a separator
+* `getLength()`
+
+#### types.Map(length)
+* `getKey(index)` / `setKey(index, value)`
+* `getValue(index)` / `setValue(index, value)`
+* `getLength()`
+
+#### types.Null()
+* `types.Null.NULL` : Singleton instance of `types.Null`.
+
+#### Example
+
+```javascript
+sass.renderSync({
+  data: '#{headings(2,5)} { color: #08c; }',
+  functions: {
+    'headings($from: 0, $to: 6)': function(from, to) {
+      var i, f = from.getValue(), t = to.getValue(),
+          list = new sass.types.List(t - f + 1);
+
+      for (i = f; i <= t; i++) {
+        list.setValue(i - f, new sass.types.String('h' + i));
+      }
+
+      return list;
+    }
+  }
+});
+```
+
 ### includePaths
 Type: `Array<String>`
 Default: `[]`
--- a/binding.gyp
+++ b/binding.gyp
@@ -4,7 +4,19 @@
       'target_name': 'binding',
       'sources': [
         'src/binding.cpp',
-        'src/sass_context_wrapper.cpp'
+        'src/create_string.cpp',
+        'src/custom_function_bridge.cpp',
+        'src/custom_importer_bridge.cpp',
+        'src/sass_context_wrapper.cpp',
+        'src/sass_types/boolean.cpp',
+        'src/sass_types/color.cpp',
+        'src/sass_types/error.cpp',
+        'src/sass_types/factory.cpp',
+        'src/sass_types/list.cpp',
+        'src/sass_types/map.cpp',
+        'src/sass_types/null.cpp',
+        'src/sass_types/number.cpp',
+        'src/sass_types/string.cpp'
       ],
       'include_dirs': [
         '<!(node -e "require(\'nan\')")',
@@ -15,31 +27,31 @@
               'libsass.gyp:libsass',
             ]
         }],
-	['libsass_ext == "auto"', {
-	  'cflags_cc': [
-	    '<!(pkg-config --cflags libsass)',
-	  ],
-	  'link_settings': {
-	    'ldflags': [
-	      '<!(pkg-config --libs-only-other --libs-only-L libsass)',
-	    ],
-	    'libraries': [
-	      '<!(pkg-config --libs-only-l libsass)',
-	    ],
-	  }
-	}],
-	['libsass_ext == "yes"', {
-	  'cflags_cc': [
-	    '<(libsass_cflags)',
-	  ],
-	  'link_settings': {
-	    'ldflags': [
-	      '<(libsass_ldflags)',
-	    ],
-	    'libraries': [
-	      '<(libsass_library)',
-	    ],
-	  }
+        ['libsass_ext == "auto"', {
+          'cflags_cc': [
+            '<!(pkg-config --cflags libsass)',
+          ],
+          'link_settings': {
+            'ldflags': [
+              '<!(pkg-config --libs-only-other --libs-only-L libsass)',
+            ],
+            'libraries': [
+              '<!(pkg-config --libs-only-l libsass)',
+            ],
+          }
+        }],
+        ['libsass_ext == "yes"', {
+          'cflags_cc': [
+            '<(libsass_cflags)',
+          ],
+          'link_settings': {
+            'ldflags': [
+              '<(libsass_ldflags)',
+            ],
+            'libraries': [
+              '<(libsass_library)',
+            ],
+          }
         }],
         ['OS=="mac"', {
           'xcode_settings': {
@@ -47,9 +59,7 @@
               '-std=c++11',
               '-stdlib=libc++'
             ],
-            'OTHER_LDFLAGS': [
-              '-stdlib=libc++'
-            ],
+            'OTHER_LDFLAGS': [],
             'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
             'GCC_ENABLE_CPP_RTTI': 'YES',
             'MACOSX_DEPLOYMENT_TARGET': '10.7'
@@ -73,6 +83,7 @@
         }],
         ['OS!="win"', {
           'cflags_cc+': [
+            '-fexceptions',
             '-std=c++0x'
           ]
         }]
--- a/lib/index.js
+++ b/lib/index.js
@@ -137,6 +137,67 @@ function getOptions(options, cb) {
 }
 
 /**
+ * Executes a callback and transforms any exception raised into a sass error
+ *
+ * @param {Function} callback
+ * @param {Array} arguments
+ * @api private
+ */
+
+function tryCallback(callback, args) {
+  try {
+    return callback.apply(this, args);
+  } catch (e) {
+    if (typeof e === 'string') {
+      return new binding.types.Error(e);
+    } else if (e instanceof Error) {
+      return new binding.types.Error(e.message);
+    } else {
+      return new binding.types.Error('An unexpected error occurred');
+    }
+  }
+}
+
+/**
+ * Normalizes the signature of custom functions to make it possible to just supply the
+ * function name and have the signature default to `fn(...)`. The callback is adjusted
+ * to transform the input sass list into discrete arguments.
+ *
+ * @param {String} signature
+ * @param {Function} callback
+ * @return {Object}
+ * @api private
+ */
+
+function normalizeFunctionSignature(signature, callback) {
+  if (!/^\*|@warn|@error|@debug|\w+\(.*\)$/.test(signature)) {
+    if (!/\w+/.test(signature)) {
+      throw new Error('Invalid function signature format "' + signature + '"');
+    }
+
+    return {
+      signature: signature + '(...)',
+      callback: function() {
+        var args = Array.prototype.slice.call(arguments),
+            list = args.shift(),
+            i;
+
+        for (i = list.getLength() - 1; i >= 0; i--) {
+          args.unshift(list.getValue(i));
+        }
+
+        return callback.apply(this, args);
+      }
+    };
+  }
+
+  return {
+    signature: signature, 
+    callback: callback
+  };
+}
+
+/**
  * Render
  *
  * @param {Object} options
@@ -172,13 +233,9 @@ module.exports.render = function(options, cb) {
   var importer = options.importer;
 
   if (importer) {
-    options.importer = function(file, prev, key) {
+    options.importer = function(file, prev, bridge) {
       function done(data) {
-        console.log(data); // ugly hack
-        binding.importedCallback({
-          index: key,
-          objectLiteral: data
-        });
+        bridge.success(data);
       }
 
       var result = importer.call(options.context, file, prev, done);
@@ -187,6 +244,31 @@ module.exports.render = function(options, cb) {
         done(result);
       }
     };
+  }
+
+  var functions = options.functions;
+
+  if (functions) {
+    options.functions = {};
+
+    Object.keys(functions).forEach(function(signature) {
+      var cb = normalizeFunctionSignature(signature, functions[signature]);
+
+      options.functions[cb.signature] = function() {
+        var args = Array.prototype.slice.call(arguments),
+            bridge = args.pop();
+
+        function done(data) {
+          bridge.success(data);
+        }
+
+        var result = tryCallback(cb.callback, args.concat(done));
+
+        if (result) {
+          done(result);
+        }
+      };
+    });
   }
 
   options.data ? binding.render(options) : binding.renderFile(options);
@@ -206,8 +288,22 @@ module.exports.renderSync = function(options) {
 
   if (importer) {
     options.importer = function(file, prev) {
-      return { objectLiteral: importer.call(options.context, file, prev) };
+      return importer.call(options.context, file, prev);
     };
+  }
+
+  var functions = options.functions;
+
+  if (options.functions) {
+    options.functions = {};
+
+    Object.keys(functions).forEach(function(signature) {
+      var cb = normalizeFunctionSignature(signature, functions[signature]);
+
+      options.functions[cb.signature] = function() {
+        return tryCallback(cb.callback, arguments);
+      };
+    });
   }
 
   var status = options.data ? binding.renderSync(options) : binding.renderFileSync(options);
@@ -228,3 +324,12 @@ module.exports.renderSync = function(options) {
  */
 
 module.exports.info = process.sass.versionInfo;
+
+/**
+ * Expose sass types
+ */
+
+module.exports.types = binding.types;
+module.exports.TRUE = binding.types.Boolean.TRUE;
+module.exports.FALSE = binding.types.Boolean.FALSE;
+module.exports.NULL = binding.types.Null.NULL;
--- a/src/create_string.cpp
+++ b/src/create_string.cpp
@@ -1,0 +1,15 @@
+#include <nan.h>
+#include <string.h>
+#include "create_string.h"
+
+
+char* create_string(Local<Value> value) {
+  if (value->IsNull() || !value->IsString()) {
+    return 0;
+  }
+
+  String::Utf8Value string(value);
+  char *str = (char *)malloc(string.length() + 1);
+  strcpy(str, *string);
+  return str;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ matrix:
 before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   - sudo apt-get update
-  - sudo apt-get install gcc-4.8 g++-4.8
-  - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 20
-  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 20
+  - sudo apt-get install gcc-4.7 g++-4.7
+  - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.7 20
+  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.7 20
   - g++ --version
   - sudo apt-get update -qq
   - git submodule update --init --recursive
--- a/libsass.gyp
+++ b/libsass.gyp
@@ -61,14 +61,12 @@
               '-std=c++11',
               '-stdlib=libc++'
             ],
-            'OTHER_LDFLAGS': [
-              '-stdlib=libc++'
-            ],
+            'OTHER_LDFLAGS': [],
             'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
             'GCC_ENABLE_CPP_RTTI': 'YES',
             'MACOSX_DEPLOYMENT_TARGET': '10.7'
           }
-         }],
+        }],
         ['OS=="win"', {
           'msvs_settings': {
             'VCCLCompilerTool': {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "coverage": "node scripts/coverage.js",
     "install": "node scripts/install.js",
     "postinstall": "node scripts/build.js",
-    "pretest": "node_modules/.bin/jshint bin lib test",
+    "pretest": "node_modules/.bin/jshint bin lib scripts test",
     "test": "node_modules/.bin/mocha test"
   },
   "files": [
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -53,7 +53,7 @@ function afterBuild(options) {
  */
 
 function build(options) {
-  var arguments = [
+  var args = [
     path.join('node_modules', 'pangyp', 'bin', 'node-gyp'),
     'rebuild',
   ].concat(
@@ -63,9 +63,9 @@ function build(options) {
     })
   ).concat(options.args);
 
-  console.log(['Building:', process.sass.runtime.execPath].concat(arguments).join(' '));
+  console.log(['Building:', process.sass.runtime.execPath].concat(args).join(' '));
 
-  var proc = spawn(process.sass.runtime.execPath, arguments, {
+  var proc = spawn(process.sass.runtime.execPath, args, {
     stdio: [0, 1, 2]
   });
 
--- a/src/binding.cpp
+++ b/src/binding.cpp
@@ -1,112 +1,39 @@
 #include <nan.h>
 #include <vector>
 #include "sass_context_wrapper.h"
-
-char* create_string(Local<Value> value) {
-  if (value->IsNull() || !value->IsString()) {
-    return 0;
-  }
-
-  String::Utf8Value string(value);
-  char *str = (char *)malloc(string.length() + 1);
-  strcpy(str, *string);
-  return str;
-}
-
-std::vector<sass_context_wrapper*> imports_collection;
-
-void prepare_import_results(Local<Value> returned_value, sass_context_wrapper* ctx_w) {
-  NanScope();
-
-  if (returned_value->IsArray()) {
-    Handle<Array> array = Handle<Array>::Cast(returned_value);
-
-    ctx_w->imports = sass_make_import_list(array->Length());
-
-    for (size_t i = 0; i < array->Length(); ++i) {
-      Local<Value> value = array->Get(static_cast<uint32_t>(i));
-
-      if (!value->IsObject())
-        continue;
-
-      Local<Object> object = Local<Object>::Cast(value);
-      char* path = create_string(object->Get(NanNew<String>("file")));
-      char* contents = create_string(object->Get(NanNew<String>("contents")));
-
-      ctx_w->imports[i] = sass_make_import_entry(path, (!contents || contents[0] == '\0') ? 0 : strdup(contents), 0);
-    }
-  }
-  else if (returned_value->IsObject()) {
-    ctx_w->imports = sass_make_import_list(1);
-    Local<Object> object = Local<Object>::Cast(returned_value);
-    char* path = create_string(object->Get(NanNew<String>("file")));
-    char* contents = create_string(object->Get(NanNew<String>("contents")));
-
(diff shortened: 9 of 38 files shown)


am11:
@matryo, first of all thanks for the effort.  :+1: 

Welcome to GitHub! Wow you joined today and your first PR is HUGE! I would have split it into more commits so to make it easy to manage. 

You are aiming to fix https://github.com/sass/node-sass/issues/332, which is in our 2.1 milestone. Therefore, we will consider this after v2.0.0 is shipped. I would have discussed it before going through this kind of major change.

Regarding the implementation, we decided to implement this format:

```javascript

sass.render({
  data: '#{headings(2,5)} { color: #08c; }',
  functions: {   // the idea is to make these definitions least complicated

    headings: function (from, to) {
      return process(from, to);
    },

    anotherHeadings: function (from, to, done) {
      done(process(from, to));
    }

  }
});
```

which is achievable if you keep the custom functions definition in C++ keeping JS usage syntax flexible (the way we did for custom importer).

Thirdly, I am not sure about the new Sass types you added (`new sass.types.String('h' + i))`). Are these arbitrary types? In general we do not add additional syntaxes in node-sass beyond what libsass has to offer. Especially not to solely meet the requirement of a separate framework such as Compass, which is beyond the scope of this project.

@mgreter, @xzyfer, any thoughts on this?

mgreter:
Looks pretty ok to me :+1: Also the new types are pretty much expected from my side, since they represent the values possbile in libsass (https://github.com/sass/libsass/wiki/API-Sass-Value). I personally also implemented/overloaded "toString" for each type, which could solve some "issues" @matryo mentioned when handling such types in node (maybe take a look at how I did it for [`perl-libsass`] [1]).

I also like the config part by @matryo better, since the snippet @am11 posted is not taking advantage of the "possibilities" libsass offers here. There is no way to define default values for arguments, if you do not explicitly include the function signature somewhere. And I also do it that way in `perl-libsass`!

[1]: https://github.com/sass/perl-libsass/blob/master/lib/CSS/Sass/Type.pm

matryo:
@am1 I understand that this PR's approach is pretty unusual and I am really sorry about that. I did not realize any one had already started a draft in that direction. Just to give you some context this was a major road block for us and I needed to get this completed in a couple of days, which unfortunately did not leave me much opportunity to consult the maintainers team. That said I am totally eager to take all the time necessary to make sure this code is in line with what you guys envisioned for 2.1 if you think this is a solid base for #332.

am11:
@mgreter, now that you have described it, the syntax looks pretty OK to me too. :smile: 

Would it make sense to have both syntaxes for function descriptor? Mostly the function definition lack default values. So, when we have this *special* need for default parameter:

```javascript
'headings($from: 0, $to: 6)': function (from, to) {
  return processIt(from, to);
},
```

otherwise:

```javascript
headings: function (from, to) {
  return processIt(from, to);
},
```

@matryo, we may need to refine it and this will take some time to digest the whole PR.

binding.gyp
```diff
@@ -5,6 +5,10 @@
       'sources': [
         'src/binding.cpp',
         'src/sass_context_wrapper.cpp',
+        'src/custom_function_bridge.cpp',
+        'src/custom_importer_bridge.cpp',
+        'src/create_string.cpp',
+        'src/sass_types.cpp',
```
am11: CodeStyler: This is a sorted list.

lib/index.js
```diff
@@ -204,13 +207,9 @@ module.exports.render = function(options) {
   var importer = options.importer;
 
   if (importer) {
-    options.importer = function(file, prev, key) {
+    options.importer = function(file, prev, bridge) {
       function done(data) {
-        console.log(data); // ugly hack
```
am11: This is history attached to it. Without this, it fails with CLI usage and manually calling render with importer. If you have find a way to fix it. Kudos! But please make sure of it.

lib/index.js
```diff
@@ -221,6 +220,29 @@ module.exports.render = function(options) {
...
+        }
+
+        var result = callback.apply(this, args.concat(done));
+
+        if (result) {
+          done(result);
+        }
+      };
+    });
+  }
```
am11: If possible, we need to reuse the same logic for CLI usage. Just note for self. Let me know if / whenever you are adding the CLI support (before that). We need to adapt and enhance on the way jscs, jshint and other linters support custom importers. I will then explain it. :)

lib/index.js
```diff
@@ -238,7 +260,7 @@ module.exports.renderSync = function(options) {
 
   if (importer) {
     options.importer = function(file, prev) {
-      return { objectLiteral: importer(file, prev) };
+      return importer(file, prev);
```
am11: This is awesome! :+1:

src/create_string.cpp
```diff
@@ -0,0 +1,15 @@
+#include <nan.h>
+#include <string.h>
+#include "create_string.h"
```
am11: This is good refactoring! :+1:

README.md
```diff
@@ -92,6 +92,66 @@ The callback function is passed a results object, containing the following keys:
...
+
+      for (i = f; i <= t; i++) {
+        list.setValue(i - f, new sass.types.String('h' + i));
+      }
+
+      return list;
+    }
+  }
+});
+```
```
am11: I think it is time to maintain a Wiki for node-sass. Lets refactor the README as well.

mgreter:
Btw. I find it pretty cool that @matryo was able to pull off this implementation without asking anybody about the internals directly! I hope that the time I've put into the libsass wiki was worth it! @matryo : If you have anything to add to the Wiki, please do so!!

am11:
I agree. This proves two things: 1) libsass documentation is very reliable and 2) @matryo is an excellent C++ developer!

src/sass_types.h
```diff
@@ -0,0 +1,225 @@
...
+      static char const* get_constructor_name() { return "SassMap"; }
+      static Sass_Value* construct(const Arguments&);
+
+      static void initPrototype(Handle<ObjectTemplate> proto) {
+        proto->Set(NanNew("getLength"), FunctionTemplate::New(GetLength)->GetFunction());
+        proto->Set(NanNew("getKey"), FunctionTemplate::New(GetKey)->GetFunction());
+        proto->Set(NanNew("setKey"), FunctionTemplate::New(SetKey)->GetFunction());
+        proto->Set(NanNew("getValue"), FunctionTemplate::New(GetValue)->GetFunction());
+        proto->Set(NanNew("setValue"), FunctionTemplate::New(SetValue)->GetFunction());
+      }
```
am11: Wouldn't it make sense to take `initPrototype` and get_constructorName definitions to `cpp` as well?

Same applies to `CoreValue` class.

lib/index.js
matryo: That got me thinking that we probably don't have to wrap the importer callback at all anymore.

src/sass_types.h
```diff
@@ -0,0 +1,225 @@
...
+
+      static NAN_METHOD(GetValue);
+      static NAN_METHOD(SetValue);
+  };
+
+
+  class Color : public CoreValue<Color> {
+    public:
+      Color(Sass_Value* v) : CoreValue(v) {}
+      static char const* get_constructor_name() { return "SassColor"; }
```
am11: Since we are using C++11, you may want to declare it as:

```c++
constexpr char* constructor_name = "SassBoolean";
```

How about modifying *the must implemented by derived class* members with `virtual` inheritance?

Also, you may want to take a look at http://stackoverflow.com/a/16296215/863980.

src/sass_types.h
matryo: Absolutely. My rationale for leaving them in the header was better readability of what is made available to the JS runtime but moving them to the `cpp` makes a lot of sense as well.

lib/index.js
matryo: Oops, I must have removed this thinking this was part of my debug logs. But now I'm intrigued, is there an issue ticket that explains why this hack was introduced?

src/sass_types.h
```diff
@@ -0,0 +1,225 @@
...
+      static Value* unwrap(Handle<v8::Value>);
+  };
+
+
+  // Include this in any SassTypes::Value subclasses to handle all the heavy lifting of constructing JS
+  // objects and wrapping sass values inside them
+  template <class T>
+  class CoreValue : public Value {
+    public:
+      static char const* get_constructor_name() { return "SassValue"; }
```
am11: Since we are using C++11, you may want to declare it as:

```c++
constexpr char* constructor_name = "SassValue";
```

How about modifying *the must implemented by derived class* members with `virtual` inheritance?

Also, you may want to take a look at http://stackoverflow.com/a/16296215/863980.

lib/index.js
am11: This is true. We can remove this block from here (we would probably need to keep it in renderSync).

lib/index.js
am11: Yes here: https://github.com/sass/node-sass/pull/615

> If anyone could ever explain this!

I was also intrigued about it. Took me sometime to *devise* this random hack. :D

It could have fixed in this overhaul. I am still very interested to know about what was causing it. :)

lib/index.js
am11: Note: `render` (the `async` variant) allows both callback as well as immediate return.

src/sass_types.h
matryo: I did not know about this `constexpr` keyword, looks interesting, thanks for the pointers! You are right we probably don't need a function here. As for virtual inheritance, that was a no-go because the constructor name is used from `get_constructor()` which is static.

src/sass_types.h
am11: You have already made the code much readable than it was before. IMO with this, it will make the code discoverability much easy for future contributors.

My vote is for even making code/implementation files separate for each derived class. :)

lib/index.js
matryo: Ok, I see. Well I am glad to announce that yes this indeed got fixed along with my overhaul :D which I guess is a good thing since the previous hack would not play nice with the `--stdout` option of the cli.

The reason this was happening is because of missed wakeup, i.e. the `notify()` call executed before the `wait()` call, leaving `wait()` hang indefinitely. To work around this, `wait()` has to be guarded by a loop that verifies whether the condition that we are waiting for may have already been met. Because that's a very common pattern `wait()` can actually receive an optional predicate as its second argument that does just that. That's how I used it:

https://github.com/matryo/node-sass/commit/8fdf5c4a85e1d4b135c8e97b7348c826c9ae3a4b#diff-953483bd71622742606ddb26e23f1e62R103

am11: Brilliant work! and thanks for explaining it!! :100: 

We were using `uv_mutex` and it was later we switched to std mutex. I had no idea that we can check already semaphored cond var in predicate.  :D

Everyday learning new things on GitHub.

matryo:
I pushed a few commits that should cover all of the recommendations made thus far, namely:

* Removed the importer wrap in `renderSync()`
* Reorganized `binding.gyp`
* Split sass types into separate modules
* Use of `constexpr` static variables in `CoreValue` subclasses in lieu of a static function
* Added support for potentially omitting function signature, used @mgreter's suggestion of using `fn(...)`

Also made some improvements on the following points:
* Made the sass types' constructors more flexible so they can be called without the `new` keyword
* Errors not caught by custom functions are now properly bubbled up to libsass by returning `SASS_ERROR` values.

As far as the API doc for libsass is concerned, I concur it is really nice! Unlike a lot of API docs it does not beat around the bush and goes straight to the point, I found it very useful during this exercise. So great job!

am11:
Excellent job there @matryo! And thanks for adding the alternative syntax . I will test it and give you further feedback. By it looks, it seems very promising as is. :)

README.md
chriseppstein: First off, thank you for working on this. It's super important functionality.

The property setters in these Sass value objects are very worrisome to me. Sass values are immutable. If this makes it possible to mutate the Sass values created from a sass file then, it's a violation of the language's design. If it doesn't then this makes it seem like you can, and would make it hard for them to debug what is going on when it doesn't work. My preference would be to make these objects completely immutable, but barring that, I think the objects passed into the function from the sass compiler should be marked as "readonly" and have any attempt to use the setters result in an error. After looking into the code more closely it seems that you're just binding libsass value objects to javascript here so this critique may be more towards libsass than the javascript binding layer. (ping @akhleung)

A few more things that we've done in the ruby api (can totally be added after this lands):

* `null`, `true` and `false` are always the same (immutable) object.
* All objects can be cast to a boolean to test if they are truthy or falsey. Only `null` and `false` ever return false.
* These objects should have a `toSass()` or `toString()` that returns a string representation for debugging.
* Maps should have `nth($index)` function that returns the key/value pair at the nth index. If they are mutable, setters, inserts, deletes by index make sense too.
* Numbers should have a comparator that does the css unit conversions if need be.
* All Sass values should support an equality check that returns true iff the sass `==` operation would return true.

chriseppstein: Also, the `Map` type should have a `keys` method. How else can you iterate over the map?

chriseppstein:
@am11

> Thirdly, I am not sure about the new Sass types you added (new sass.types.String('h' + i))). Are these arbitrary types? In general we do not add additional syntaxes in node-sass beyond what libsass has to offer. Especially not to solely meet the requirement of a separate framework such as Compass, which is beyond the scope of this project.

New types are very much in the scope of node-sass. Without them, how would you return a sass value? Some js types could be auto-boxed into Sass values but many can not. In ruby sass, these types are part of the core sass project. E.g. https://github.com/sass/sass/blob/stable/lib/sass/script/value/number.rb

am11:
> Without them, how would you return a sass value? 

I thought these will return cstrings from our binding and LibSass will take care of the rest. Unlike ruby-sass, node-sass is only a wrapper around libsass. Besides, I do not know about the use-case when the custom functions are required to return anything but the stringly typed value, which LibSass will eventually stamp out to resultant CSS as is?

README.md
matryo: @chriseppstein This is great input! You are right, the `SassTypes::Value` objects exposed to the v8 runtime constitute a very thin binding layer to the corresponding `Sass_Value` objects from `libsass` and as such expose all the same setters.

That said, custom functions are invoked with their arguments passed by value (i.e. `SassTypes::Value`’s constructor performs a deep copy of the underlying `Sass_Value` object via `sass_clone_value`) and similarly the results are returned by value (see `SassTypes::Value::get_sass_value()`). As a consequence it would not be possible for a custom function to mutate sass values created from a sass file. This immutable virtue is thus preserved for all values from the perspective of the sass files.

I understand that Sass as a language only exposes immutable values and as such it would make a lot of sense to hold onto this philosophy when it comes to custom functions especially if you guys think this would relieve ambiguities potentially perceived by the end developer.

However, If I can offer my two cents as a developer who spends a good deal of time working with JS: this is by no means a necessity. The fact that Sass’ design does not favor mutable values does not have to bleed onto the conventions established in Javascript. While I can totally make sense of immutable strings and numbers, I believe immutable maps and lists would feel awkward to a lot of JS developers out there. Additionally making these complex types immutable may lead to inferior performance results (and I said “may” because I have not looked at what’s behind the public libsass API, so I couldn't affirm that for sure).

In any case, I believe you are absolutely right in pushing for more user-friendly objects. And while it would make sense to me to leave all the subclasses of `SassTypes::Value` strictly as a translation layer, I can totally see a variety of facade objects in the Javascript realm that would expose all the goodies that you mentioned.

As far as the `Map` type is concerned, you would iterate over the map the same way you would if you had to use `libsass` directly:

```javascript
function(myMap) {
	var i, l, key, value;

	for (i = 0, l = myMap.getLength(); i < l; i++) {
		key = myMap.getKey(i);
		value = myMap.getValue(i);
	}
}
```

matryo:
@am11 From what I understood of the API, custom functions registered with `sass_option_set_c_functions()` are expected to return a pointer to a `Sass_Value` struct, which has to be of the adequate type. Meaning if the function advertises that it returns a color, it should return a `Sass_Value` that was created with `sass_make_color()`. If it were to return a string (i.e. a `Sass_Value` that was created with `sass_make_string()`), the return value could not be re-injected into a function that expects a color as its argument (e.g. `lighten()`) making our custom functions hardly composable, i.e. it would be impossible to do something like: 

```sass
div {
    color: lighten(foo(), 30%);
}
```

...if foo returned a string.

As I said earlier we could however let the end developer return strings and have a post-processing stage (either in `index.js` or in binding land) that detects patterns (length literals, color literals, etc.) and automagically transform them into the adequate type.

am11:
@matryo, that makes sense. :)
I was wondering if we can make custom functions implementer underlying sass-type agnostic, unless it is utterly necessary (which it is), so to avoid learning curve overhead. 

> that detects patterns (length literals, color literals, etc.) and automagically transform them into the adequate type.

I think that won't be necessary. With the proper documentation and tests cases to exemplify the usage (as you have already added both :+1: ), the implementers of node-sass API shall be able to define custom functions without much of a hassle. Since the main advantage of node-sass is performance; I am afraid pattern sniffing with regex will have a negative affect on perf ?

chriseppstein:
As one of the heaviest users of the Ruby function api that is provided by Sass, this patch is exactly what I was expecting. returning javascript objects that are converted to `Sass_Value`s is exactly what any developer coming from the Ruby environment would have expected.

mgreter:
I can only support @chriseppstein words! This is exactly as we did it in `perl-libsass`. As others and I already have mentioned, there is some room for further improvements to i.e. make maps (objects), list (arrays) and other types (strings) feel more native in a js environment.
And this patch is IMO a good starting point for that!

chriseppstein:
A few other things I was expecting to see here:

1. Sass options

I would expect `this.options` to be an object containing the current sass compilation options. Additionally, it should contain some runtime values like the current css output file and the current sass file and the top level sass file. These are critical to writing functions that can interact with the filesystem and handle URL generation correctly.

2. Sass environment

In ruby, you can access the current selector context I would expect `this.environment.selector` to be accessible as long as libsass has the script-parent feature.

You can also access the current variables defined in the global scope `this.globalVariables` is essential to being able to re-use sass configuration in the javascript language just like a sass-defined function can.

3. Calling to other Sass functions

It is very useful to be able to invoke sass builtin functions from within a custom function. It's also nice to be able to invoke functions defined in sass. For example, the `global-variable-exists` function that is a built-in in Sass should be able to tell us if some global is defined. I would expect `this.environment.call('global-variable-exists', sass.types.String("foo"))` or something similar to work.

Obviously what we have is a good first step, and I wouldn't block landing this PR because these features are missing (Ruby Sass didn't have them at first either)

am11:
@chriseppstein, (sorry for this getting stretched and GH has zero comment threading/nesting) here:

> As one of the heaviest users of the Ruby function api that is provided by Sass, this patch is exactly what I was expecting. returning javascript objects that are converted to  Sass_Value s is exactly what any developer coming from the Ruby environment would have expected.

do you mean return value's type of custom function should be specified by user as @matryo's solution (`sass.types.blah`) or should the type be programatically inferred from returned regular JS object by our API or binding code?

matryo:
@chriseppstein You raised three very interesting points. At the moment it unfortunately looks like libsass' public API cannot provide what we would need to implement something like this. After looking at libsass' internals I can see that there would be a rather straight forward way to implement those features, but that would be using a private API, which is a bit questionable I guess. 

I will take some time this week to augment the JS binding objects so as to provide a more user friendly API.

xzyfer:
@matryo feel free to PR Libsass with the additional information you require. I'm sure @mgreter will happily assist you with anything you need.

chriseppstein:
@am11  

> do you mean return value's type of custom function should be specified by user as @matryo's solution (`sass.types.blah`) or should the type be programatically inferred from returned regular JS object by our API or binding code?

I meant that it should be `sass.types.Blah`.

chriseppstein:
FYI: I've built a project that depends on this. It makes it trivial to use npm to download sass extensions containing sass files and custom sass functions written in javascript.

https://github.com/sass-eyeglass/eyeglass

I am waiting for this patch to land and be released before I can release this project.

am11:
@matryo, v2.0.0 (and then v2.0.1) is out. Please rebase this so we can continue.

```bash
cd into/node-sass

# first add the remote for original repo namely node-sass
git remote add node-sass https://github.com/sass/node-sass

# rebase with node-sass master branch
git pull --rebase node-sass master

# solve the merge conflicts and then:
git add .
git rebase --continue

# force push; to rewrite the history
git push -f
```

chriseppstein:
We noticed that importers seem to only be supported for async rendering. How should we support them for `renderSync()`?

chriseppstein:
The functions signature allows for async functions, but it seems that in order for a function to support `renderSync()` we'd have to have a synchronous version for each function.

It seems to me that `renderSync` is evil, but if we're going to support it we probably need to allow options to have a `functions` property that expects async compatible functions and a `functionsSync` property that specifies functions to be used in synchronous mode.

At the very least it seems like we should give an error if an async function is passed to renderSync.

am11:
Some notes regarding `renderSync`, we initially added support for sync callbacks in `renderSync` importers. It worked for most part unless the snag was found; libuv does not let you live with it and presents unanticipated issues. I spent quite sometime on libuv channel to figure that one out. @txdv helped me a lot in building importers and suggested not to improvise asynchrony with `renderSync` (which is pure synchronous on all accounts).

Currently, importers allow both async cb and synchronous `return` for `render` and only `return` for `renderSync`.

am11:
@matryo, ping.

chriseppstein:
@am11 Thanks for the background on renderSync. If someone is writing functions for use by a third-party, they won't know if that function is going to be passed to `render` or `renderSync` so they need to make two versions of their function, yes? Is there no way in synchronous invocation to handle async functions so that users could write just one version?

Jakobo:
@chriseppstein aside from a node-fibers solution not really. 

Two possible ideas:
1. With promises coming perhaps we can deprecate `renderSync`
2. If we need `renderSync` perhaps we can use the options proposal to tell the importer / custom function the intended rendering mode (and allow async-only functions to pass through gracefully and warn)

matryo:
@chriseppstein If third party libraries want to be sync-friendly they would have to test for the presence of the `done` callback as the second argument to decide on the best course of action (or as @Jakobo suggested using the option context from #681)

@am11 I unfortunately did not finish the improved type bindings although I made some good progress last week. I will definitely take some time during the next couple of days to wrap this up.

chriseppstein:
> test for the presence of the `done` callback.

This makes sense, thanks!

Jakobo:
@matryo is there a place I can pull your branch in progress so I can test the #681 changes or should I wait for changes to land in your `custom_functions` branch?

am11:
@matryo, please take your time. This is one of the major most feature coming node-sass, since the inception of node-sass. I was just pinging to know if there is anything we can help with. But seems like everything is under control! :smiley:  :+1: 

@Jakobo,

We used to have `renderSync`, `renderFileSync`, `render` and `renderFile`. Those were squashed to `render` and `renderSync`. We now detect file / data context based on the provided options.

It is a separate discussion that how the API will look in future, but having a sync variant around is better IMO. 

Regarding promises, apparently that pattern might not land as folks are in agreement to adapt node-style callbacks over promises or reactive-extensions. See the related issue.  

Moreover, you can consider your PR decoupled from this one and we can proceed. We will make sure all existing tests pass on merging this one. Meanwhile, we can continue our development flow; improving other aspects of node-sass.

@chriseppstein, `render` is the most popular usage of node-sass API. If you are using `render`, you can consume third parameter (the `callback`) in importer function like @matryo mentioned as well as you can `return` immediately. In case of `renderSync`, importer can only `return` as it is not equipped to accept cb by design.

mgreter:
I can only repeat that it's a pitty that `perl-libsass` has all these nice features already implemented, while most people are waiting for `node-sass` to catch up! So a little tease to hopefully close https://github.com/sass/node-sass/issues/332. And be sure I will push the `node-sass`API to the limit once it is merged. I want to see even more js integration (as `perl-libsass` already has) for like compare or native coerce `toString` in the future! :wink: :+1:

matryo:
Hi guys, I realize it's been quite some time since my last update. Something unexpected came up in my schedule. I am essentially in the same state as last time, except I should actually be able to free up some time during the next few days.

I also just wanted to throw an idea out there. It all goes back to the discoverability aspect of custom functions and the confusion that those `SassTypes ` could occasion for some developers. Looking back at  @am11 's comment on Feb 3 about having custom functions return strings that would then be interpolated into the right type. This grew on me quite a bit as I think many developers would be naturally tempted to do this rather than instantiating one of the `SassTypes` object, especially for simple literals such as strings, colors, numbers. And this wouldn't be too difficult to implement as we could just ask libsass to evaluate whatever string we get back from a custom function (if the function does not already return a `SassTypes::Value` already of course). I actually built a quick proof of concept, if you guys are interested. Let me know.

Lastly @chriseppstein, I did not get a chance to catch you at EmberConf, but that was a very nice closing talk! And eyeglass definitely looks like something I want to use :wink:

am11:
@matryo, that will be perfect! This way, for the node-sass API implementer, we will provide both signature and signature-less functions as well as typed and type-less returns and callbacks!  :+1: 

I am bit busy with build related stuff at the moment, but I will keep an eye on your commits.

chriseppstein:
@matryo I think this string idea is of dubious value and it's something we can add later without breaking any compatability. My suggestion is to wait on it and just land the important stuff now and then iterate.

chriseppstein:
@matryo eyeglass depends on this patch landing in 3.0, so that's why I'm hoping to land this sooner than later :)

am11:
@matryo, v3.0.0-pre is released. We would like to merge this feature for beta.
I agree with @chriseppstein that we can always iterate over once the basic functionality has arrived. Also, this way we will get sometime to act upon the feedback by the consumers (beta-testers).
If you are still busy, would you like us to take it from here?

saper:
I think this one is a blocker for 3.0

matryo:
@chriseppstein I think you're right, string interpretation is not critical and we can go piecemeal on this. 

For the purpose of 3.0 do you guys think that what we have so far is good? The only thing I would personally suggest to add is @chriseppstein 's recommendation of mapping `true`, `false` and `null` to immutable singleton instances. Adding this later on could indeed break users' code. I have a commit ready for this, so just let me know.

am11:
@matryo, normally the workflow that follow is: we commit the code and then ask for the review 
In worse case scenario, if one commit turned out to be a fiasco in a PR, you can always [rebase it out](http://www.git-scm.com/book/en/v2/Git-Branching-Rebasing) while keeping the others.

I will take care of CLI, which (on second thought) is not urgent for v3 either. IMO, advanced user relying on functionality like custom function, don't normally rely on CLI for their workflow. So CLI support (which is already a spaghetti) can wait until after v3 is RTW'd.

chriseppstein:
ok. so let's land the singleton instances and merge this baby. This is getting time critical for me at work for an internal release -- I may need to do a custom build to our internal npm repo if we can't land this soon. :sob:

saper:
I have some issue compiling this with clang an node-0.12:

https://gist.github.com/saper/304a1750c0e822e29083

first error:

```
gmake: Entering directory `/home/saper/sw/test/node_modules/node-sass/build'
clang++ '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' '-DBUILDING_NODE_EXTENSION' -I/home/saper/.node-gyp/0.12.0/src -I/home/saper/.node-gyp/0.12.0/deps/uv/include -I/home/saper/.node-gyp/0.12.0/deps/v8/include -I../node_modules/nan  -fPIC -pthread -Wall -Wextra -Wno-unused-parameter -m64 -O3 -ffunction-sections -fdata-sections -fno-omit-frame-pointer -fno-rtti -std=c++0x -fexceptions -frtti -MMD -MF ./Release/.deps/Release/obj.target/binding/src/binding.o.d.raw  -c -o Release/obj.target/binding/src/binding.o ../src/binding.cpp
In file included from ../src/binding.cpp:3:
In file included from ../src/sass_context_wrapper.h:8:
In file included from ../src/custom_function_bridge.h:5:
../src/callback_bridge.h:154:57: error: cannot initialize a parameter of type 'v8::Isolate *'
      with an lvalue of type 'void (const v8::FunctionCallbackInfo<v8::Value> &)'
    Local<FunctionTemplate> tpl = FunctionTemplate::New(New);
 ```

**Comment by am11** [comment]
@matryo, at this point, could you just rebase and merge the conflicts?

@chriseppstein, we will **definitely** land it into `v3.0.0-beta.0`. :sunglasses: 
Either we have custom functions or there would be no beta! :zap:

**Comment by chriseppstein** [comment]
When I return an error from a custom js function I get the following output:

```
error in C function eyeglass-asset-uri: Asset not found: fake.png
Backtrace:
	sass/assets.scss:2, in function `eyeglass-asset-uri`
	sass/assets.scss:2, in function `asset-url`
	stdin:1
```

Which seems bad, because it's not a C function. I assume this is libsass? Maybe we can get them to generalize their error message?

mgreter:
@chriseppstein we would need more info to debug this. I'm not sure if the message or the backtrace is the concern. IMO this is how libsass currently reports errors that are "thrown" in custom c functions. I actually have started to implement error reporting like ruby sass does, but that whole part is a lot of work in itself (and will not really be doable, before we have error spec tests in libsass).

OK, by re-reading it I guess the `C` should be replaced by a `custom` or whatever seems fit, correct? This is due to the implementation, as custom function are in fact always C functions (that's how the implementor calls libsass). Not sure how we could improve that other than just calling them `custom`, but I'm open for suggestions!

chriseppstein:
@mgreter Right. It's just that it's called a `C` function. If the user wrote a javascript or perl function, this error message seems kind of silly, but I'm sure it wouldn't trip them up too much. I think `custom` functions might be confused with the functions declared with `@function`. Perhaps `native` function or `host language` function. Or libsass could let the function registration tell it what kind of function it is.

matryo:
Ok, I am in the middle of rebasing that stuff. Let me tell you it's not the most fun I've had ;)

Right now I am running into an issue where the build script can't build `Release/sass.a`. I am hoping upgrading XCode will solve it, I will keep you posted!

am11:
@matryo I have just finished rebasing it. ;)
Let me push some stuff.

am11:
@matryo, except for binding.cpp, the rest is in accord with node.js v0.10, v0.12 and io.js v1-v1.5. https://github.com/am11/node-sass/commit/ae70faeeb710f9a2a120f25de94e8c12d8eed414

Since this PR is started, there have been a lot of development in node.js and nan worlds. So let keep collaborating and get this thing done. I am available on gitter chat: https://gitter.im/sass/node-sass.

am11:
You can just take the changes from this commit, basically its all your changes, I have just fixed some stuff which is new in Nan (and yet v8).

src/sass_types.h
am11: FTR,
Sorry I had to revert back `constexpr` usage to your original function style owing to the reason that Visual Studio is yet to bring it to the table: http://blogs.msdn.com/b/vcblog/archive/2014/11/17/c-11-14-17-features-in-vs-2015-preview.aspx.

am11:
@matryo see #783.

am11:
:tada: 

It is official now!! :sunny: :star2: :star: :spades: :100: 

Brilliant work @matryo :+1:

callumlocke:
👍👍👍 awesome

am11:
@matryo, do we need Boolean and Null exported twice: separately as types `exports.FALSE`, `exports.TRUE`, `exports.NULL` and then `exports.types.Boolean.TRUE`, `exports.types.Boolean.FALSE`, `exports.types.NULL`?

chriseppstein:
:heart: You guys rock my socks off.

Snugug:
I second his :heart:

Jakobo:
:tada: Awesome! :heart:

matryo:
@am11 They indeed do not necessarily have to be on both and we could refactor the code to have those singleton references exported just once. It just felt easier to instantiate those singletons within `Boolean::get_constructor` and `Null::get_constructor` and attach them to the constructor object, and copy those references to `module.exports` in a second pass for convenience. I guess we could also just leave it as is and just not advertise the `exports.types.Boolean.TRUE`, `exports.types.Boolean.FALSE`, `exports.types.NULL` references.

PR #650, opened 2015-02-05
Don't exit if an error occurs while watching

dprotaso:
If the user specified to watch a directory/file and an error occurs we shouldn't stop the process.


xzyfer:
:+1:

Issue #655, opened 2015-02-09
RFC: Binary Coverage Matrix


am11:
As @kkoopa mentioned at https://github.com/sass/node-sass/issues/627#issuecomment-70419098:

> A module must be rebuilt whenever  `NODE_MODULE_VERSION` changes.  It changed from joyent/node 0.10 -> joyent/node 0.11, from joyent/node 0.11 -> iojs/io.js 1.0, and atom/atom-shell has its own fork going with yet another `NODE_MODULE_VERSION`.

we need to confine a binary coverage matrix for the most used versions of runtimes (related: https://github.com/sass/node-sass/issues/602#issuecomment-73408256).

### Updated:

The current version of node-sass is <del>v2.0.0-beta</del> v2.0.1 and the distributable binary availability of node-sass is given below:

| Availability | Runtime | Version <br /> `NODE_MODULE_VERSION` | Platform | Architecture |
| :---: | :----: | :----: | :----: | :----: |
|  | atom-shell | 0.21 | Linux | ia32 |
|  | atom-shell | 0.21 | Linux | x64 |
|  | atom-shell | 0.21 | Mac | x64 |
|  | atom-shell | 0.21 | Windows | ia32 |
|  | atom-shell | 0.21 | Windows | x64 |
|  | io.js | 1.2 | FreeBSD | ia32 |
|  | io.js | 1.2 | FreeBSD | x64 |
| x | io.js | 1.2 | Linux | ia32 |
| x | io.js | 1.2 | Linux | x64 |
| x | io.js | 1.2 | Mac | x64 |
|  | io.js | 1.2 | Solaris | ia32 |
|  | io.js | 1.2 | Solaris | x64 |
| x | io.js | 1.2 | Windows | ia32 |
| x | io.js | 1.2 | Windows | x64 |
|  | node.js | 0.12 | FreeBSD | ia32 |
|  | node.js | 0.12 | FreeBSD | x64 |
| x | node.js | 0.12 | Linux | ia32 |
| x | node.js | 0.12 | Linux | x64 |
| x | node.js | 0.12 | Mac | x64 |
|  | node.js | 0.12 | Solaris | ia32 |
|  | node.js | 0.12 | Solaris | x64 |
| x | node.js | 0.12 | Windows | ia32 |
| x | node.js | 0.12 | Windows | x64 |
|  | node.js | 0.10 | FreeBSD | ia32 |
|  | node.js | 0.10 | FreeBSD | x64 |
[...]

PR #657, opened 2015-02-10
Build: Produces identifiable binary name

am11:
The new format is:
`{platform}-{arch}-{runtime}-{major.minor}`
where `major` and `minor` version belong to
the runtime.

Related Issue: #655.

//cc @xzyfer

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,5 +1,6 @@
 var fs = require('fs'),
-    path = require('path');
+    path = require('path'),
+    utils = require('./utils');
 
 /**
  * Get binding
@@ -8,11 +9,10 @@ var fs = require('fs'),
  */
 
 function getBinding() {
-  var name = process.platform + '-' + process.arch;
   var candidates = [
     path.join(__dirname, '..', 'build', 'Release', 'binding.node'),
     path.join(__dirname, '..', 'build', 'Debug', 'binding.node'),
-    path.join(__dirname, '..', 'vendor', name, 'binding.node')
+    path.join(__dirname, '..', 'vendor', utils.getBinaryIdentifiableName(), 'binding.node')
   ];
 
   var candidate = candidates.filter(fs.existsSync)[0];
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Get unique name of binary for current platform
+ *
+ * @api public
+ */
+
+module.exports.getBinaryIdentifiableName = function() {
+  return process.platform + '-' +
+         process.arch + '-' +
+         runtime + '-' +
```
xzyfer: Looks like you factored out `runtime`

xzyfer:
Will this approach handle things that aren't node or io, like atom shell?

lib/utils.js
am11: LOL yes and one more thing. Repushing now :smiley:

am11:
@xzyfer, I have not yet looked into atom shell yet. We can incorporate it later. However, we can add binary for io.js in node-sass-binaries v2.0.0.

xzyfer:
Yep! It's definitely a step forward

lib/index.js
```diff
@@ -8,11 +9,10 @@ var fs = require('fs'),
  */
 
 function getBinding() {
-  var name = process.platform + '-' + process.arch;
   var candidates = [
     path.join(__dirname, '..', 'build', 'Release', 'binding.node'),
     path.join(__dirname, '..', 'build', 'Debug', 'binding.node'),
```
am11: @xzyfer, @kevva ,
I think Debug should take precedence over Release? because if you explicitly build Debug binary, then your primary intent is to debug that binary in debug mode. :)

kevva: y

am11: > because if you explicitly build Debug binary with  node-gyp rebuild -d , then your primary intent is to debug that binary in debug mode

am11:
@xzyfer, tested with atom shell, it returns `win32-ia32-atom-1.0`. :smirk:

xzyfer:
Nailed it :+1:

PR #658, opened 2015-02-10
Include correct header for malloc related functions.

ahacking:
The C malloc() functions require inclusion of ``stdlib.h``

This fixes compilation on FreeBSD.

### DIFF
--- a/src/sass_context_wrapper.h
+++ b/src/sass_context_wrapper.h
@@ -1,3 +1,4 @@
+#include <stdlib.h>
 #include <nan.h>
 #include <condition_variable>
 #include "libsass/sass_context.h"


src/binding.cpp
```diff
@@ -1,3 +1,4 @@
+#include <stdlib.h>
```
am11: I think this would not be necessary, if you add it once in `src/sass_context_wrapper.h`.

src/sass_context_wrapper.cpp
```diff
@@ -1,3 +1,4 @@
+#include <stdlib.h>
```
am11: Please only add it in `src/sass_context_wrapper.h` and it will take care of the rest.

am11:
Thanks for the contribution. :+1: 

I have left some comments, PTAL.

src/sass_context_wrapper.cpp
am11: Also consider adding preprocessor:

```c++
#ifdef __FreeBSD__
```

ahacking: It is standard ANSI C to include stdlib.h for malloc, and not platform specific. See http://en.m.wikipedia.org/wiki/C_dynamic_memory_allocation

am11: I know this is a standard header but this is only required for freebsd at the moment. On other platforms, gcc is compiling without it.

ahacking:
Moved <stdlib.h> to common header and rebased.

ahacking:
@am11 Including <stdio.h> is correct.  Failing to include <stdlib.h> means ``malloc()`` (if not being defined fortuitously through some indirect include) will be treated as returning ``int`` which is wrong. It may still 'work' if you're lucky when sizeof(int) == sizeof(void*).

Taking the responsible approach when using malloc/calloc/free is to include <stdlib.h> as per the C standard. If we were including for example a Linux/glibc non-standard header like ``<malloc.h>`` then yes that deserves being surrounded in an ``#ifdef``guard, but putting a platform specific guard for an expressly portable inclusion is a bad idea and just increases risk of bad things happening and also creates unnecessary maintenance.

I am finding that master of node-sass generates a Segmentation Fault with my app.scss. I have run the tests and whilst there is a timeout failure with the huge test,  there isn't a segfault, so my app.scss is triggering a regression (it works fine with 1.x). I haven't been able to determine where or what in the code is causing the problem yet but if I can isolate the cause I will open a separate issue.

am11:
@ahacking, thanks! :tada: 

Regarding the issue, please open it in the issue tracker with steps to reproduce the error: Sass code, command you used to execute etc.

am11:
@ahacking, note that it could be due to the unsupported language features in the `.scss` file. Usually segfaults are due to LibSass code (but not necessarily).

PR #660, opened 2015-02-10
Misc. Fixes

am11:
* Build: Use node-gyp from dependencies.
  * Code: Minor refactoring.
* CI: Tests io.js runtime.
  * Promotes node.js to v1.2.
* Test: Adds outFile related tests. 
* Test: Adds sourceMap path related tests.


am11:
Coverage!!! :zap:

austinpray:
@am11 keep going!! 100%%% coverage or go home

xzyfer:
Can you give some context to what you mean by 

> Use node-gyp from dependencies

am11:
@xzyfer, for example I used to run `node-gyp rebuild` or `node scripts/build -f`. Both of these uses node-gyp installed at global scope (`npm install -g node-gyp`). Also, node-gyp was not in our package dependency. There are two problems with this:

- only pangyp (a fork of node-gyp) supports io.js at the moment (soon official node-gyp will bring the support).
- people seldom update their global installations.

With this change, all you need to do is to run `node scripts/build`, `node scripts/build --force` (or `-f`) or `node scripts/build --debug` (or `-d`). You can use both -f and -d flags (`-fd` or `-df` isn't supported though). This will execute (shell out) runtime-aware build command.

am11:
io.js is failing because it is building binary with io.js runtime but `npm test` job is agnostic of runtime at the moment. That would be a complicated solution to make CIs scripts runtime-aware. Something like: https://github.com/koajs/koa/commit/6d5f506c99ac31014949e166e2fedcc428a38d60.

Also something has changed in io.js v1.1 (and perhaps corresponding libuv), that it is failing some tests locally with that as well. (was passing all with io.js v1.0.x though).

Nonetheless, we can build io.js 1.1 binaries for node-sass 2.0.0 with pangyp (node scripts/build). :sunglasses:

am11:
@austinpray that is the plan! :)

Just gained a bit more with this diff:

```diff
- try { 
-  err = JSON.parse(err); 
- } catch (e) { 
-  err = { message: err }; 
- } 
- 
+    err = JSON.parse(err); 
```

Note that I removed try-catch because we don't need defensive programming. It is an "interface promise" by LibSass API that the errors shall be returned in JSON format. If there is an anomaly then the bug belongs to LibSass. :beetle:

xzyfer:
Thanks for the clarification @am11. The global node-gyp caught me out recently as well. This is much better.

PR #663, opened 2015-02-11
Improve install script

rubennorte:
The main goal of this PR is to solve an error. When there are some kind of errors downloading the binary file (network errors mostly) the file where it was going to be downloaded to is not unlinked and left as an empty file and then the build script "thinks" that the download went ok and does not do the manual build. I changed the code to create the file just when it is necessary.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -2,7 +2,9 @@ var fs = require('fs'),
     path = require('path'),
     request = require('request'),
     mkdirp = require('mkdirp'),
-    exec = require('shelljs').exec;
+    exec = require('shelljs').exec,
+    npmconf = require('npmconf'),
+    packageInfo = require('../package.json');
 
 require('../lib/extensions');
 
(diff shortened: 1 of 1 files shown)


scripts/install.js
```diff
@@ -49,25 +41,26 @@ function download(url, dest, cb) {
...
-        getProxyFromEnv = false;
-        cb(options);
-        return;
-      }
-    });
+  npmconf.load({}, function (er, conf) {
+    var proxyUrl;
+
+    if (!er) {
+      ['https-proxy', 'proxy', 'http-proxy'].some(function(setting) {
```
am11: :+1:

am11:
Thanks for your contribution. :+1: 

My hands are full with this binary patch thingy to confirm the changes. @xzyfer, can you please take a look if this safe to merge at this point?

scripts/install.js
xzyfer: FWIW this could be reduced to 

```js
var proxyUrl = ['https-proxy', 'proxy', 'http-proxy'].filter(function(setting) { return conf.get(setting); })[0];
```

xzyfer: Or even simpler yet?

```js
var proxyUrl = conf.get('https-proxy') || conf.get('proxy') || conf.get('http-proxy');
```

xzyfer:
LGTM. Waiting on CI.

am11:
:tada:

PR #667, opened 2015-02-12
Code: Minor fixes.

am11:
//cc @xzyfer.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -22,11 +22,13 @@ function download(url, dest, cb) {
     var returnError = function(err) {
       cb(typeof err.message === 'string' ? err.message : err);
     };
+
     request.get(url, options).on('response', function(response) {
       if (response.statusCode < 200 || response.statusCode >= 300) {
         returnError('Can not download file from ' + url);
         return;
       }
+
       response.pipe(fs.createWriteStream(dest));
     }).on('error', returnError);
   });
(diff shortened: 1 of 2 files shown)


scripts/install.js
```diff
@@ -45,13 +47,9 @@ function applyProxy(options, cb) {
...
 
     if (!er) {
-      ['https-proxy', 'proxy', 'http-proxy'].some(function(setting) {
-        var npmProxyUrl = conf.get(setting);
-        if (npmProxyUrl) {
-          proxyUrl = npmProxyUrl;
-          return true;
-        }
-      });
+      proxyUrl = ['https-proxy', 'proxy', 'http-proxy'].filter(function(setting) {
```
xzyfer: ```proxyUrl = conf.get('https-proxy') || conf.get('proxy') || conf.get('http-proxy');```

xzyfer: The block below can become 

```
proxyUrl = proxyUrl || env.HTTPS_PROXY || env.https_proxy || env.HTTP_PROXY || env.http_proxy;
```

am11: Agree that is much better. :+1: 

Fixed by 63331e0.

For better approach in the scenario, where the given array's size is larger, see http://stackoverflow.com/q/28476629/. Lets see what kind of responses we get there! :)

xzyfer: I use lodash for that https://lodash.com/docs#find

Surprising node-sass is one of the few codebases without lodash or underscore haha

xzyfer: FYI

```
['https-proxy', 'proxy', 'http-proxy'].reduce(function(prev, cur) { 
  return prev || conf.get(cur); 
}, false);
```

am11: Yes! underscore definitely provides a great deal of functions when it comes to Array. We can include it in njs projects, but nativity and [vanilla.js](http://vanilla-js.com/) is something I always prefer. :smiley: 

Thanks to node.js to rely on one JS runtime and does not present JS incompatibility issues like those in cross-browser scenarios.

PR #668, opened 2015-02-12
Misc. Fixes

am11:
* README: Removes v0.11 workaround section
  * Reason: njs v0.11 is not relavent any more.
  * Solution: Update to node v0.12 with which<br />
                  node-sass v2.0.1 is compatible OOTB.

* CI: Updates io.js to 1.2 for AppVeyor.

* Code: Throws `Error` object.
  * Returns json-string error from LibSass as is.
  * Extends empty error object with parsed JSON.<br/>
  Note: The reason why [`NanError`](https://github.com/rvagg/nan#nanerrormessage-nantypeerrormessage-nanrangeerrormessage) is not used<br />because LibSass returns error as JSON<br /> string with `message` as member, so we would<br/> have to reparse and reassign the obj in that case.<br/>Issue URL: #675.

* README: Updates error info. (related to previous commit)

* Code: Uses more C++ style nomenclature. 

* Code: Resolves symbolic link.  <br/>On Windows, process.execPath does not
get resolved symlink path. This commit
fixes the issue.<br/>Related Issue: iojs/io.js#851.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,9 +279,6 @@ npm install -g node-gyp
 node-gyp rebuild  # to make debug release, use -d switch
 ```
 
-### Workaround for node `v0.11.13` `v0.11.14`
-Follow the steps above, but comment out this [line](https://github.com/sass/node-sass/blob/e01497c4d4b8a7a7f4dbf9d607920ac10ad64445/lib/index.js#L181) in `lib/index.js` before the `npm install` step. Then uncomment it back again, and continue with the rest of the steps (see issue [#563](https://github.com/sass/node-sass/issues/563)).
-
 ## Command Line Interface
 
 The interface for command-line usage is fairly simplistic at this stage, as seen in the following usage section.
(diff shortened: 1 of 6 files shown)


kevva:
Good stuff lately @am11! This look good to me.

am11:
Thanks for the feedback @kevva. :)

am11:
@xzyfer, PTAL.

Note: It does not fix the AppVeyor issue with iojs v1.2 because they are not redirecting node --> iojs when iojs is an active runtime, but they are instead having a concrete node.exe with v1.2 which possesses wrong proc info. Probably they will fix it with the way nvmw handle it: https://github.com/appveyor/ci/issues/139#issuecomment-74421338. Otherwise, we will have to induce that cmd via appveyor.yaml (while deleting node.exe from iojs directory).

xzyfer:
Nothing jumps out at me :+1:

Issue #674, opened 2015-02-13
Return calculated Source Map output path in stats object


sindresorhus:
Currently I have to manually calculate the output path of the source map file based on the `sourceMap` option value. Would be handy if node-sass returned this so every consumer didn't have to duplicate that logic.

https://github.com/sindresorhus/grunt-sass/blob/bb1136d18cdca2be8529ed92a2620b391c2b20c6/tasks/sass.js#L30

am11:
Agree. :+1:

am11:
This was addressed by 2aebbd1c8b6fd2c792d87e4f3405a84d6915d62b via https://github.com/sass/node-sass/pull/686. 

You would be able to access `this.options.sourceMap` in (both method-like or node-like) callbacks.

Issue #676, opened 2015-02-13
More Node like API


sindresorhus:
The API is currently a bit awkward as it doesn't follow the Node convention.

Instead of:

```js
sass.render({success: function () {}, error: function () {});
```

It should be:

```js
sass.render({}, function (error, result) {});
```

<del>Or embrace promises:

```js
sass.render({}).then(function() {}).catch(function () {});
```
</del>

am11:
We refined API in v2 (https://github.com/sass/node-sass/issues/547), but I agree this is not the end. Promises is the best approach IMO and the syntax will be compatible with PostCSS as well.

Incidentally, with:

```javascript
sass.render({}).then(function() {}).catch(function () {});
```

in `render()` are you proposing to send in `options` object and `then()` will have a success callback?

We would also need to consider custom importers and the upcoming custom functions (https://github.com/sass/node-sass/pull/644).

sindresorhus:
> in render() are you proposing to send in options object and then() will have a success callback?

Yes.

> We would also need to consider custom importers and the upcoming custom functions (#644).

They belong in the options object.

am11:
@sindresorhus, what do you think if we instead go with reactive pattern: http://reactive-extensions.github.io/RxJS/?

sindresorhus:
I like the reactive pattern, but I strongly feel the API should be promise based as it's the standardized describe async operations. Users can easily use any reactive pattern upon a promise interface.

dlmanning:
My two cents: please not promises. Error handling from promises is still kind of fubar. It's trivially easy to wrap the traditional node callback pattern in a promise if your project is using them, but I think it's still premature for a project as widely depended upon as node-sass to adopt them for its api.

xzyfer:
My thoughts here. Leave the existing callback API. I'm aware it's not popular opinion, but it is most accessible.

As @dlmanning points out it's trivial to progressively enhance callbacks to promises, the reverse not so much (it's certainly doable). Arguing that anyone who uses node-sass should be using promises because it's _the right way_ is no argument at all. 

As for the reactive approach, all my points on promises apply, with the addition that it has yet to see wide adoption. Add to that native async/await is coming in ES7, making this the primary API seems premature.

Generally speaking I believe we should aim for providing the lowest developer friction to widest audience, providing sane defaults were possible. If we're going to jump through hoops to support old CentOS I see no reason not to give the attention to the node side of things.

sindresorhus:
[...]

PR #681, opened 2015-02-15
Render options in callbacks as this.options

Jakobo:
This patch and tests introduces a context for all callbacks currently
in node-sass. Right now, they run w/ an ambiguous context, making the
`this` keyword not helpful to importer developers (and eventually
custom function developers). This creates a context object, adds the
render options to said context object, and then uses that context for
importer, success, and error callbacks via `this.options`.

By using a context instead of just `.call(options...)` there is a hook
for later should we find additional environmental information is
valuable to importer / custom function developers such as the node
environment, node-sass version, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The API for using node-sass has changed, so that now there are only two variable
 `data` is a `String` containing the scss to be rendered by [libsass]. One of this or `file` options are required, for both render and renderSync. It is recommended that you use the `includePaths` option in conjunction with this, as otherwise [libsass] may have trouble finding files imported via the `@import` directive.
 
 #### deprecated: success / error
-> `success` is a `Function` to be called upon successful rendering of the scss to css. This option is required but only for the render function. If provided to `renderSync` it will be ignored. The error object take
+> `success` is a `Function` to be called upon successful rendering of the scss to css. This option is required but only for the render function. If provided to `renderSync` it will be ignored.
 >
 > The callback function is passed a results object, containing the following keys:
 >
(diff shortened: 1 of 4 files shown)


am11:
Thanks for the PR. 

Incidentally, does it also fix https://github.com/sass/node-sass/issues/674?

am11:
Also this is related: https://github.com/sass/node-sass/issues/676. And we are targeting that for vNext (2.1).

It may require the change in this functionality again.

Jakobo:
Regarding #674 it would be the options object from `getOptions`, so yes the `this.options.sourceMap` would resolve. 

For #676 I have no problem rebasing this, especially after we add in node style callbacks or promises to the API. I don't really see `getOptions` changing much other than dropping the success/error registration. More than happy to adjust the PR in the next few days.

chriseppstein:
This patch's changes needs to be applied to #644 handling of custom function invocation so that `this.options` works in custom functions as well.

Jakobo:
I agree @chriseppstein. If we can get custom functions rebased, I will include those in the patch as well. 

I'd like to make sure there are no concerns with the scope of our importers and custom functions 
 being an object literal that describes the current render environment. It seems logical without introducing backcompat issues.

Jakobo:
This is now rebased off of #684

am11:
@Jakobo, did you merged your branches by mistake?  It was suppose to have one commit, now there are eight commits. Also CI build is failing. :(

Jakobo:
@am11, I'm sorry, I thought you wanted this rebased off of #684 (rebase is dbf31bac), so it used node style callbacks.

> If you can rebase #681, we can merge it right away.

I will close this and create a new PR based off of HEAD

am11:
@Jakobo, I meant rebase against the node-sass master as there were merge conflicts at the time..

Jakobo:
Ah okay. Then please see #686 which is directly based off of master with `npm test` succeeding locally. Thanks for clarifying.

Issue #677, opened 2015-02-13
Options: Inherit LibSass defaults


am11:
As per this discussion which transpired here: https://github.com/sass/node-sass/issues/673, we need to inherit all options defaults from LibSass (except for paths). The key is to have conditional setting of those options in `binding.cpp`.

//cc @sindresorhus

am11:
We have two options, add `if`s on setting every option, set all options for all cases and set default (the current behavior). I think the current behavior is better, i.e. not uglifying C++ code.

@xzyfer, thoughts?

am11:
The current behavior is less noisy at C++ side, so closing this one..

PR #686, opened 2015-02-19
Render options in callbacks as this.options

Jakobo:
Replaces #681 (single focused commit based on sass/node-sass@master)

This patch and tests introduces a context for all callbacks currently
in node-sass. Right now, they run w/ an ambiguous context, making the
`this` keyword not helpful to importer developers (and eventually
custom function developers). This creates a context object, adds the
render options to said context object, and then uses that context for
importer, success, and error callbacks via `this.options`.

By using a context instead of just `.call(options...)` there is a hook
for later should we find additional environmental information is
valuable to importer / custom function developers such as the node
environment, node-sass version, etc.

Additionally, info() is exposed in the context so any node modules
imported would be able to determine the libsass & node-sass versions
used during the render call.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -127,7 +127,6 @@ function getSourceMap(options) {
  * @param {Object} options
  * @api private
  */
-
 function getOptions(options) {
   options = options || {};
   options.comments = options.source_comments || options.sourceComments || false;
@@ -140,6 +139,9 @@ function getOptions(options) {
   options.sourceMap = getSourceMap(options);
   options.style = getStyle(options) || 0;
 
+  // context object represents node-sass environment
+  options.context = { options: options };
+
   if (options.imagePath && typeof options.imagePath !== 'string') {
     throw new Error('`imagePath` needs to be a string');
   }
@@ -149,7 +151,7 @@ function getOptions(options) {
 
   options.error = function(err) {
     if (error) {
-      error(util._extend(new Error(), JSON.parse(err)));
+      error.call(options.context, util._extend(new Error(), JSON.parse(err)));
     }
   };
 
@@ -158,7 +160,7 @@ function getOptions(options) {
     var stats = endStats(result.stats);
 
     if (success) {
-      success({
+      success.call(options.context, {
         css: result.css,
         map: result.map,
         stats: stats
@@ -207,7 +209,7 @@ module.exports.render = function(options) {
         });
       }
 
-      var result = importer(file, prev, done);
+      var result = importer.call(options.context, file, prev, done);
 
       if (result) {
         done(result);
@@ -232,7 +234,7 @@ module.exports.renderSync = function(options) {
 
   if (importer) {
     options.importer = function(file, prev) {
-      return { objectLiteral: importer(file, prev) };
+      return { objectLiteral: importer.call(options.context, file, prev) };
     };
   }
 
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ Note: If this option is provided to renderSync it will be ignored. In case of `r
 #### importer (starting from v2)
 `importer` is a `Function` to be called when libsass parser encounters the import directive. If present, libsass will call node-sass and let the user change file, data or both during the compilation. This option is optional, and applies to both render and renderSync functions. Also, it can either return object of form `{file:'..', contents: '..'}` or send it back via `done({})`. Note in renderSync or render, there is no restriction imposed on using `done()` callback or `return` statement (dispite of the asnchrony difference).
 
+The options passed in to `render` and `renderSync` are available as `this.options` within the `Function`.
+
 #### includePaths
 `includePaths` is an `Array` of path `String`s to look for any `@import`ed files. It is recommended that you use this option if you are using the `data` option and have **any** `@import` directives, as otherwise [libsass] may not find your depended-on files.
 
--- a/test/api.js
+++ b/test/api.js
@@ -299,7 +299,7 @@ describe('api', function() {
         }
       });
     });
-	
+
     it('should override imports with "data" as input and fires callback with contents', function(done) {
       sass.render({
         data: src,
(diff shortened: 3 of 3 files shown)


lib/index.js
```diff
@@ -140,6 +140,14 @@ function getOptions(options) {
   options.sourceMap = getSourceMap(options);
   options.style = getStyle(options) || 0;
 
+  // context object represents node-sass environment
+  options.context = {
+    info: module.exports.info,
+    state: {},
+    sync: false,
```
am11: Why do we need these three? `info` and `sync` are useless IMO, `state` can be set by user if they need. //cc @kevva.

Jakobo: A good example of where these three items are useful is in scenarios where the custom importers (and eventually custom functions) come from npm modules. In these contexts, the importer function probably does not have any information about how it was invoked.

* state: This persistent object allows an importer or custom function to keep track of data across invocations. An example of this would be providing a custom importer that allows for import-once style functionality. I'm happy to remove this if we would prefer to let custom importer authors develop their own solutions
* sync: At the time of importer, the importer has no knowledge of if it was invoked via render or renderSync. While this is not as important for importers, it is definitely important for custom functions, as some custom functions may only work in a purely async manner. I'd really like to see this stay so that importer/function authors can know more about how they were invoked.
* info: It seemed valuable that an importer (or eventually custom function) could ask about the node-sass / libsass versions. I'm okay with removing this if we don't have a clear use case.

am11: Thanks for explaining the use-cases. 

* We are deep cloning the option object. 
* Since the user can do `this.options.whatever = { 'bah': 'blah' }` to fulfill their custom requirements, they can append all that to options object the way they please. 
* The idea really is to keep the API itself less noisier and sleeker.

Jakobo: That totally makes sense. In ruby sass, there is a `custom` element and developers are told not to write to the top level hash. http://sass-lang.com/documentation/file.SASS_REFERENCE.html Should we instead rename this to `custom` for consistency? I do think there's a lot of value in people not writing into the top level hash or the option hash.

am11: @Jakobo, 

* node-sass is not complying with Ruby-Sass, since two runtimes are intrinsically different and we are more focused on node-way or JavaScript-way of doing things here.
* we are not restricting / filtering the options object anyway and the option object is open-ended. Users are free to consume it.
* info() have no or very obscure usage in importers.

> I do think there's a lot of value in people not writing directly into the top level hash or the option hash.

1- Why?
2- What is the purpose of having deep-nested clone?

Jakobo: Totally makes sense. Thanks for explaining the logic and rationale. :)

For 1) if we don't restrict writing to the options hash, technically any new options we add risk breaking backwards compatibility. We could just tell import / custom function devs to namespace their items stored in the options hash to avoid conflicts.

For 2) importers used in gulp or grunt pipelines use the same options collection repeatedly. It seemed safer to work on a clone of the options. Also happy to pull this out.

More than happy to remove any/all of these items if we think they're outside the goals of node-sass.

am11: @Jakobo, generally these are very good ideas. But after few iterations, you will notice that new use-cases will emerge for the features we initially took lightly. As soon as you try to do something creative here, it may cause chaos in the, the tertiary, quaternary .. downstream of node-sass.

Therefore, I think it is a good idea to keep the core API side as simple/dumb as possible unless of course it is utterly necessary. 

For example, we are even resolving the output/source-map file paths just because libsass require it. And we will be removing the (remaining) default values for options and let libsass set them (https://github.com/sass/node-sass/issues/677).

For this PR. I propose to let the consumers:

* use or misuse `options` object and realise if something goes wrong while misusing it. :)
* clone the `options` object before passing it to node-sass if they are sharing it elsewhere for other purposes.
* figure out whether the custom function or importer is sync or async if necessary. If they are really defining a function to work with both render[Sync], they can just `return` the value and it will work for both. For more mysterious use-case (callback for async and return for sync), they are free to override `this.options` with custom payload.
* make use of `this.options` in importers, success and error callbacks.

Jakobo:
Based on feedback from @am11, I have removed any additional items from the context object. We can revisit adding them if there are use cases that make it absolutely necessary. The context object is now much more straightforward:

`context = { options: options }` allows `this.options` to work in success, error, and importer callbacks. Thanks for the reviews, feedback, and rationale explanations.

package.json
```diff
@@ -44,6 +44,7 @@
   ],
   "dependencies": {
     "chalk": "^0.5.1",
+    "clone": "^1.0.0",
```
am11: This is not needed anymore.

lib/index.js
```diff
@@ -149,7 +151,7 @@ function getOptions(options) {
 
   options.error = function(err) {
     if (error) {
-      error(util._extend(new Error(), JSON.parse(err)));
+      error.call(options.context, util._extend(new Error(), JSON.parse(err)));
```
am11: Can we just say `error.call(options, ...)` and remove options.context?

Jakobo: I don't think that would be a good idea. It would paint us into a bit of a corner if we need to surface any additional environment information from libsass or node-sass to callbacks.

package.json
Jakobo: Agree. Will remove. Sorry for missing it. :)

lib/index.js
am11: Since we are not cloning `options` on `context` instantiation, I think one problem with this approach would be the shallow copy dropping functions definitions (`importer` et al.).

Jakobo: I'm sorry @am11, I don't think I understand here. I wrote a test to see if anything was being dropped from the options object or we were somehow losing our function definitions.

```js
it('should preserve options by reference', function(done) {
  var options;
  options = {
    data: src,
    success: function() {
      assert.strictEqual(this.options, options);
      assert.strictEqual(this.options.importer, options.importer);
      done();
    },
    importer: function() {
      return {
        contents: 'div {color: yellow;}'
      };
    }
  };
  sass.render(options);
});
```

Because we are mutating objects by reference, everything seems to arrive as expected in the callbacks. Should I change something in the above test to capture your concerns? Thanks!

am11: Oh nice! I thought it will drop those. Thanks for the example! (we can add this test to the mix as well) :+1: 

Separately, are you suggesting if we drop `options.context`, then any new member to `options` set by node-sass will be lost in callbacks? Can you please provide a concrete example for [this](https://github.com/sass/node-sass/pull/686#discussion_r24995696), which shows that it will fail on some account in case we drop `.call(options.context, ...)` in favour of `.call(options, ...)`?

Jakobo: Sure.

I'm worried about the case where we write this as `importer.call(options, ...)`. For this, options such as `file` will be accessible as `this.file`. It will work as long both libsass and node-sass do not need to share any additional information with custom importers and functions.

In ruby-sass which libsass seeks parity with, there is a concept of a global environment. I can see us providing this in a more node-like way as opposed to the ruby pattern of creating a global. If we are currently binding to `options`, then we would be unable to add this environment information without a breaking change (as references to `this.file` would suddenly fail). The `context` object fulfills this role, allowing us to later add environment information `{ options: options, env: getEnvironment() }` without breaking anyone's code.

am11: Thanks and points for future consideration. :+1: 

TBH, I haven't delve into global environment, I cannot say for a fact whether it will land in libsass world or not. Maybe @xzyfer can confirm it.

However, I do think that it will not break the user code in this case even if global environment is introduced in libsass. This is because whatever libsass is sharing with node-sass, it arrives via the binding layer which uses v8 lib (via nan wrapper). That allows us to assign / inject any object to a `persistent` (v8 terminology) JS object. So this sharing with custom imports or functions is taken care by the binding code, where we can alter the state of other objects in the current loop (libuv's `uv_loop`).

am11: Either way it does not hurt to keep the context. If you can rebase out the package.json change, we can merge this. :)

Jakobo:
Latest in the PR @ 2aebbd1 has the package.json changes rebased out. It should be a clean commit now.

am11:
Excellent! :tada:

Issue #694, opened 2015-02-22
Better support for io.js rapid release cycle


xzyfer:
Io.js 1.3 has been released and is now failing to download the binary.

@am11 I think we should update the install script to be less strict on version mathing for io.js. Their releases more frequently than ours, and are unlikely to have broken older binaries.

I think in the long run it's better for us to assume binaries are good until proven broken (at least for io.js) rather than trying to keep releases in sync.

Globegitter:
Yep I think that sounds good. Keeping support for all 1.x.x versions sounds imo reasonable.

am11:
That was fast! Its like yesterday they released v1.2.. :stuck_out_tongue_winking_eye: 

See the comment by @kkoopa https://github.com/sass/node-sass/issues/627#issuecomment-70419098. As soon as the minor version changes, it signifies a breaking change and requires a binary rebuild.

However, I can confirm that our iojs-v1.2 binary for win-x64 is working with iojs-v1.3.

For the resolution, I agree that instead of changing the binary names, we should change the install script to fallback to first available version. Here is one way to go about it: where we call the download function, retry with one lesser version in case of 404 and repeat it thrice before giving up.

Globegitter:
@am11 Doesn't a change in a minor version mean that the API has been changed in a backwards-compatible manner? That is at least following http://semver.org/ closely. That was my reasoning behind the comment.

This more fundamental discussion aside, I think this sounds like a good solution :+1:

am11:
The current stable version of node.js is v0.12.0 and the major version hasn't been changed yet. It could be that all changes ever made in node.js are backward compatible but those do not account for the breaking changes presented by v8 and libuv (on which the C++ add-on binaries are dependent upon).

Globegitter:
Oh yeah that is a good point @am11, especially when getting into the land of C/C++ :)

xzyfer:
An alternate suggestion would be to change the version string in the binary path from the compiled version to the lowest supported version i.e.

```
- darwin-x64-iojs-1.0
- darwin-x64-iojs-1.2
- darwin-x64-iojs-1.7
```

This essentially translates to

```
- darwin-x64-iojs-1.0 < 1.2
- darwin-x64-iojs-1.2 < 1.7
- darwin-x64-iojs-1.7+
```

Then use the github [contents API](https://developer.github.com/v3/repos/contents/#get-contents) to query the directory listing. 

[...]

PR #695, opened 2015-02-22
Build: Uses v8 version instead of runtime name.

am11:
The new binary path format is: 
`<platform>-<arch>-<v8 version>/binding.node`

Issue URL: #694.


xzyfer:
:+1:

saper:
Please see [#694  (comment)](https://github.com/sass/node-sass/issues/694#issuecomment-77773989) why I think we should use `NODE_MODULE_VERSION` instead.

Issue #669, opened 2015-02-12
CLI: print results to stdout by default, when no output target is supplied


am11:
From @xzyfer's [comment](https://github.com/sass/node-sass/issues/626#issuecomment-70050489):

> what are your thoughts on removing the `--stdout` and making it the default behaviour if an output directory or file isn't supplied. This the behaviour of Ruby sass and Libsass.

Adding to v2.1 milestone.

am11:
Addressed by 433e939 via #701.

qm3ster:
`README.md` still says
`Output will be saved with the same name as input Sass file into the current working directory if the --output flag is omitted.`
Fixed in #1363 please merge.

am11:
@qm3ster, merged, thanks!
Aside, note that `--output` accepts the directory path instead of path including filename. So `node-sass blah.scss --output myblah.css` will actually write to `{cwd}/myblah.css/blah.css` (it will also do the `mkdir myblah`, if that was non-existent). If the intent is to write to a file with different name, then we can redirect output as `node-sass blah.scss > myblah.css`.

qm3ster:
Yup, well aware, I use both successfully.

PR #701, opened 2015-02-23
CLI: Retires stdout option

am11:
When no destination or output folder is provided,
it will print the output to stdout.

If the source-map option was set, it will print
it as well.

Issue URL: #669.
PR URL: #701.

//cc @xzyfer


xzyfer:
:sob: it's beautiful

am11:
:tada:

PR #714, opened 2015-02-27
pangyp required at runtime for post-install from build.js

hmalphettes:
```
➜  tmp  npm i node-sass
|
> node-sass@2.0.1 install /Users/hugues/proj/tmp/node_modules/node-sass
> node scripts/install.js

Can not download file from https://raw.githubusercontent.com/sass/node-sass-binaries/v2.0.1/darwin-x64-iojs-1.4/binding.node

> node-sass@2.0.1 postinstall /Users/hugues/proj/tmp/node_modules/node-sass
> node scripts/build.js

module.js:322
    throw err;
          ^
Error: Cannot find module '/Users/hugues/proj/tmp/node_modules/node-sass/node_modules/pangyp/bin/node-gyp'
    at Function.Module._resolveFilename (module.js:320:15)
    at Function.Module._load (module.js:262:25)
    at Function.Module.runMain (module.js:485:10)
    at startup (node.js:112:16)
    at node.js:863:3
Build failed
```

pangyp is assumed to be installed as a local module by script/build.js.
But it is only installed for development.

This PR fixes that issue.

Issue #693, opened 2015-02-22
npm ERR! node-sass@2.0.1 postinstall: `node scripts/build.js` npm


metasean:
While trying to install an npm package that requires gulp-sass, I'm getting installation errors for node-sass.

### Background to explain output and log

I have one Snow Leopard and one Lion partition on my MacBook.  I'm using Snow Leopard, but accessing a shared projects directory that is based in the Lion partition.  For clarity in the output and log, I've changed the usernames to indicate which is which.  

I have been using this dual system with npm for almost a year now.  I've got over 700 packages in my user .npm directory, and never had a problem like this.  In this case, I was installing the node packages for MEAN.js and the other 76 packages installed without a hiccup. 

## npm install node-sass

When trying to run `npm install node-sass` from iojs-v1.2.0, I get the following (I started with node v0.10.36, v0.33.00, and node v0.12.0 and got similar results with each):
```
-
> node-sass@2.0.1 install /Volumes/Lion Partition/Users/userOnLion/projects/adLoc-meanjs-node/node_modules/node-sass
> node scripts/install.js


> node-sass@2.0.1 postinstall /Volumes/Lion Partition/Users/userOnLion/projects/adLoc-meanjs-node/node_modules/node-sass
> node scripts/build.js

`darwin-x64-iojs-1.2` exists; testing
module.js:341
  Module._extensions[extension](this, filename);
                               ^
Error: dlopen(/Volumes/Lion Partition/Users/userOnLion/projects/adLoc-meanjs-node/node_modules/node-sass/vendor/darwin-x64-iojs-1.2/binding.node, 1): Library not loaded: /usr/lib/libc++.1.dylib
  Referenced from: /Volumes/Lion Partition/Users/userOnLion/projects/adLoc-meanjs-node/node_modules/node-sass/vendor/darwin-x64-iojs-1.2/binding.node
  Reason: image not found
    at Error (native)
    at Module.load (module.js:341:32)
    at Function.Module._load (module.js:296:12)
    at Module.require (module.js:351:17)
    at require (module.js:370:17)
    at Object.<anonymous> (/Volumes/Lion Partition/Users/userOnLion/projects/adLoc-meanjs-node/node_modules/node-sass/lib/index.js:188:15)
    at Module._compile (module.js:446:26)
    at Object.Module._extensions..js (module.js:464:10)
    at Module.load (module.js:341:32)
    at Function.Module._load (module.js:296:12)
npm ERR! Darwin 10.8.0
[...]

PR #715, opened 2015-02-27
postinstall fails with iojs-1.4.1

hmalphettes:
I encountered the same error than kevinSuttle here: https://github.com/sass/node-sass/issues/693#issuecomment-76340269

iojs-1.4.1 installed via nvm.

```
➜  tmp  npm i node-sass
|
> node-sass@2.0.1 install /Users/hugues/proj/tmp/node_modules/node-sass
> node scripts/install.js

Can not download file from https://raw.githubusercontent.com/sass/node-sass-binaries/v2.0.1/darwin-x64-iojs-1.4/binding.node

> node-sass@2.0.1 postinstall /Users/hugues/proj/tmp/node_modules/node-sass
> node scripts/build.js

module.js:322
    throw err;
          ^
Error: Cannot find module '/Users/hugues/proj/tmp/node_modules/node-sass/node_modules/pangyp/bin/node-gyp'
    at Function.Module._resolveFilename (module.js:320:15)
    at Function.Module._load (module.js:262:25)
    at Function.Module.runMain (module.js:485:10)
    at startup (node.js:112:16)
    at node.js:863:3
Build failed
```

On my Macos machine there is no homebrew installation of nodejs or iojs; it is important because homebrew does not install the proper patched npm for iojs and one ends-up having to use pangyp as an alternative to node-gyp when that happens.

The stack trace shows that pangyp, a devDependency, was not installed.
So I moved pangyp to a runtime dependency and got past that problem: https://github.com/sass/node-sass/pull/714#issuecomment-76371437

However the sass binary produced remained incompatible.

I have reverted the use of pangyp (https://github.com/hmalphettes/node-sass/commit/5a113952bf517ebb96f50204431b8f2d8844500d) and put back node-gyp and everything is working like a charm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "coveralls": "^2.11.2",
     "jscoverage": "^0.5.9",
     "jshint": "^2.6.0",
-    "mocha-lcov-reporter": "^0.0.1",
-    "pangyp": "^2.1.0"
+    "mocha-lcov-reporter": "^0.0.1"
   }
 }
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -3,7 +3,7 @@ var fs = require('fs'),
     spawn = require('child_process').spawn,
     mkdir = require('mkdirp'),
     Mocha = require('mocha');
-    
+
 require('../lib/extensions');
 
 /**
@@ -50,11 +50,8 @@ function afterBuild(options) {
  */
 
 function build(options) {
-  var arguments = ['node_modules/pangyp/bin/node-gyp', 'rebuild'].concat(options.args);
-
-  console.log(['Building:', process.runtime.execPath].concat(arguments).join(' '));
-
-  var proc = spawn(process.runtime.execPath, arguments, {
+  var bin = options.platform === 'win32' ? 'node-gyp.cmd' : 'node-gyp';
+  var proc = spawn(bin, ['rebuild'].concat(options.args), {
     stdio: [0, 1, 2]
   });
 


am11:
Thanks for your contribution. I am in the middle of changing this build related stuff in another PR.

There are two issues with this PR:

1 - You are assuming that end user will already have node-gyp globally installed, which might not be the case. This is exactly what we were assuming before I added pangyp as dependency (as io.js appeared at the same time).

2 -

> ..put back node-gyp and everything is working like a charm.

Yes it seems like it (I am actually surprised, last time I tried building io.js binary, node-gyp just said "no!"). Even though things are looking pretty promising, we should not make any assumption unless this is merged: https://github.com/TooTallNate/node-gyp/pull/564. Till then, pangyp is the build tool of choice.

hmalphettes:
@am11 thanks for your attention.

I had not realised that node-gyp could ever be not installed when running `npm install`. I'll research that.

Yes I can't wait for https://github.com/TooTallNate/node-gyp/pull/564 and the upstream https://github.com/iojs/io.js/pull/493 to be resolved.

Should we close this PR then?

am11:
> Should we close this PR then?

Yes. For now, sticking with pangyp would be make sense until those issues in up streams are sorted out.

We had actually achieved a quite stable build process before io.js days. The aim is to re-achieve that stability. We are also trying to avoid manual building step on end terminals as much as possible (in favour of pre-build binaries).

am11:
FYI: https://github.com/sass/node-sass/pull/717.

Issue #710, opened 2015-02-26
Time to ship 3.0.0?


xzyfer:
We have a couple of breaking changes on master already, a couple more in the works. 

I think we should ship vNext with the update sass binary paths (https://github.com/sass/node-sass/pull/695). This opinion in motivated largely due to the volume of io 1.3 issues and knowing that 1.4 is just around the corner.

am11:
> knowing that 1.4 is just around the corner.

Haha! That is so true.

I agree vNext should be 3.0.0 instead. I think we should wait for @matryo to finish #644.

Then I suggest we have alpha release, followed by beta and then RTW. This to nail out proxy related issue: https://github.com/sass/node-sass/issues/705 as well as the other planned for the vNext milestone. 

@xzyfer did you have any luck with FreeBSD binary? I haven't tried it yet..

xzyfer:
> @xzyfer did you have any luck with FreeBSD binary? I haven't tried it yet..

honestly I haven't looked into it but I can try to make some time

> I think we should wait for @matryo to finish #644.

Sounds good. I mostly wanted to emphasise there's not benefit in waiting for Libsass 3.2 before shipping these great updates. Libsass 3.2 is going to be a while off yet, partially because most of my time is taken up by CSSConf AU at the moment.

xzyfer:
I'd also love to have https://github.com/sass/node-sass/pull/704 as part of vNext. This will help to ween people off this feature, and prevent the need for a 4.0 with Libsass@3.2 drops.

xzyfer:
Presented without comment - https://github.com/iojs/io.js/blob/v1.x/CHANGELOG.md#2015-02-26-version-141-rvagg

am11:
Yes! see the "four minutes ago" story https://github.com/appveyor/ci/issues/147#issuecomment-76293987. :smiley:

xzyfer:
Haha

am11:
> @am11 can you confirm that io js 1.4.1 is using the same version on v8 as 1.2?

Nah it is changed! :(

```cmd
C:\Users\Adeel\Source\Repos\node-sass>nvmw use iojs-v1.0.0
Now using iojs v1.0.0

C:\Users\Adeel\Source\Repos\node-sass>node -p process.versions.v8
3.31.74.1

C:\Users\Adeel\Source\Repos\node-sass>nvmw use iojs-v1.1.0
Now using iojs v1.1.0

[...]

Issue #683, opened 2015-02-17
consider adding tests directory to npmignore due to windows 260 character path limitation


dcbarans:
As the title states. I've hit a wall where windows just falls on its face with paths longer than 260 characters. Here's an example path (not including the parent module its inside + my working directory path):
```
node-sass/test/fixtures/spec/spec/extend-tests/104_test_nested_extender_counts_extended_subselectors/104_test_nested_extender_counts_extended_subselectors'
```
The problem is compounded severely when node-sass is used as a module with other modules.

There's 2 solutions i see quickly.
* Add tests directory to npmigore. Though not sure if that's a path you'd be willing to explore.
* Shorten the naming convention of your test directories. Some are crazy long :)

am11:
On Windows 8.1 x64, I keep node-sass under VS Git directory: `C:\Users\Adeel\Source\Repos\node-sass`. I don't have this issue. However, if we have further deep nesting this crazy limitation enforced by Windows will definitely present issues. There is only one true workaround for this issue: use longpath tool; since it is a restriction imposed by OS not NTFS itself (you may want to take a look at the `Pri.LongPath` workaround described here: https://github.com/npm/npm/issues/3697#issuecomment-62281110).

I am afraid we cannot do much about it, because spec is a submodule in our fixture and those file names belong to spec repo: https://github.com/sass/sass-spec.

xzyfer:
@am11 do the npm install scripts still need the specs out of curiosity?

am11:
@xzyfer, not really. With recent modifications, we now only execute *one* test on `npm install` (via mocha code usage) to verify if the binary is compatible and whether to fallback to manual build. What do you think if we (somehow) inline that test code and add `test/` to npmignore as suggested by @dcbarans?

mgreter:
By the way, this is a windows limit as described at: http://stackoverflow.com/a/1880453/1550314

xzyfer:
Presumably would could move the tests we want into the node-sass tests suite and `.npmignore` sass-spec. Seems like a waste for people to download sass-spec each time they npm install, and if it's causing legitimate issues we should ditch it.

It's worth pointing out that the directory depth problem in sass-spec is going to get worse before it gets better i.e. https://github.com/sass/sass-spec/pull/262/files

am11:
I concur. We will discard `test/` directory from npm package and `mocha` will move down to dev-dependencies (since it will then be only required by the coverage script).

dcbarans:
[...]

PR #717, opened 2015-02-27
Misc. improvements

am11:
* Removes tests from publishing. #683
  * Uses new strategy to validate binary after
    the download.
  * Moves mocha under dev-dependencies in
    package.json.

----

* Truncates the forth portion from v8 version
  as discussed in #710.

----

* Moves pangyp from dev to main dependencies in
  package.json.

----

* CLI: Overrides default version info.
  * Based on sindresorhus/meow#2.
  * Adds alias -v.
  * Improves info text; inspired by Less CLI. :)
  * Updates readme.

The new version text looks like:

> $ node-sass -v
>
node-sass&emsp;2.0.1&emsp;&emsp;(Wrapper)&emsp;&emsp;&emsp;&emsp;[JavaScript]
libsass&emsp;&emsp;&ensp;3.1.0&emsp;&emsp;(Sass Compiler)&emsp;&thinsp;&thinsp;[C/C++]

@lukeapage, hope you don't mind!  :copyright: :wink:

----

* Repo: Updates CONTRIBUTING.md.
  * Adds instructions.
  * Removes obsolete info.

This is for DRY'ing comments on repeating issues.
 
---

* CI: Updates and formats configuration files.


am11:
@andreitognolo, does SnapCI actually build and run the tests? It never fails (even when we have something broken) and reports back success in like 5 seconds! Are we missing something here?

.npmignore
```diff
@@ -0,0 +1,8 @@
+*.log
+.DS_Store
+.sass-cache
+build
+lib-cov
+node_modules
+vendor
+test
```
am11: This is `.gitignore` union `test`, because [npm docs](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) says in absence of .npmignore, it takes .gitignore into account and when .npmignore is present, it essentially overrides .gitignore.

lib/extensions.js
```diff
@@ -28,10 +30,26 @@ function getRuntimeInfo() {
  */
 
 function getBinaryIdentifiableName() {
+  var v8SemVersion = process.versions.v8;
```
xzyfer: All this code is essentially just 
```js
var v8SemVersion = process.versions.v8.split('.').slice(0, 2).join('.')
````

File: lib/extensions.js [hunk 3]
```diff
@@ -28,10 +30,26 @@ function getRuntimeInfo() {
...
+  }
+
   return [process.platform, '-',
           process.arch, '-',
-          process.versions.v8].join('');
+          v8SemVersion].join('');
+}
+
+function getSassInfo() {
+  return [
```
xzyfer: :+1:

lib/extensions.js
am11: Shouldn't it be `.slice(0, 3)` to extract semver?

In some releases (for instance node v0.12.0), there are three portions (3.28.73). While in others (all iojs releases till date) have four portions.

I thought it would be good to save a join when v8 version already have three portions.

xzyfer: You're correct it should be `(0,3)`

> I thought it would be good to save a join when v8 version already have three portions.

Since this is essentially a one off operation I don't see any benefit in saving a `join` when it comes at the cost of added verbosity.

It took me a couple reads of this this code to understand it and I already knew what it was trying to do.

am11: Makes sense. :+1: 

Fixed by aa70dbb.

PR #726, opened 2015-03-01
Removal of option hashed based success and error callbacks

Jakobo:
WIP

This change removes the object hash success/error callbacks in
favor of node style callbacks. This includes README and test
updates, including the removal of two tests that are now
obsolete.

This also updates the CLI code pathway to use the node style
callback.

BREAKING API CHANGE
This removes the success/error option keys as ways to register
callbacks into node-sass. Instead, a function should be passed as
the second parameter for the asyncronous `render` call. It takes
a standard node signature of `error, callback`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ var result = sass.renderSync({
 ```
 
 ### Options
-The API for using node-sass has changed, so that now there is only one options hash. In the options hash, some items are optional, others may be mandatory depending on circumstances.
+The API for using node-sass has changed. It supports an options hash and a node-style callback function for `render`, and just the options hash for `renderSync`. In the options hash, some items are optional, others may be mandatory depending on circumstances.
 
 #### file
 `file` is a `String` of the path to an `scss` file for [libsass] to render. One of this or `data` options are required, for both render and renderSync.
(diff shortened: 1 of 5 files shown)


Jakobo:
Some additional notes for @am11 

**options.success / options.error exist internally** I didn't want to make substantial changes internally to the binding code, so it made sense to preserve the options.error / options.success internally. However, attempts to set them from the options hash will be ignored.

**what is the callback object added to the context** Without success/error callbacks directly available to the importer, an importer needs to hook in to the completion of the render lifecycle. While I'd love to do this with EventEmitter, that seems too ambitious for this PR. By providing the callback in the context, importers (and eventually custom functions) can wrap the callback in order to perform cleanup operations. For example:

```js
sass.render({
  ...
    var xnxx = this;
    // possible thanks to the context object
    if (!this.xnxxInitialized) {
      this.xnxxInitialized = true;
      this.xnxxPreviousCB = this.callback;
      this.callback = function() {
        // do cleanup of any caches and stateful information
        // pass off to original callback
        xnxx.xnxxPreviousCB.apply(this, arguments);
      };
    }
  }
  ...
```

This pattern can be used for importer developers looking to provide any sort of stateful importers with the most straightforward example being `importOnce`

**test slowness** This one is perplexing me and I could use some guidance. I keep seeing wild variance in the test times. Sometimes locally I'll sail through all 673 tests. Other times [like this](https://travis-ci.org/sass/node-sass/jobs/52603313) I melt down at a random point with a uv issue. Still other times, I've seen the CLI time out or take 500ms+ to complete a test. I don't think it's any code I've introduced, but I also don't have a good way to verify that.

I'll be out of town March 2-9 with limited access to a computer. I'll make every effort to be reachable though to update this PR based on feedback. Thanks again @am11 for your help in reviewing this.

lib/index.js
```diff
@@ -140,19 +140,14 @@ function getOptions(options, cb) {
   options.style = getStyle(options) || 0;
 
   // context object represents node-sass environment
-  options.context = { options: options };
-
-  var error = options.error;
-  var success = options.success;
+  options.context = { options: options, callback: cb };
```
am11: Would this not overwrite the one your are setting in render.js? `renderOptions.context = { .. };`

lib/index.js
```diff
@@ -140,19 +140,14 @@ function getOptions(options, cb) {
...
 
+  // options.error and options.success are for libsass binding
```
am11: :+1:

lib/render.js
```diff
@@ -86,9 +86,20 @@ module.exports = function(options, emitter) {
...
+  var renderCallback = function(err, result) {
+    if (err) {
+      error(err);
+    }
+    else {
+      success(result);
+    }
+  };
+
+  renderOptions.context = { options: renderOptions, callback: renderCallback };
```
am11: This will be overwritten, so we should remove it?

test/api.js
```diff
@@ -401,27 +370,8 @@ describe('api', function() {
...
-        done();
-      });
-    });
-
-    it('should throw error status 1 for bad input', function(done) {
-      sass.render({
-        data: '#navbar width 80%;'
-      }, function(err) {
-        assert(err.message);
-        assert.equal(err.status, 1);
```
am11: :+1: for removing the special tests for njs cb style. :)

README.md
```diff
@@ -165,22 +140,16 @@ sass.render({
   file: '/path/to/myFile.scss',
   data: 'body{background:blue; a{color:black;}}',
   success: function(result) {
```
am11: We should just remove the old usage from the example.

am11:
Thanks for the contribution. :+1: 

It all looks fine to me. Please review the comments for minor changes.

Regarding tests flakiness:

As you know, this is due to the way our binding code is written, which is re-structured in the other WIP PR (#644). However, I restarted the tests and it is passing all tests on TravisCI (~2 min 30 sec for each job). 

For TravisCI, it is the time of the day when you are running the tests. If you run at busy hours, it fails tests intermittently, again due to the way our binding code is written and mocha hung-ups.

lib/index.js
Jakobo: It would. My error to fix. Thanks.

README.md
Jakobo: :+1:

Jakobo:
Unrelated, but worth mentioning. As part of the `3.x` series, should we consider creating a `CHANGELOG.md` to document API changes for people consuming node-sass? I know semver goes a long way in helping, but as the second round of major changes, it may be a good idea.

am11:
Thanks for making the updates and rebasing. :+1: 

LGTM. :shipit: 

> a  `CHANGELOG.md`

I think this is a brilliant idea. I am planning to split `README.md` usage part into a Wiki, so users can related with node-sass API documentation segregated in sections (just like we relate to LibSass' API doc), as opposed to one long scrolling readme.

Some people suggest it is a good idea to have docs in repo, so to get version control (which GH Wiki is provisioned as well, but is not included in the repo). For that matter, `CHANGELOG` would suffice that need.

Issue #711, opened 2015-02-26
result.css should probably be a UTF-8-encoded buffer, not a string


joliss:
When non-ASCII characters occur in the output, libsass helpfully emits a `@charset` at-rule. For example,

```sass
p::before { content: "Hëllö world!"; }
```

generates

```css
@charset "UTF-8";
p::before {
  content: "Hëllö world!"; }
```

However, node-sass's `result.css` is a string (a sequence of Unicode characters), so it is up the user of node-sass to decide the encoding. UTF-8 is a common default, but if the user decided to a different encoding (like latin1 or UTF-16), the `@charset` at-rule would be wrong and the file would stop working.

```js
fs.writeFileSync('out.css', result.css, { encoding: 'utf16le' }) // override Node's default of 'utf8'
```

So because we're explicitly declaring that the file is UTF-8, we should actually go ahead and convert it to a Buffer (a sequence of bytes) using the UTF-8 encoding. In other words, what I'm suggesting node-sass should do boils down to this:

```js
function renderSync(...) {
  ...
  res.css = new Buffer(res.css, 'utf8') // encode (this is a hack, of course)
  return res
}
```

Regarding `result.map`: The [source map spec](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1) (is this up-to-date?) says: "For simplicity, the character set encoding is always UTF-8." So it might make sense to make `result.map` a UTF-8-encoded buffer as well.

am11:
@joliss, thanks for sharing your thoughts. However, IMVHO it is node-sass API's implementer responsibility to make sure they are using correct `utf8` encoding.

The default encoding of Buffer is `utf8`. The consumer can instantiate a Buffer if they want even with a slimmer version: 

```javascript
var sass = require('node-sass');
var result = sass.renderSync({data: '.test { display: none }' });
var myPreferredCssObj = new Buffer(result.css);  // utf8 encoded Buffer
var myPreferredMapObj = new Buffer(result.map);  // utf8 encoded Buffer
``` 

The reason why we decided not to write output files in first place is because we realized that saving to file-system is not the only scenario out there. Consumers might want the output directly to some stream or store in DB or send it via web service etc. 

Besides, the return values are instantiated via v8 in C++ code and we try our best to forward the raw output back to user without altering it. 

[...]

PR #729, opened 2015-03-02
API: Returns result as `Buffer`s not string

am11:
* CSS and Map is casted as Buffer from binding.<br/>
Advantages:
  * Buffers are super fast.
  * User can write them to file as is.
  * User can forward to response stream as is.
  * User can convert the result it to string as is:
    * `result.css.toString()`
    * `result.map.toString()` or<br/>
      `JSON.stringify(result.map)`
* Updates tests and CLI usage accordingly.
* Updates README to reflect the changes.

Issue URL: #711.
PR URL: #729.

//cc @joliss, @xzyfer, @kevva

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -42,10 +42,10 @@ module.exports = function(options, emitter) {
     };
 
     if (!options.dest || options.stdin) {
-      emitter.emit('log', result.css);
+      emitter.emit('log', result.css.toString());
 
       if (options.sourceMap) {
-        emitter.emit('log', result.map);
+        emitter.emit('log', result.map.toString());
       }
 
       return done();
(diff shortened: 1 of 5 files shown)


lib/render.js
```diff
@@ -58,13 +58,13 @@ module.exports = function(options, emitter) {
...
 
-      fs.writeFile(options.dest, result.css, function(err) {
+      fs.writeFile(options.dest, result.css.toString(), function(err) {
         if (err) {
           return emitter.emit('error', chalk.red(err));
         }
 
         emitter.emit('warn', chalk.green('Wrote CSS to ' + options.dest));
-        emitter.emit('write', err, options.dest, result.css);
+        emitter.emit('write', err, options.dest, result.toString().css);
```
xzyfer: Should this be `result.css.toString()`?

am11: Absolutely! Nice catch! :smiley: 

Fixed by: 65619ba.

PR #741, opened 2015-03-07
fix test case

fishbar:
`getBinding()` check three file paths,   the test only remove one of them,  
and the error is rise by the `fs.renameSync`

so `jscoverage` is working fine, case now is ok
@am11 

ps: the project should using Makefile to do the `install`. `test` actions

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,3 +1,6 @@
+/*!
+ * node-sass: lib/index.js
+ */
 var fs = require('fs'),
     path = require('path'),
     util = require('util');
@@ -154,6 +157,8 @@ function getOptions(options, cb) {
  */
 
 var binding = require(getBinding());
+
+module.exports.getBinding = getBinding;
 
 /**
  * Render
--- a/test/api.js
+++ b/test/api.js
@@ -1,3 +1,6 @@
+/*!
+ * node-sass: test/api.js
+ */
 var assert = require('assert'),
     fs = require('fs'),
     path = require('path'),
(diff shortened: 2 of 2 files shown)


fishbar:
also jscoverage can be used in this way:
```
mocha -r node_modules/jscoverage -=covout=html test
```
it will generate an  html covreporter

am11:
Hello @fishbar, first of all thank you for looking into it, making this PR and to reply my email! :smiley: 

There are few points I want to highlight:
* We do not want to export `getBinding` to our API.
* The `getBinding` function will always select one (same) path out of three. Unless we recompile the binding module different configuration (which does not happen during the build).

Now my question is:
* When the following condition is returning true:

  ```javascript
  if ((err instanceof Error) && /`libsass` bindings not found. Try reinstalling `node-sass`?/.test(err)) {
  // I know this because I had console.log('blah') here when I test with Travis and it printed blah.
  ..
  }
  ```

  why will jscoverage not recognize that <code>throw new Error('`libsass` bindings not found. Try reinstalling `node-sass`?');</code> was hit?

fishbar:
print the err.stack , you will find the answer.

when i run the test (on mac), the `err` is trigger by `fs.renameSync(originalBin, renamedBin);`
because the originalBin is not exists 

and when you `require('../lib');`  at the second time, `getBinding()` find the path `build/release/bind.node`
so that is why the miss-covered line always missed.

as i mentioned in the pr comments, `getBinding()` check three diff path. in your test case, you only deal with `vender/...`

am11:
I have restructured the whole binary locating process with https://github.com/sass/node-sass/pull/743.

am11:
Apparently, you are building binary with explicit `node-gyp build` command, but we build binary with `node scripts/build` command which do more than just building the binary and always copy the binary in `vendor/` directory. This is why we both are talking about two different problems.

am11:
@fishbar i pushed this test commit: https://github.com/am11/node-sass/commit/16a1e673c87aad433c60ca57ae97d2d658fd3ef1 after running the tests, travis shows this log: https://travis-ci.org/sass/node-sass/jobs/53746796#L583. As you can see it printed `check 0`   and then entered the condition where it throws (and printed `entering 0`). But coveralls suggests it never entered there: https://coveralls.io/builds/2078602/source?filename=lib%2Fextensions.js#L126.

Issue #712, opened 2015-02-27
Use Release API for binary downloads


nschonni:
While a tag shouldn't be altered once it's been made, the GitHub release API allows files to be attached to the release https://developer.github.com/v3/repos/releases/#upload-a-release-asset
This would mean that platform specific binaries could be built and attached after the code had been tagged. This would be a solution to the recent issue of frequent io.js releases in between official tagged releases. People could just check out the latest release tag and build for the new platform to be supported.

am11:
At some point, we were gauging the options to either use gh-API or amazon S3 bucket with node-pre-gyp (#56), until this CentOS incompatible binary issue showed up.

BTW, shouldn't we use this API for uploading the binaries only? I mean the download URL will remain the same: `raw.githubusercontent.com..`, no?

nschonni:
I think the new download URL would be something like `https://github.com/sass/node-sass/releases/download/v3.0.0/darwin-x64-4.1.0.21.zip` (https://developer.github.com/v3/repos/releases/#get-a-single-release-asset) for OSX io.js 1.4 given what you found in https://github.com/sass/node-sass/issues/710#issuecomment-76313012 plus the switch back to using the V8 version as the identifier.

am11:
Nice one! So basically we can upload assets to node-sass repo, instead of having a separate node-sass-binaries repo.

I think something like this will do the trick: https://gist.github.com/am11/a4a708120f16934f5d81 (credit goes to this guy: http://stackoverflow.com/a/26680032/863980).

We can then place `upload.js` in `scripts/` directory and call it like:

```bash
node scripts/build -f
# then
node scripts/upload --user am11 --auth myGitHubOAuthToken
```
`myGitHubToken` was build from public access token section: https://github.com/settings/applications#personal-access-tokens with public-repo check marked. (on personal device, we can probably store the token in env var and `echo` it on publishing the binary.

am11:
Incidentally, will these upload assets appear in the repository?

nschonni:
I think the binaries should still stay in the separate repo, and then we just copy them to the release when the release goes live.

am11:
I agree. This way, we can take binaries contribution even after the node-sass build is released.

We can think about knitting together https://github.com/sass/node-sass/issues/720, https://github.com/sass/node-sass/issues/389, https://github.com/sass/node-sass/issues/689 and this one to form a versatile build strategy, which will allow user to specify :

[...]

PR #743, opened 2015-03-07
Build: Centralises binary naming conventions (build changes part 2)

am11:
* Delivers binary name, paths and download URL
  from lib/extensions.js.

* Allows user to set binary name as environment
  variable with `SASS_BINARY_NAME`.
  * Note: this will supersede default name.

* Allows user to set binary name as in package
  file `"nodeSassConfig": { "binary-name": "" }`.
  * Note: this will supersede default name and
    `SASS_BINARY_NAME` environment variable.

* Allows user to set binary name as parameter to
  invoke any node-sass script with
  `--binary-name` flag.
  * Note: this will supersede default name, name
    defined in package `nodeSassConfig` as well
    as the `SASS_BINARY_NAME` environment
    variable.

* This binary name will be used to construct the:
  * Binary path.
  * Binary download URL.
  * Upload URL.

* Allows user to set binary path as environment
  variable with `SASS_BINARY_PATH`.
  * Note: this will supersede default path.

* Allows user to set binary name as in package
  file `"nodeSassConfig": { "binary-path": "" }`.
  * Note: this will supersede default path and
    `SASS_BINARY_PATH` environment variable.

* Allows user to set binary path as parameter to
  invoke any node-sass script with
  `--binary-path` flag.
  * Note: this will supersede default path, path
    as well as the `SASS_BINARY_PATH` environment
    variable.

* This binary path will be used when:
  * Requiring node-sass package.
  * Downloading binary.
  * Uploading binary.

* Wraps all extensions in `process.sass`
  namespace.

Issue URL: #712.

//cc @nschonni, @xzyfer


test/api.js
```diff
@@ -104,20 +104,16 @@ describe('api', function() {
...
-        var resolved = require.resolve('../lib');
-        delete require.cache[resolved];
-
         fs.renameSync(originalBin, renamedBin);
-        // try to re-require it
-        require('../lib');
+        process.sass.getBinaryPath(true);
       }, function(err) {
+        fs.renameSync(renamedBin, originalBin);
+        console.log(err);
```
nschonni: Was the `console` just for troubleshooting?

am11: It certainly was. Nice catch! :smile:

nschonni:
@am11 what do you think about using a `nodeSassConfig` config element inside the `package.json` instead of the environment variables? EX: http://jshint.com/docs/

am11:
@nschonni, I think it is a neat idea. Environment variables are bit over for this kind of a thing. But should we keep them anyway?

Precedence order (example):

- Process arguments `--binary-path ..`.
- `"nodeSassConfig": { "binary-path": ".." }`.
- Environment variable `SASS_BINARY_NAME`.

This way we will still not be the only (crazy) ones giving so many options, because we have node-jscs package for that! :wink:  (they have like 50 ways to provide settings)

am11:
@nschonni, I have added the support for `nodeSassConfig.binaryName` and `nodeSassConfig.binaryPath`.

PTAL.

Issue #758, opened 2015-03-13
Command line compile throwing ": No such file or directory" - v3.0.0-pre


lennym:
Using the command line tool as part of an npm script fails with an error of `: No such file or directory` for even the simplest scss files.

Using Ubuntu 14.04, node v.0.10.29

The same command with v2.0.1 works fine.

Test case: https://github.com/lennym/node-sass-testcase

callumacrae:
I'm getting a similar error, but I'm getting `env: node\r: No such file or directory`

am11:
Is your `test.scss` in the current directory?

I tested as follow:

```cmd
C:\temp>notepad test.scss

C:\temp>type test.scss
div {
    color: red;
}

C:\temp>node_modules\.bin\node-sass -v
node-sass       3.0.0-pre       (Wrapper)       [JavaScript]
libsass         3.1.0           (Sass Compiler) [C/C++]

C:\temp>node -v
v0.10.36

C:\temp>node_modules\.bin\node-sass test.scss test.css
Rendering Complete, saving .css file...
Wrote CSS to C:\temp\test.css

C:\temp>type test.css
div {
  color: red; }

C:\temp>
```

callumacrae:
Ah, I thought you might be using Windows.

Have you got access to a non-Windows system to test this one? I'm using OS X, and @lennym is using Ubuntu 14.04. I suspect it's something to do with the \r in my error.

lennym:
@am11 Yes, the code in lennym/node-sass-testcase is exactly what I'm running.

I've pushed it to travis to test on a couple of other environments too, and it's failing in the same way: https://travis-ci.org/lennym/node-sass-testcase/builds/54256968

am11:
Ok on Ubuntu 12.4. :)

```bash
vagrant@precise64:/tmp$ uname -a
Linux precise64 3.11.0-15-generic #25~precise1-Ubuntu SMP Thu Jan 30 17:39:31 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

vagrant@precise64:/tmp$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 12.04.4 LTS
Release:        12.04
Codename:       precise

vagrant@precise64:/tmp$ nvm install v0.10.29
######################################################################## 100.0%
Now using node v0.10.29

vagrant@precise64:/tmp$ nano test.scss

vagrant@precise64:/tmp$ cat test.scss
div {
[...]

PR #760, opened 2015-03-13
Repo: Adds .gitattributes

am11:
Issue URL: #758.

//cc @mgreter

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,35 @@
+# Set default behaviour, in case users don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files we want to always be normalized and converted
+# to native line endings on checkout.
+
+# Declare files that will always have LF line endings on checkout.
+*.c text eol=lf
+*.css text eol=lf
+*.cpp text eol=lf
+.editorconfig text eol=lf
+.gitattributes text eol=lf
+.gitignore text eol=lf
+.gitmodules text eol=lf
+*.gyp text eol=lf
+*.h text eol=lf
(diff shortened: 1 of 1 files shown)


kevva:
This feels like something that should be set on the user level imo.

am11:
@kevva, it shouldn't be part of repo?

kevva:
Well, ideally, contributors would have this setup in their editors, but I guess that's not always the case.

xzyfer:
http://editorconfig.org/ might also be worth a look

xzyfer:
:+1:

mgreter:
@kevva IMO it's a must to have a `.gitattributes` file for projects that are intended to work across multiple architectures. The file @am11 used as the base here has been in use by me since a few years, and I didn't just add it for the "fun" of adding it! The libsass team faces the same issue with the spec tests, since they currently are checked out either with `\r\n` or `\n`, according to the running OS. The spec-runner currently needs to normalize these linefeeds before comparing the results. In case of libsass this is actually desired, since we should use `\r\n` as the default linefeed when running on windows!

am11:
@mgreter, forgot to add `node-sass` (the main node entry-point for CLI) itself. :)

@xzyfer, the `.editorconfig` is already setup for this repo. In VS, it even gives me intellisense in .editorconfig file due to this https://github.com/editorconfig/editorconfig-visualstudio/pull/12! It seems like git was normalising all files because lack of `.gitattributes`. I think npm can also use something like `.npmattributes` to address https://github.com/npm/npm/issues/2097.

Issue #744, opened 2015-03-09
node-sass + libsass source coupling


saper:
For better packaging I would like to have a complete source of the project in a single zip file.
Normally this is not a problem in Github, since I can simply pulls the zip file of every commit:
(e.g. https://github.com/sass/node-sass/archive/v2.0.1.zip)

Due to our use of submodules, however, the zipfile does not include a proper version of `libsass` as well as test specs are missing (I normally do not need them).

There are few possible solutions:

1. include `libsass` in our git and have one .zip file including libsass and our node bindings.

2. download two zipfiles separately, e.g. https://github.com/sass/node-sass/archive/v2.0.1.zip and https://github.com/sass/libsass/archive/3.1.0.zip

3. teach `node-sass` to use an out-of-tree libsass (use extracted sources or use a library)

My preferred solution would be 3, is that feasible?

am11:
Thanks for your proposal.
This is similar to https://github.com/sass/node-sass/issues/389, which is actively under consideration at https://github.com/sass/node-sass/issues/712 and ultimately tackled by https://github.com/sass/node-sass/pull/743. Lets discuss it at #712.


The idea really is to make user point to any libsass commit and recompile the binary. The steps which our script might follow would be:

```bash
# in node-sass root directory
#
# when user will run `node scripts/update`
#
# read the value of npmSassConfig.libsass
# from package.json, which is the SHA1 commit hash
# to libsass. Then:
git clone https://github.com/sass/libsass
git reset --hard my_favourite_SHA1_commit_hash
npm update    # installs missing dev dependencies
node scripts/build -f
```

I think people can do all that today manually. This little automation will not bringing an extraordinary advantage anyway..


---

> Due to our use of submodules, however, the zipfile does not include a proper version of  libsass 

What do you mean, can you expand on it? The packaged libsass code is always pointing to the version of libsass we are releasing node-sass against.

> as well as test specs are missing

Yes and new we have removed the entire `test` directory from the production package for vNext. User can always use `git` to pull it. The idea really is to keep the package size as slim as possible.

Now that you have mentioned, I will remove libsass from package as well. :wink:

saper:
I read #712 and #743 and came to the conclusion they are about *binary* package downloads, i.e. package with an already compiled binding file. 

What most packagers need is just a version number and a source file corresponding to this version.

[...]

Issue #389, opened 2014-08-17
Expose config param to allow specific libsass commit for build


nschonni:
Although we ship prebuilt binaries, it may be beneficial to allow users to rebuild their own bindings to a specific version of libsass. The use case I'm seeing with this is for those that want the latest libsass, but not require us to re-release and prep for a non-tagged libsass release.

Maybe something like this in the package.config
```json
{
...
  "libsass": "commitHash"
}
```
I lean towards a config value rather than a `.node-sass.yml`, although that would be a valid approach too.

@andrew thoughts?

am11:
At present, if end-user wants to get the binary out of the desired libsass and given they have compiler GCC 4.6+ (or VC 2013+) and python v2.7 installed, they can:

>
1 - first install node-sass:
>
```bash
cd /tmp
npm install node-sass
cd node_modules/node-sass
>
# the following is to overcome the bug in v2.0.1, as pangyp (temporary fork of node-gyp
# supporting io.js) was mistakenly added as devDependency. This has been mitigated:
npm install pangyp
```
>
2 - then update upstream submodule:
>
```bash
cd src/libsass
git remote add libsass https://github.com/sass/libsass
git fetch libsass
git checkout <my-favourite-hash>
cd ../..
```
>
3 - rebuild binary:
>
```bash
node scripts/build --force
# force flag tells the script to ignore the presence of existing binary
```

Are we aiming to provision automating step 2 here; replacing `<my-favourite-hash>` with `require('package').libsassHash` (since `libsass` in now reserved for libsass version in our package.json)?

So user can do:

```bash
# pwd: node-sass root directory
nano package.json
# updated the value of libsassHash with desired hash, save and close editor, then
node scripts/build --force --update
# where --update would be the new flag which would tell the build script to: git checkout <hash>
```

xzyfer:
I believe this use case has been addressed by the `LIBSASS_EXT ` flag introduced in https://github.com/sass/node-sass/pull/763.

You can now supply the on disk location of the Libsass you wish to use with

```
LIBSASS_EXT=/path/to/libsass npm install node-sass
```

saper:
There are two ways you can specify the external library:

    LIBSASS_EXT=auto

to detect library using `pkg-config`

or

    LIBSASS_EXT=yes
    LIBSASS_CFLAGS=-I<path to sass.h header>
    LIBSASS_LDFLAGS=-L<path to libsass.so or libsass.a library>
[...]

Issue #139, opened 2013-08-08
Check for other libsass installations before installing from binary


farneman:
It's very helpful that your're providing binaries to install libsass. However I'm curious if it would be feasible to hcekc for other installations of libsass before installing from the included binary? For instance checking against popular package management systems like Homebrew that can install libsass. There maybe other complications (such as version mismatches) that make this impossible to do reliably right now, I'm just curious.

PR #746, opened 2015-03-09
Update process spawning in coverage.js for mocha 2.2.x

danielstjules:
Bumps mocha to 2.2.1, and fixes the error highlighted in https://github.com/mochajs/mocha/issues/1585:
```
0.80s$ npm run-script coverage
> node-sass@2.0.1 coverage /home/travis/build/sass/node-sass
> node scripts/coverage.js
/home/travis/build/sass/node-sass/node_modules/.bin/_mocha: 1: /home/travis/build/sass/node-sass/node_modules/.bin/_mocha: /bin: Permission denied
npm ERR! node-sass@2.0.1 coverage: `node scripts/coverage.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the node-sass@2.0.1 coverage script.
npm ERR! This is most likely a problem with the node-sass package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node scripts/coverage.js
npm ERR! You can get their info via:
npm ERR!     npm owner ls node-sass
npm ERR! There is likely additional logging output above.
npm ERR! System Linux 2.6.32-042stab090.5
npm ERR! command "/home/travis/.nvm/v0.10.36/bin/node" "/home/travis/.nvm/v0.10.36/bin/npm" "run-script" "coverage"
npm ERR! cwd /home/travis/build/sass/node-sass
npm ERR! node -v v0.10.36
npm ERR! npm -v 1.4.28
npm ERR! code ELIFECYCLE
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /home/travis/build/sass/node-sass/npm-debug.log
npm ERR! not ok code 0
````

PR #763, opened 2015-03-15
Build with the unbundled libsass

saper:
If PKGCONFIG is set to YES in the proces
environment, use pkg-config to locate
system libsass to be linked.

This way there is no need to checkout
a src/libsass submodule at all and
the library from the package management
system can be used.

Otherwise fallback to the bundled libsass.

By default, use the bundled library.

PR: #139
PR: #389
PR: #744

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -4,56 +4,43 @@
       'target_name': 'binding',
       'sources': [
         'src/binding.cpp',
-        'src/sass_context_wrapper.cpp',
-        'src/libsass/ast.cpp',
-        'src/libsass/base64vlq.cpp',
-        'src/libsass/bind.cpp',
-        'src/libsass/cencode.c',
-        'src/libsass/constants.cpp',
-        'src/libsass/context.cpp',
-        'src/libsass/contextualize.cpp',
-        'src/libsass/copy_c_str.cpp',
-        'src/libsass/error_handling.cpp',
-        'src/libsass/eval.cpp',
-        'src/libsass/expand.cpp',
-        'src/libsass/extend.cpp',
-        'src/libsass/file.cpp',
-        'src/libsass/functions.cpp',
-        'src/libsass/inspect.cpp',
-        'src/libsass/json.cpp',
-        'src/libsass/node.cpp',
-        'src/libsass/output_compressed.cpp',
-        'src/libsass/output_nested.cpp',
-        'src/libsass/parser.cpp',
-        'src/libsass/prelexer.cpp',
-        'src/libsass/remove_placeholders.cpp',
-        'src/libsass/sass.cpp',
-        'src/libsass/sass2scss.cpp',
-        'src/libsass/sass_context.cpp',
-        'src/libsass/sass_functions.cpp',
-        'src/libsass/sass_util.cpp',
-        'src/libsass/sass_values.cpp',
-        'src/libsass/source_map.cpp',
-        'src/libsass/to_c.cpp',
-        'src/libsass/to_string.cpp',
-        'src/libsass/units.cpp',
-        'src/libsass/utf8_string.cpp',
-        'src/libsass/util.cpp'
+        'src/sass_context_wrapper.cpp'
       ],
       'include_dirs': [
-        '<!(node -e "require(\'nan\')")'
-      ],
-      'cflags!': [
-        '-fno-exceptions'
-      ],
-      'cflags_cc!': [
-        '-fno-exceptions'
-      ],
-      'cflags_cc': [
-        '-fexceptions',
-        '-frtti'
+        '<!(node -e "require(\'nan\')")',
       ],
       'conditions': [
+        ['libsass_ext == ""', {
+            'dependencies': [
+              'libsass.gyp:libsass',
+            ]
+        }],
+	['libsass_ext == "auto"', {
+	  'cflags_cc': [
+	    '<!(pkg-config --cflags libsass)',
+	  ],
+	  'link_settings': {
+	    'ldflags': [
+	      '<!(pkg-config --libs-only-other --libs-only-L libsass)',
+	    ],
+	    'libraries': [
+	      '<!(pkg-config --libs-only-l libsass)',
+	    ],
+	  }
+	}],
+	['libsass_ext == "yes"', {
+	  'cflags_cc': [
+	    '<(libsass_cflags)',
+	  ],
+	  'link_settings': {
+	    'ldflags': [
+	      '<(libsass_ldflags)',
+	    ],
+	    'libraries': [
+	      '<(libsass_library)',
+	    ],
+	  }
+        }],
         ['OS=="mac"', {
           'xcode_settings': {
             'OTHER_CPLUSPLUSFLAGS': [
@@ -67,7 +54,7 @@
             'GCC_ENABLE_CPP_RTTI': 'YES',
             'MACOSX_DEPLOYMENT_TARGET': '10.7'
           }
-         }],
+        }],
         ['OS=="win"', {
           'msvs_settings': {
             'VCCLCompilerTool': {
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -53,7 +53,15 @@ function afterBuild(options) {
  */
 
 function build(options) {
-  var arguments = [path.join('node_modules', 'pangyp', 'bin', 'node-gyp'), 'rebuild'].concat(options.args);
+  var arguments = [
+    path.join('node_modules', 'pangyp', 'bin', 'node-gyp'),
+    'rebuild',
+  ].concat(
+    [ 'libsass_ext', 'libsass_cflags',
+      'libsass_ldflags', 'libsass_library' ].map(function(subject) {
+      return ['--', subject, '=', process.env[subject.toUpperCase()] || ''].join('');
+    })
+  ).concat(options.args);
 
   console.log(['Building:', process.sass.runtime.execPath].concat(arguments).join(' '));
 
--- a/libsass.gyp
+++ b/libsass.gyp
@@ -1,0 +1,94 @@
+{
+  'targets': [
+    {
+      'target_name': 'libsass',
+      'type': 'static_library',
+      'sources': [
+        'src/libsass/ast.cpp',
+        'src/libsass/base64vlq.cpp',
+        'src/libsass/bind.cpp',
+        'src/libsass/cencode.c',
+        'src/libsass/constants.cpp',
+        'src/libsass/context.cpp',
+        'src/libsass/contextualize.cpp',
+        'src/libsass/copy_c_str.cpp',
+        'src/libsass/error_handling.cpp',
+        'src/libsass/eval.cpp',
+        'src/libsass/expand.cpp',
+        'src/libsass/extend.cpp',
+        'src/libsass/file.cpp',
+        'src/libsass/functions.cpp',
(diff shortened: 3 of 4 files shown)


scripts/build.js
```diff
@@ -53,7 +53,11 @@ function afterBuild(options) {
  */
 
 function build(options) {
-  var arguments = [path.join('node_modules', 'pangyp', 'bin', 'node-gyp'), 'rebuild'].concat(options.args);
+  var arguments = [
+    path.join('node_modules', 'pangyp', 'bin', 'node-gyp'),
+    'rebuild',
+    [ '-PKGCONFIG=', process.env.PKGCONFIG || 'NO' ].join(""),
```
am11: Use single quotes in `join()`.

binding.gyp
```diff
@@ -4,91 +4,26 @@
...
-          ]
-        }],
-        ['OS!="win"', {
-          'cflags_cc+': [
-            '-std=c++0x'
-          ]
-        }]
+        [ 'PKGCONFIG == "YES"', {
+	  'cc_flags': [
+		'<!(pkg-config --cflags libsass)',
```
am11: Indentation check.

am11:
Thanks for the contribution. :ok_hand: 

I think, we can use `LIBSASS_PKG_RELEASE` instead of `PKGCONFIG`, which is a very general name.

Also, in JS code, you can just check the presence of `process.env.LIBSASS_PKG_RELEASE` and emit `YES` and `NO`. This way user will just need to `set` (`export`) or `unset` the environment variable.

@mgreter, do we already have RPM packages available for RH distros? Any thoughts about this?

saper:
I also wonder whether to have another switch to link external static `libsass.a` library instead of the shared one.

Don't have Windows running right now, can you check the build? I am not sure about this 

          'product_prefix': 'lib',

in `libsass.gyp`

am11:
We can probably have `SASS_C_LIB_EXT` (sass-c-lib-external, does it sound reasonable?) expects string value and `SASS_C_LIB_SHARED` just need to be set (value will be ignore, its presence would matter).

Windows CI (AppVeyor) is also configured: https://ci.appveyor.com/project/sass/node-sass/build/719(their free package is slower than a tortoise, so we check back after an hour or so). Also, it will show that it fails some test (due to the [timeout issue](https://github.com/sass/node-sass/pull/746#issuecomment-80629996)), but the only thing that matters is it builds and start running tests.

am11:
AppVeyor is fine. (ignore the io.js job -- the last one -- that is due to https://github.com/iojs/io.js/issues/751).
Welcome to Windows hell! :fire:

am11:
Regarding `'product_prefix'`, I think we can just skip it? It seems like node-gyp's code has its instance defined as emtpty string ([link](https://github.com/TooTallNate/node-gyp/blob/ccf27bf13cf53973034af6002edcb76ede0bc7ec/addon.gypi#L4)). Not sure if that is relevant though..

saper:
You are right, let's drop the prefix as it generates `liblibsass.a` on Windows. We can live with `sass.a` for a bundled library on UNIX-likes.

What about this:

Variable `LIBSASS_EXT`

* `auto` - detect library and include path using `pkg-config`
* `yes` - use 
  `LIBSASS_LDFLAGS` and `LIBSASS_CFLAGS` provided by the builder/packager
* `no` - (default) use bundled sass.a

Variable `LIBSASS_STATIC` (only used if `LIBSASS_EXT!=no`)

* `yes` - link statically `libsass.a`
* `no` - (default?) link to a shared library `libsass.so`

For http://github.com/sass/node-sass-binaries binary building I will be using `LIBSASS_EXT=auto` and `LIBSASS_STATIC=yes` to make sure no dependencies will be needed for the dropped binary.

am11:
This is a sound idea. I think we can try it for beta and see the reaction from masses. :)

saper:
I have given up on `LIBSASS_STATIC`, so there will be no such option (because it would require some deeper `gyp` magic to do it properly and portably). I can achieve my goal by linking `libsass.a` directly.

saper:
Re "": too much python recently, sorry. 

I decided to lowercase gyp variables because the convention there is that uppercase are only pre-defined variables (like `OS`). And by (another) convention we use uppercase environment variables.

am11:
Thanks!

am11:
LGTM!

BTW, I am going forward with `v3.0.0-alpha.1` with few fixes including these build enhancements. Do you have anything else in mind, you need to add? You can send PR for FreeBSD binaries anytime on [node-sass-binaries](https://github.com/sass/node-sass-binaries/) repo (and update the coverage matrix in the README there).

//cc @xzyfer

xzyfer:
:+1:

saper:
Thanks! This is great.

I have run into tiny issue that prevents me from doing

    $ cd /some/fresh/directory
    $ npm install /existing/deployment/node-sass

without `npm` refetching or recompiling stuff again (not sure if we have an open PR about it).

In short: I think we should check the binary first if it is already there and then proceed with download/rebuild. What do you think? Is it worth opening another PR?

am11:
@saper, I think we don't have this feature at all:

> Installing node-sass from local directory or network path while avoiding re-download of the binary.

If you are thinking about a use-case, like we can install Ruby's gem locally (from a `.gem` file, with no internet connection), then I have to admit, I just tried it for the first time after reading your comment. But I am not sure if there is a way to determine the *source* from where the npm had downloaded the package from within our install script. Our install script runs *from* the destination *after* all the package files are placed at destination.

saper:
I was thinking on using the same test as we do now to check the downloaded binary (as a quick test).

But actually I think we should sign binaries and verify that.

am11:
:tada:
Interesting idea of signing the binaries. But I think in our mini world of Sass, we can live with simple testing on install (for now).

There are number of features we need to implement for LibSass v3.2.0. Once we get stable again, we can continue enhancing the build aspect of node-sass.

pgib:
@saper Is there a trick to getting `LIBSASS_EXT=auto` working on FreeBSD (10.1amd64)? I've installed the `textproc/libsass` port, but node-sass keeps wanting to rebuild using the bundled version.

saper:
Currently we have an old libsass in the ports. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=198489 purposes an upgrade to 3.1.0. Besides, you need to call something like

    env CC=clang CXX=clang++ LINK=clang++ GYP_DEFINES=clang=1 SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=yes npm install

In this case it builds and foes not load. I have so half-working node-sass port that is doing this.

am11:
@saper, based on the FreeBSD bug, please see this conversation (till end): https://github.com/sass/libsass/issues/382#issuecomment-82022311. 

Which means that long standing `"needs old CentOS support"` issue is fixed without us needing to apply patches to libsass code every time we need to compile it for linux with GCC4.4.7 (because with docker, CentOS 5.11 is getting GCC 4.7.x, problem solved !)

Perhaps Docker is the answer for FreeBSD ports as well, if the binaries built against old FreeBSD are forward compatible, then find a oldest-possible version of FreeBSD in docker contribs with has GCC 4.7.x support.

saper:
I think even gcc 4.6 worked for me. It would have helped if libsass compiled with gcc 4.2 :) but really instead of Docker we just could have invoked `pkg add gcc-4.6` or even `pkg add node-sass` (but this kind of implies installing globally, just like  `npm install -g`). 

On a FreeBSD system many compiler versions can live in parallel (system gcc42 or clang, gcc46, 47, 49, 5...), so for that particular problem we don't need docker or jails.

am11:
> we don't need docker or jails

This is awesome! We also have plans to release v3 binaries for Solaris. Hopefully they have the similar toolchians available .

BTW, is `nullptr` supported in 4.6? I had to apply this 4.4.7 patch: https://github.com/am11/libsass/commits/gcc-4.4.7-patch to compile v2/v3-alpha binaries, I also tested v4.6 with CentOS (v6.3 I guess, which has default tool-chain packaged with gcc46). But hey, anyway as far it is compiling with whatever version without us altering the libsass code, it is all good! (-8

am11:
Unfortunately, this is causing issues with npm install: https://github.com/sass/node-sass/issues/780#issuecomment-86259009.

```
npm install node-sass@3.0.0-beta.2
```

It builds only libsass, because npm publish ignores both binding.gyp as well as libsass.gyp, if you have more than one gyp defined in package.json (like we have now).

Apparently, npm does not allow to have more than one gyp file. 

Although it is not explicitly clear in docs, but it does mention "**a** .gyp file": https://docs.npmjs.com/misc/scripts. 

Also, I can't find a reference, if someone has tried this kind of thing before: separating C++ libs within node.js module and having multiple gyp files in node.js project. Nevertheless, it would be an interesting issue to open with npm/npm repo.

If this is only concerning the `.gyp` file, can't we just bundle both of them into one?

Otherwise we would need to revert this back..

saper:
The problem is that we are not using a built-in npm facitlity to build bindings via gyp files directly - because we are using `pangyp` and not the default `node-gyp`. Maybe this can be tweaked in `package.json` too. We are also probably incorrectly using `install` hook for building which results with problems like `npm install <localpath>` trying to rebuild the whole thing again. We might need to switch to using the `prepublish` hook instead. I am not very well versed in the npm lifecycle so I need to dig more to understand this. http://github.com/mapbox/node-pre-gyp seems to be using `install` lifecycle script too and they might be running into similar issues.

But I think that d89b901212acaa7009dc7dec6dc9d1b9c7c45220 solves the original problem. Those two .gyp files are chained together if necessary.

I have built a pretty nice `autoconf` structure to reduce a number of C++11 features required to build http://github.com/sass/libsass, so I really need this feature to allow using libsass that was configured using the `./configure` script before building. This currently cannot be ported easily to `node-gyp` so we will need to rely on a relatively complete C++11 compiler to build a bundled libsass; but soon autoconf will allow us to compile libsass on pretty ancient GCC editions, too - but this has to be a standalone build (i.e. not a part of building node-sass).

am11:
> because we are using  pangyp  and not the default  node-gyp 

Which is because io.js is not supported by node-gyp yet.

> Maybe this can be tweaked in  package.json  too

I am all for such a tweak which we can handle at our part, rather shifting the problem to the consumer domain.

> But I think that  d89b901  solves the original problem.

Actually it did not solved any problem, but presented another flavour of same problem. 

> I have built a pretty nice  autoconf  structure to reduce ...

That is all good stuff. We are fixing that issue differently using newer compiler with older glibc. But all that is besides the point. *If npm publish is not equipped for handing multiple gyp files, any such cool fancy feature would have to wait unless this limitation is lifted.*

To recap:
The npm publish fails to package more than one `.gyp` files. Between binding.gyp and libsass.gyp, it took the later one, renamed it to binding.gyp and used it on install (don't ask me why). Even if npm folks fix this issue today, which is highly unlikely, it will not be landed for all the node-sass consumers using older npm. We need to find reference to at least one package, where they are using multiple gyp files in the project Or we can open this issue with npm/npm and hope someone prescribe a quasi-official workaround.

For right now, I am reverting this back for 3.0 beta and hoping it to come back as soon we have publishable solution ready.

saper:
Can you show me how to reproduce this? Nothing in the `npm`, `node-gyp` or `pangyp` code touches `binding.gyp` like this...

Issue #601, opened 2015-01-04
cmd-line: Recursive option doesn't work as expected


egtann:
Tested with node-sass@1.2.3 and @2.0.0-beta, which both seem to have the issue.

Steps to reproduce:

    $ npm install -g node-sass
    $ node-sass -r app/styles OR ./app/styles OR app/styles/ OR ./app/styles/
    > Error: File to read not found or unreadable: app/styles/

    $ node-sass app/styles/*.scss
    > Works fine
    
Expected behavior: I should be able to pass in a folder with the recursive option.

Edit: I see now that --recursive is just for watching files. You might consider making it apply to inputs as well, which is the expected behavior from normal cmd-line tools (cp, chmod, etc.).

kevva:
You might need to use it along with the `--watch` flag.

egtann:
Right, but this is part of a package.json build script, so I'm not watching the files.

Watch and recursive should probably be independent options. If not, perhaps the docs/command should be updated to the following?

    -w  Watch
    -wr  Watch recursively

kevva:
Nah, but `--recursive` should definitely work without `--watch` though.

egtann:
Also, even with *.scss, it only outputs the first file it finds and saves it as .scss, overwriting the original file.

    $ ls .
    > base.scss       list_item.scss

    $ node-sass -r *.scss -o ./
    > Rendering Complete, saving .css file...
    > Wrote CSS to list_item.scss

Notice that it outputs to .scss, overwriting my original file! Good thing I had it in a Vim buffer...

am11:
@egtann, for the original issue reported. I would consider as a feature request (to provide recursively compile all files in given directory structure) rather than a bug. :)

I have added it in 2.1 milestone. https://github.com/sass/node-sass/milestones.

The second issue you have reported is fixed. I don't have the exact commit reference handy But this is what I get with the latest build (v2.0.0-beta):

```powershell
C:\temp1> .\node_modules\.bin\node-sass --version
2.0.0-beta

C:\temp1> .\node_modules\.bin\node-sass -r \temp\foo.scss -o ./
Rendering Complete, saving .css file...
Wrote CSS to foo.css
```

Also, we do not have glob paths support as well as directory for input file yet. For watch, we certainly do: [node-sass#L170](https://github.com/sass/node-sass/blob/604ccf5894f0100d58ff6c3b4ebeb1e85f83f5ed/bin/node-sass#L170).
This will be a nice feature!

[...]

PR #782, opened 2015-03-22
Fixed watch behavior

### LINKED ISSUE #781: Compile a single file while watching the whole directory
I have been digging around in here for a while but didn't an existing issue for this.  Here is the command I am running:

```
$ node-sass --watch=sass --output-style=expanded --source-map="main.css.map" --recursive sass/main.scss -o build
```

It correctly builds the file and the source map.  But it does not watch the other files imported by `main.scss`.  I tracked down the problem to this line:

```javascript
var graph = grapher.parseDir(options.src, { loadPaths: options.includePath });
```

I believe this is supposed to be:

```javascript
var graph = grapher.parseDir(options.watch, { loadPaths: options.includePath });
```

This fixes my issue, but I don't know if this breaks other use cases.  The tests seem to be failing before making the above change, but no new failures happen when I run it with the patch.  Specifically the one that tests watched files still passes.  Other than that I am not entirely sure which manual cases to test.

Should I submit a PR for this?   Is there something I am missing about how this is working?  Thanks for any help and or direction you can point me in.

wesleytodd:
This fixes #781.  Using the watch command now finds all of the dependencies and will recompile when one of them changes.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -53,7 +53,8 @@ var cli = meow({
     'recursive',
     'source-map-embed',
     'source-map-contents',
-    'source-comments'
+    'source-comments',
+    'watch'
   ],
   string: [
     'functions',
@@ -64,8 +65,7 @@ var cli = meow({
     'output',
     'output-style',
     'precision',
-    'source-map-root',
-    'watch'
+    'source-map-root'
   ],
   alias: {
     c: 'source-comments',
@@ -82,7 +82,8 @@ var cli = meow({
     'indent-width': 2,
     linefeed: 'lf',
     'output-style': 'nested',
-    precision: 5
+    precision: 5,
+    recursive: true
   }
 });
 
@@ -139,7 +140,7 @@ function getEmitter() {
  */
 
 function getOptions(args, options) {
-  options.src = options.watch ? options.watch : args[0];
+  options.src = args[0];
 
   if (args[1]) {
     options.dest = path.resolve(process.cwd(), args[1]);
@@ -161,25 +162,19 @@ function getOptions(args, options) {
  */
 
 function watch(options, emitter) {
-  var dir = options.watch;
-  var gaze = new Gaze();
+  var glob = options.recursive ? '**/*.{sass,scss}' : '*.{sass,scss}';
+  var src = isSassFile(options.src) ? options.src : path.join(options.src, glob);
+  var graph = grapher.parseDir(path.resolve(path.dirname(src)), { loadPaths: options.includePath });
+  var watch = [];
 
-  if (dir === true) {
-    dir = [];
-  } else if (!Array.isArray(dir)) {
-    dir = [dir];
+  // Add all files to watch list
+  for (var i in graph.index) {
+    watch.push(i);
   }
 
-  dir.push(options.src);
-  dir = dir.map(function(d) {
-    var glob = options.recursive ? '**/*.{sass,scss}' : '*.{sass,scss}';
-    return isSassFile(d) ? d : path.join(d, glob);
-  });
-
-  gaze.add(dir);
+  var gaze = new Gaze();
+  gaze.add(watch);
   gaze.on('error', emitter.emit.bind(emitter, 'error'));
-
-  var graph = grapher.parseDir(options.src, { loadPaths: options.includePath });
 
   gaze.on('changed', function(file) {
     var files = [file];
--- a/test/cli.js
+++ b/test/cli.js
@@ -207,6 +207,28 @@ describe('cli', function() {
         fs.appendFileSync(src, 'body{background:white}');
       }, 500);
     });
+
+    it('should watch the full sass dep tree for a single file', function(done) {
+      var src = fixture('watching/index.scss');
+      var foo = fixture('watching/foo.scss');
+
+      fs.writeFileSync(foo, '');
+
+      var bin = spawn(cli, [
+        '--output-style', 'compressed',
+        '--watch', src
+      ]);
+
+      bin.stdout.setEncoding('utf8');
+      bin.stdout.once('data', function(data) {
+        assert(data.trim() === 'body{background:white}');
+        done();
+      });
+
+      setTimeout(function() {
+        fs.appendFileSync(foo, 'body{background:white}');
+      }, 500);
+    });
   });
 
   describe('node-sass in.scss --output out.css', function() {
--- a/test/fixtures/watching/foo.scss
+++ b/test/fixtures/watching/foo.scss
@@ -1,0 +1,1 @@
+body{background:white}
--- a/test/fixtures/watching/index.scss
+++ b/test/fixtures/watching/index.scss
@@ -1,0 +1,1 @@
+@import './foo';


test/cli.js
```diff
@@ -172,6 +172,28 @@ describe('cli', function() {
...
+      fs.writeFileSync(foo, '');
+
+      var bin = spawn(cli, [
+        '--output-style', 'compressed',
+        '--watch', src
+      ]);
+
+      bin.stdout.setEncoding('utf8');
+      bin.stdout.once('data', function(data) {
+        assert(data.trim() === 'body{background:white}');
```
am11: After this line, `fs.unlinkSync(fixtures('watching/index.css'))` and then `git rm test/fixtures/watching/index.css` to prevent it from bleeding into source-control.

wesleytodd: Fixed!

am11: Thanks! Can you please rebase into one commit?

am11:
Please elaborate, what will happen when:

* `watch = file.scss` and `args[0] = file2.scss` are set
* `watch = directory/` and `args[0] = file2.scss` are set.

wesleytodd:
`watch = file.scss` and `args[0] = file2.scss`

This use case is probably the only way you can go wrong with this.  This will recompile `file2.scss` whenever `file.scss` changes.  I cannot imagine anyone wanting that behavior, but that is the way it works now.

`watch = directory/` and `args[0] = file2.scss`

Similarly to the above, this will recompile file2.scss whenever anything in `directory/*.{scss,sass}` changes.

wesleytodd:
To be honest this conversation makes we want to change the behavior completely.  Seems like `--watch` should only be a boolean.  And `arg[0]` should only be the files to watch.  If you want me to make this PR do that instead I can.

am11:
> Seems like  --watch  should only be a Boolean

:+1: :100: 

Please go ahead.

As a side note, in case the user has used custom importer option with watch, it will `require()` that function just once. But while the file is being watched, any change to the importer function would be ignored. To fix this, I think we should [unload and re-require](http://stackoverflow.com/a/6677355/863980) the import file before the watcher requests compile-on-change.

Thoughts?

bin/node-sass
```diff
@@ -148,25 +148,30 @@ function getOptions(args, options) {
  */
 
 function watch(options, emitter) {
-  var dir = options.watch;
-  var gaze = new Gaze();
+  var watchDir = isSassFile(options.watch) ? path.dirname(path.resolve(options.watch)) : options.watch,
+      glob = options.recursive ? '**/*.{sass,scss}' : '*.{sass,scss}',
+      gaze = new Gaze(),
+      dir;
```
kevva: Use multiline `var` statements.

wesleytodd: I had that, but the top of the file was using commas, so I changed it.  Are there contribution docs on this stuff that I can read?

wesleytodd:
@am11 I will do that then.  I am not sure what the custom importers do, but I am sure I can figure it out, or if you have docs you can point me to.  Otherwise, if I have any questions I will be sure to ask them here.

am11:
@wesleytodd, you can find the usage in example: https://github.com/sass/node-sass/#examples.
When a user of node-sass set the custom importer, LibSass calls the function as it encounters any Sass `@import` (as opposed to CSS3 import directive). This way user can alter the file-path and read the data (in case the `@import` had URL -- because LibSass doesn't download files from URLs -- or even database connection string etc.) and return the altered result to LibSass to continue compilation.

Separately, in this PR, you may be able to address https://github.com/sass/node-sass/issues/601 by simply retiring `--recursive` option and make the default behavior recursive. (-8

am11:
@wesleytodd, any chance we can wrap this up before v3 stable?

wesleytodd:
Sorry, been really busy this week.  When is the v3 stable deadline?

am11:
TYT! :sunglasses:  :+1: 

Our stable release (v3.0.0) will be aligned with that of libsass v3.2.0. Going by their milestone, https://github.com/sass/libsass/milestones/3.2, we have probably few more days.

I am sorry, but we have included some new CLI features. I hope there won't be many merging issues in your feature branch.

Let us know if you need assistance in moving things forward.

wesleytodd:
Was the cli change anything that effects this feature?  The merge was clean but I haven't tested the feature yet.

I will work on getting the new behavior working tomorrow, but if there is anything else I should know about cli changes it would be good to know now :), as I have not been keeping track of the rest of this projects active issues.

wesleytodd:
After rebasing I am getting this when running the tests:

```
node-sass/lib/index.js:382
module.exports.TRUE = binding.types.Boolean.TRUE;
                                   ^
TypeError: Cannot read property 'Boolean' of undefined
```

Any ideas?

wesleytodd:
Whelp.....I made the mistake of getting io setup on my new laptop, and then found this issue: https://github.com/shama/gaze/issues/173

I asked for their recommendation, so if we cannot get node@0.12.x and io.js support we might want to consider going with something similar to this: https://github.com/latentflip/stylizer/pull/6

Thoughts?

wesleytodd:
So I am no longer sure about that gaze issue.  I got it working, I just had a few silly errors that I hadn't noticed in my code.  If these tests pass in CI then I think this is ready.

am11:
Both CI builds are passing. :+1: :tada: 

Regarding gaze, I saw that issue, but it never failed functioning with io.js et al. on any platform runtime I have used so far. I think it is because the version we are using, v0.5 wasn't relying on C++ module? It is unclear from the history of repo, but going by the code, it seems like it was pure JS module back then. Hence compatibility is not the issue.

Regarding --recursive option, do you think we can live without it, by making recursive a default behavior (and call it a fix for https://github.com/sass/node-sass/issues/601)?

wesleytodd:
I think that it makes sense to keep recursive as an option.  There is certainly a situation where you might not want to compile files deeper in the tree.  But I do think the recursive option is more common and should be default.

am11:
I agree with you. So can we can probably just set it here: (https://github.com/wesleytodd/node-sass/blob/bd710c61c76d49cd4decdb59b6a8c68bf07b91b6/bin/node-sass#L86) like this:

```javascript
..
  precision: 5,
  recursive: true
}
```

Would you like to fix it in this PR? :)

wesleytodd:
Updated.

am11:
Thanks for your contribution @wesleytodd. :tada: 

Looks good to me. :+1:

wesleytodd:
@am11 Glad I could help!!  Its a good project, so I am sure I will be using it a lot.  Maybe find some more time to help out in the future.

am11:
> Maybe find some more time to help out in the future.

Most welcome and keep coming back! :sunglasses: :+1:

am11:
Few things I noticed:

* With `meow` package, the CLI arguments parser, converts all non-`true` values to false. Meaning:
  * `node-sass blah.scss --watch --recursive false` == `node-sass blah.scss --watch --recursive untrue` == `node-sass blah.scss --watch --recursive foo`
  * while `node-sass blah.scss --watch --recursive true` == `node-sass blah.scss --watch --recursive` == `node-sass blah.scss --watch`
* You are the [100th contributor](https://github.com/sass/node-sass/graphs/contributors) to this project, so here you go: :100: :trophy: :smiley:

wesleytodd:
@am11 I believe that is the behavior from [minimist](https://github.com/substack/minimist).  This is a very commonly used package, so my guess is this behavior is fine, despite it's oddities.  We could submit an issue over there if you have a better suggestion for these behaviors.

Yeah me!  Thanks for the :trophy:!

wesleytodd:
Also, while pursuing the minimist source I notices it supports `--no-recursive`.

And here is the issue in minimist for reference: https://github.com/substack/minimist/blob/master/index.js#L158

am11:
I agree that `--no-recursive` would be better suited here, when recursive is a default option.

PR #783, opened 2015-03-22
Feature: Custom Functions

am11:
<del>Supersedes #644.</del>

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,6 +13,9 @@ require('./extensions');
 
 var binding = require(process.sass.getBinaryPath(true));
 
+// Expose sass types
+module.exports.types = binding.types;
+
 /**
  * Get input file
  *
@@ -172,13 +175,9 @@ module.exports.render = function(options, cb) {
   var importer = options.importer;
 
   if (importer) {
-    options.importer = function(file, prev, key) {
+    options.importer = function(file, prev, bridge) {
       function done(data) {
-        console.log(data); // ugly hack
-        binding.importedCallback({
-          index: key,
-          objectLiteral: data
-        });
+        bridge.success(data);
       }
 
       var result = importer.call(options.context, file, prev, done);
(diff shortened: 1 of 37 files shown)


src/binding.cpp
```diff
@@ -233,117 +184,92 @@ void make_callback(uv_work_t* req) {
     node::FatalException(try_catch);
   }
 
-  if (ctx_w->importer_callback) {
-    uv_close((uv_handle_t*)&ctx_w->async, NULL);
-  }
+  //if (ctx_w->importer_bridge) {
+  //  uv_close((uv_handle_t*)&ctx_w->async, NULL);
+  //}
```
am11: TODO: remove when everything else is fine.

am11:
This is a historical moment!

historical-moment

src/create_string.cpp
```diff
@@ -0,0 +1,12 @@
+#include "create_string.h"
```
saper: Need

      #include <stdlib.h>

for `malloc()`

lib/index.js
```diff
@@ -189,6 +188,27 @@ module.exports.render = function(options, cb) {
...
+
+      options.functions[signature] = function() {
+        var args = Array.prototype.slice.call(arguments),
+            bridge = args.pop();
+
+        function done(data) {
+          bridge.success(data);
+        }
+
+        var result = callback.apply(this, args.concat(done));
```
Jakobo: `callback.apply(options.context, args.concat(done))` for consistency with importers?

am11: Yes this is what I just told @matryo on gitter chat :)
This PR is experimental. the real meat is still #644. :wink:

Jakobo: No worries. Let me know how (if) I can help. :)

matryo: @Jakobo Yes this makes complete sense. Ultimately I can see that the context could be used to call built-in sass functions, or functions defined through the sass files.

Jakobo: Totally. We've already seen the context object be useful in `import-once` importer development, and I have no doubt that image manipulation custom functions will memoize using the `context` for performance. :)

am11: @Jakobo, wow image manipulation, like generating and mapping sprites? One day I will learn all the powers and mysteries behind `this.conext`, but I don't get time to actually *use* node-sass like others do! :smile_cat:

am11: @Jakobo, now that #644 is merged, would you like to send a PR for `this.context` and the tests which goes with it? (-8

You may want to play with this cool new feature as well. :wink:

PR #785, opened 2015-03-23
malloc() needs <stdlib.h>

saper:


am11:
Thanks fro contribution @saper. I think we should not use malloc here in first place. It should be C++ style memory allocations (`new` and `delete`).

But that can wait until later! :+1: 

Is it compiling on FreeBSD otherwise?

saper:
Yes, it compiles and even the tests pass. I also compiled libsass 3.2.0 beta2  separately and it linked fine via `LIBSASS_EXT`. 

Didn't try older versions of FreeBSD yet (and 32 bit).

PR #790, opened 2015-03-23
Workaround iojs/io.js#751 Windows iojs.exe alias

saper:


saper:
I am getting build failures on iojs:

https://gist.github.com/cacacd602e84db9dde72

am11:
@saper, there is a reason we didn't opted for this route. We shouldn't be hardcoding the executable name in the configuration file. It should work with `npm task`.

Wait for io.js to fix this issue on Windows. Meanwhile running tests locally on Windows passes all tests.

saper:
I don't believe iojs will fix iojs/io.js#751 quickly. There is no good solution in sight.

Of course this is an ugly workaround. I'd rather have tests running (somehow, not sure if we can make them run this way). 
Alternatively of course we can remove iojs from the Windows matrix...

What is `npm task`?

am11:
> What is  npm task ?
npm test

saper:
Ah, thanks. Well, changing appveyor script does not break `npm test`....

am11:
We need io.js in the matrix to observe if the binary builds fine, but it is allowed to fail later.
If you want to carry out further experiments, please do it in your fork. 

Sorry, but we are fine waiting for io.js to bring the solution to (which they think is a) trivial issue and so do we.

PR #798, opened 2015-03-25
remove extra slash to fix binary download fail

bevanhunt:
The extra slash was causing the binary download to fail - this was the commit where it was added - https://github.com/sass/node-sass/commit/2220ba96b9a20920ccfaaadc12978f40e069caf9


am11:
Thanks for you contribution.

> to fix binary download fail

I don't see how it fails with double slash. For instance:
https://github.com/sass/node-sass/releases/download//v3.0.0-alpha.1/darwin-x64-11_binding.node

The slashes in paths are always squashed to one (even on Windows backslashes behave the same way..)

bevanhunt:
Sorry, I just got confused with the binaries missing on the beta build. Cheers.

Issue #784, opened 2015-03-23
Provide CLI support for Custom Functions


am11:
Based on #644, we need Custom Functions support for CLI, on a par with API usage. 
The functions will have same behavior as importers, except for multiplicity: we can have multiple functions, but we can't have multiple importers ([yet](https://github.com/sass/libsass/issues/962)).

Options:

* Provide a single, non repeatable flag `--functions`, which accepts path value of type string. This path would be a path to `.js` file with multiple `modules.exports`. We will use `require()` and transform into arrays of functions.
* Provide a repeatable flag `--function`, which takes path to file with single or multiple `module.exports`. We will again use `require()`.

Caveat:

(as noted [here](https://github.com/sass/node-sass/pull/782#issuecomment-84652006))
We need to figure out a way to make `watcher` aware of any changes made in custom function and custom importer, after the first `require()`. We would probably need to re-require all the modules (after clearing the `require.chache['importer/functions']`).

am11:
Merged #815.

PR #815, opened 2015-03-28
CLI: Adds support for Custom Functions

am11:
* Adds related tests and fixtures.

Issue URL: #784.


bin/node-sass
```diff
@@ -227,6 +230,14 @@ function run(options, emitter) {
     }
   }
 
+  if (options.functions) {
+    if ((path.resolve(options.functions) === path.normalize(options.functions).replace(/(.+)([\/|\\])$/, '$1'))) {
```
am11: This is indeed ugly, but without path resolution, user would need to `./` (or `.\` in Windows) before the relative paths.

PR #818, opened 2015-03-29
Build: Adding libsass.gyp in a separate dir

am11:
* Reintroduces #763.
* Turned out it isn't a bug, it's a feature!
  * npm/npm#7780

//cc @saper.


saper:
Thanks! Yes, I can build with external `libsass` now without problems. (this time against 3.2.0beta1+)

am11:
Yeah, just waiting for libsass beta.next. I will add this feature: https://github.com/sass/libsass/pull/1000#issuecomment-87360514 in the next beta.

Separately, do you think you can arrange FreeBSD binaries for existing beta https://github.com/sass/node-sass-binaries/pull/50?

saper:
Ok, working on it!

saper:
Regarding packaging: it would be really good for things like "npm install <someotherdir>" to include an already build binary in the npm package. But that kind of introduces two kinds of package tarfiles - source only and source + specific binary only.

am11:
Great thanks! :)

Remember, [nvm](https://github.com/creationix/nvm/) is your friend for fast version switching. Also, now the Dockerfile is promising for Linux binaries: https://gist.github.com/am11/1529b2458b1a1f4bdba8, would be nice if we have similar recipe for FreeBSD and solaris too. Related: https://github.com/sass/node-sass-binaries/issues/32#issuecomment-87361249 (I tired, but can't configure solaris VM..).

saper:
My recipe is to run:

````
for a in i386 amd64
do
        for j in 8_4 9_3 10_1
        do
               for e in nodejs iojs nodejs010
                do
                        poudriere bulk -j "${e}_${j}_${a}" -p exp textproc/node-sass
                done
        done
done
````

am11:
Yes we used to bundle binary in the before node-sass-binaries repo, but turned out separating binaries from package is a good practice, which most of the packages using add-ons have adapted.

One reason I remember was that the package is a distilled form, and we cannot cherry pick the platform specific binary at install time (altering the package on the fly). So it would be a real bulky package if we bundle all binaries in the package.

With current approach, we can add binaries for specific release even after publishing.

am11:
Looks nice. Our current matrix requirement is:

* node.js v0.10.38 (ia32, x64)
* node.js v0.12.1 (ia32, x64)
* io.js v1.0.4 (ia32, x64)
* io.js v1.6.2 (ia32, x64)

Can you release 32-bit bins as well?

am11:
Perhaps this is worth giving a shot: https://registry.hub.docker.com/u/kazuyoshi/freebsd-minimal/tags/manage/ (FreeBSD docker container). 

Our recipe files will eventually be added to node-sass-binaries repo.

For Windows, I will make a `bat` or `ps1` script.

am11:
@saper, how are the FreeBSD builds coming along? Feel free to send a PR at node-sass-binaries like https://github.com/sass/node-sass-binaries/issues/32. :)

saper:
Very good, just need to restore old FreeBSD iojs 1.0.4 port out of the dustbin, since all others are ready.

am11:
:+1:

PR #830, opened 2015-04-02
ad br so readme cli options differentiate

tomfuertes:
`<span/><span/>` to more `<p/><p/>`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,6 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = space
 indent_size = 2
+
+[*.md]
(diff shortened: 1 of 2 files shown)


kevva:
Why? It simply shows the output of `node-sass --help`.

tomfuertes:
two chars at end of line gets overridden often w/ https://github.com/sass/node-sass/blob/master/.editorconfig#L9

tomfuertes:
why

kevva:
Then we should add this to the .editorconfig:

```
[*.md]
trim_trailing_whitespace = false
```

tomfuertes:
forgot that was a thing! done / squashed

Issue #821, opened 2015-03-30
Feature: Implement multiple Custom Importers support


am11:
Corresponds to https://github.com/sass/libsass/pull/1000.

am11:
Removed <kdb>Breaking Change</kdb> tag, because I will try to implement it in a backward compatible way (although libsass API marked it as a breaking change).

am11:
On second thought, instead of providing multiple variants:

(Note for priority precedence set by libsass: highest first)

```javascript
// variant 1
importer: function(url, prev, done) {}
// variant 2
importer: { priority: 1, func: function(url, prev, done) {} }
// variant 3
importer: [{ priority: 1, func: function(url, prev, done) {} }. function(url, prev, done) {}, {priority: 100, func: function(url, prev, done) {}}
// etc.
```

How about aligning `importers` (plural) with custom `function` (and retiring `importer` singular):

```javascript
importers: {
  10: function(url, prev, done) {},
  9: function(url, prev, done) {},
  2: function(url, prev, done) {}
},
headers: { // custom headers, upcoming feature: #822
  4: function(/* not sure yet */) {},
  1: function(/* not sure yet */) {}
},
functions: {
  'foo1': function() {},
  'bar1($a)': function(a) {},
  'foo2($d, $c)': function(a, b) {},
  'bar2($apa = \'alpha\', $bao = \'bravo\')': function(x, y) {}
}
```

with proper validation messages. This will be another breaking change of v3. :)

Thoughts? 

//cc @xzyfer

xzyfer:
I prefer the latter on _right way_ approach.

I presume the keys map to priority, identifier, and signature respectively? i.e. `headers` object keys should be strings not numbers.

am11:
Agree for importers.

I haven't looked into `headers` yet. @mgreter said it has similarities to importers, so I thought it would have priority attribute. But seems like it is more inclined towards functions. Lets discuss `headers` separately over #822. :)

jhnns:
Thx for letting me know.

I'm a little concerned that things might get too difficult on your side because there are so many possibilities to use your API. But as long as you can handle it... :grinning:

Snugug:
[...]

PR #832, opened 2015-04-03
API: Provides support for array of importers

am11:
* In order to skip the importer, user must
  return `sass.types.NULL` (or the shorter
  alias `sass.NULL`).
* See the added test on usage.
* Backward compatible: part of me want still
  wants to make it non-backward compatible
  because:
  * We can: in major version v3.0.
  * It will keep the API clean.
  * It will keep the docs clear: type of
    `options.importers` is array of functions.
* Updates docs.

Issue URL: #821.


am11:
@mgreter, @xzyfer, PTAL.

Also, thoughts on that breaking change note?

am11:
@mgreter, going by perl-libsass and [C-API documentation](https://github.com/sass/libsass/wiki/API-Sass-Importer#return-imports), what is the purpose of last parameter `srcmap`?

mgreter:
@am11 if the loaded content has a sourcemap accociates, you should return the content of the sourcemap in that variable. We don't do anything with it yet, but in the future libsass should remap the resulting css to the provided source-map if possible.

am11:
@mgreter, perfect! I will expose that feature to node-sass API in this PR then. :)

mgreter:
@am11 any ETA when you will release the next beta, so libsass can do another beta round too?

am11:
@mgreter, I am almost ready. Can you please release a beta with the `Sass_Compiler` fixes, so our submodule points to libsass release, instead of some random commit hash? I will wind up the two open PRs soon and publish the next beta.

mgreter:
@am11 lets move that discussion to https://github.com/sass/libsass/issues/1020 :wink:

am11:
I have added the `srcmap` but lets keep it undocumented, since it doesn't have any effect on generated source-map.

I will do the final rebase once libsass next rc/beta is released..

PR #835, opened 2015-04-04
Fix #775: Introduce SASS_BINARY_SITE environment variable

### LINKED ISSUE #775: Impossible to mirror the libsass binaries

```javascript
function getBinaryUrl() {
  return flags.binaryUrl ||
}
```
and 
```javascript
function getBinaryName() {
  var binaryName;

  if (flags.binaryName) {
  } else {
                  process.arch, '-',
  }

}
```

saper:
Provide ability to locally mirror node-sass
binaries for various versions and platforms.

SASS_DIST_SITE needs to be an URL pointing to
a collection of files organized like the Github
repository.

If SASS_DIST_SITE is set to

 http://myhost:8080/local/node-sass-bin

then

 http://myhost:8080/local/node-sass-bin/v3.0.0-beta.4/freebsd-x64-14_binding.node

should point to the FreeBSD 64 bit binary for node 0.12.0

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -50,7 +50,7 @@ function getRuntimeInfo() {
 /**
  * Get binary name.
  * If environment variable SASS_BINARY_NAME or
- * process aurgument --binary-name is provide,
+ * process argument --binary-name is provided,
  * return it as is, otherwise make default binary
  * name: {platform}-{arch}-{v8 version}.node
  *
@@ -62,10 +62,10 @@ function getBinaryName() {
 
   if (flags['--sass-binary-name']) {
     binaryName = flags['--sass-binary-name'];
-  } else if (package.nodeSassConfig && package.nodeSassConfig.binaryName) {
-    binaryName = package.nodeSassConfig.binaryName;
   } else if (process.env.SASS_BINARY_NAME) {
     binaryName = process.env.SASS_BINARY_NAME;
+  } else if (package.nodeSassConfig && package.nodeSassConfig.binaryName) {
+    binaryName = package.nodeSassConfig.binaryName;
   } else {
     binaryName = [process.platform, '-',
                   process.arch, '-',
@@ -76,21 +76,40 @@ function getBinaryName() {
 }
 
 /**
- * Retrieve the URL to fetch binary.
- * If environment variable SASS_BINARY_URL
- * is set, return that path. Otherwise make
- * path using current release version and
- * binary name.
+ * Determine the URL to fetch binary file from.
+ * By default feth from the node-sass distribution
+ * site on GitHub.
+ * 
+ * The default URL can be overriden using
+ * the environment variable SASS_BINARY_SITE
+ * or a command line option --sass-binary-site:
+ *
+ *   node scripts/install.js --sass-binary-site http://example.com/
+ *
+ * The URL should to the mirror of the repository
+ * laid out as follows:
+ *
+ * SASS_BINARY_SITE/
+ * 
+ *  v3.0.0-beta.4
+ *  v3.0.0-beta.4/freebsd-x64-14_binding.node
+ *  ....
+ *  v3.0.0-beta.5
+ *  v3.0.0-beta.5/freebsd-ia32-11_binding.node
+ *  v3.0.0-beta.5/freebsd-x64-42_binding.node
+ *  ... etc. for all supported versions and platforms
  *
  * @api private
  */
 
 function getBinaryUrl() {
-  return flags['--sass-binary-url'] ||
-         package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null ||
-         process.env.SASS_BINARY_URL ||
-         ['https://github.com/sass/node-sass/releases/download/v',
-          package.version, '/', sass.binaryName].join('');
+  var site = flags['--sass-binary-site'] || process.env.SASS_BINARY_SITE  ||
+             (package.nodeSassConfig ? package.nodeSassConfig.binarySite : null);
+  if (site) {
+		  return [site, 'v' + package.version, sass.binaryName].join('/');
+  } else {
+          return null;
+  }
 }
 
 /**
@@ -118,7 +137,7 @@ sass.versionInfo = getVersionInfo();
 /**
  * Get binary path.
  * If environment variable SASS_BINARY_PATH or
- * process aurgument --binary-path is provide,
+ * process argument --sass-binary-path is provided,
  * select it by appending binary name, otherwise
  * make default binary path using binary name.
  * Once the primary selection is made, check if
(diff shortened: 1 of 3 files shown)


am11:
We already have `binaryUrl` in `nodeSassConfig` for this purpose.

lib/extensions.js
```diff
@@ -89,8 +89,9 @@ function getBinaryUrl() {
   return flags['--sass-binary-url'] ||
          package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null ||
```
am11: You can augment this with ['v', package.version, '/', sass.binaryName].join('')`

saper: Sure we can; but this changes the meaning of `SASS_BINARY_URL` variable.

If somebody set it like this

````
SASS_BINARY_URL=http://somehost:8080/local/freebsd-x64-14_binding.node
````

it will no longer work, instead they would need to re-organize their files under `'v'` + `package.version` to make this work again.

We can deprecate the `SASS_BINARY_URL` variable if we want, but let's give a new one a new name and let's warn users that the old one will no longer work.

This can be very confusing when somebody is upgrading a larger set of machines with `npm`...

Comment (am11) [inline lib/extensions.js]: Agree. Then lets replace `sassBinaryUrl` with sassBinarySite (or preferably `sassBinaryCdn` as per the original requester https://github.com/sass/node-sass/issues/720) with the following precedence order:

* CLI argument `--sass-binary-cdn`
* `nodeSassConfig.sassBinaryCDN`
* `process.env.sassBinaryCDN`

If none of them are found, it will fall back to default github URL.

Comment (saper) [inline lib/extensions.js]: I think we should once migrate to something like https://github.com/dominictarr/rc or https://github.com/npm/npmconf so I'd love to have some consistent config variable names and camelCase makes it a bit difficult...

Comment (am11) [inline lib/extensions.js]: @saper please read my previous comment again.

Comment (saper) [inline lib/extensions.js]: Yes, I read. Btw. --sass-binary-url didn't work at all, it caused an exception (in current master).

File: lib/extensions.js [hunk 2]
```diff
@@ -81,16 +81,43 @@ function getBinaryName() {
...
+ * ... etc. for all supported versions and platforms
  *
  * @api private
  */
 
 function getBinaryUrl() {
   return flags['--sass-binary-url'] ||
-         package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null ||
+         (package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null) ||
          process.env.SASS_BINARY_URL ||
```
am11: Now that we have sassBinarySite, we do not need sassBianryUrl and the test in api.js that goes with it.

lib/extensions.js
```diff
@@ -81,16 +81,43 @@ function getBinaryName() {
...
+ *
+ * ... etc. for all supported versions and platforms
  *
  * @api private
  */
 
 function getBinaryUrl() {
   return flags['--sass-binary-url'] ||
-         package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null ||
+         (package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null) ||
```
am11: Brace was unnecessary.

lib/extensions.js
```diff
@@ -81,16 +81,43 @@ function getBinaryName() {
...
          process.env.SASS_BINARY_URL ||
-         ['https://github.com/sass/node-sass/releases/download/v',
-          package.version, '/', sass.binaryName].join('');
+         [
+           (
+             flags['--sass-binary-site'] ||
+             (package.nodeSassConfig ? package.nodeSassConfig.binarySite : null) ||
```
am11: Brace can be removed.

lib/extensions.js
```diff
@@ -81,16 +81,43 @@ function getBinaryName() {
...
+ *  v3.0.0-beta.4/freebsd-x64-14_binding.node
+ *  v3.0.0-beta.5
+ *  v3.0.0-beta.5/freebsd-ia32-11_binding.node
+ *  v3.0.0-beta.5/freebsd-x64-42_binding.node
+ *  v3.0.0-beta.5/freebsd-x64-14_binding.node
+ *  v3.0.0-beta.5/freebsd-x64-11_binding.node
+ *  v3.0.0-beta.5/freebsd-ia32-43_binding.node
+ *  v3.0.0-beta.5/freebsd-ia32-42_binding.node
+ *  v3.0.0-beta.5/freebsd-ia32-14_binding.node
+ *  v3.0.0-beta.5/freebsd-x64-43_binding.node
```
am11: Two examples would be enough:

```
http://example.com/my-path/
  v3.0.0-beta.2/freebsd-ia32-42_binding.node
  v4.0.0/darwin-x64-11_binding.node
```

lib/extensions.js
saper: "Braces" are usually `{}`, I think you mean `()`?

Check this out (on current `master`):

````
m> node scripts/install.js --sass-binary-url http://www.google.com/
/home/saper/sw/test/node_modules/node-sass/lib/extensions.js:90
                                                        ^
TypeError: Cannot read property 'binaryUrl' of undefined
    at getBinaryUrl (/home/saper/sw/test/node_modules/node-sass/lib/extensions.js:90:57)
    at Object.<anonymous> (/home/saper/sw/test/node_modules/node-sass/lib/extensions.js:114:18)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/saper/sw/test/node_modules/node-sass/scripts/install.js:11:1)
    at Module._compile (module.js:460:26)
````

Comment (am11) [inline lib/extensions.js]: Yes I meant soft brackets. :)
I thought we had a test for it and it is passing?

But anyway we need to remove sass-binary-url.

**Comment by am11** [comment]
You can port the related test from other PR.
Otherwise LGTM :+1:

**Comment by xzyfer** [comment]
>What do you think about test refactoring? https://github.com/sass/node-sass/issues/775#issuecomment-95117730

I'm not sure I have context here. Are you referring to https://github.com/sass/node-sass/pull/835#issuecomment-91781524?

If so I don't what PR he is referring to. Please fill me in.

**Comment by saper** [comment]
Sorry. It is #847. I found it very boring to copy paste tests to test this one (#835)

File: lib/extensions.js [hunk 6]
```diff
@@ -76,21 +76,40 @@ function getBinaryName() {
...
-  return flags['--sass-binary-url'] ||
-         package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null ||
-         process.env.SASS_BINARY_URL ||
-         ['https://github.com/sass/node-sass/releases/download/v',
-          package.version, '/', sass.binaryName].join('');
+  return [
+           (
+             flags['--sass-binary-site'] ||
+             (package.nodeSassConfig ? package.nodeSassConfig.binarySite : null) ||
+             process.env.SASS_BINARY_SITE || 'https://github.com/sass/node-sass/releases/download'
```
xzyfer: Please move the default url into package.json

xzyfer: If I'm reading this correct could we not simply default `package.nodeSassConfig.binarySite` to `'https://github.com/sass/node-sass/releases/download'`?

xzyfer:
https://github.com/sass/node-sass/pull/847 appear to have this patch included. Closing this.

PR #836, opened 2015-04-04
Set User-Agent: HTTP header

saper:


am11:
Thanks. But what is the real significance of adding UA header in node-sass binary download case?

saper:
It is always good to have it; I noticed we don't have one when I was testing proxy settings and #775 change - http-server just shows "undefined". I am sure corporate local mirrors would really have a proper User-Agent set for their logs.

am11:
Thanks. :+1: 

(having a reason for change is good for future reference :wink:)

xzyfer:
I agree this is useful. Scheduled it for 3.0.0.

scripts/install.js
```diff
@@ -25,6 +25,13 @@ function download(url, dest, cb) {
       cb(typeof err.message === 'string' ? err.message : err);
     };
 
+    var package = require('../package.json');
```
xzyfer: Please move this to the top with the rest of the `require`s

xzyfer:
Can you please rebase this off master? It's conflicting with recent updates.

xzyfer:
Superseded by https://github.com/sass/node-sass/pull/907

Issue #828, opened 2015-04-01
Support building directory/glob of SASS files from CLI


patrickarlt:
Coming from Ruby SASS the one feature I am missing with the node-sass command line is the ability to build a directory of SASS files. For example in Ruby SASS I can do...

```
$ sass --watch sass
>>> Sass is watching for changes. Press Ctrl-C to stop.
      write sass/one.css
      write sass/one.css.map
      write sass/two.css
      write sass/two.css.map
```

And a build is kicked off immediately and it watches and rebuilds. But the same in Node SASS produces a usage message.

```
$ node-sass --watch sass
Provide a Sass file to render

  Example
    node-sass --output-style compressed foobar.scss foobar.css
    cat foobar.scss | node-sass --output-style compressed > foobar.css
```
This is really needed for parity with Ruby SASS and for larger projects where they might be multiple separate files that need to be built.

paulcpederson:
for now, you can use [scss-cli](https://github.com/paulcpederson/scss-cli)

am11:
@patrickarlt, thanks for your feedback, we will have a look. Albeit, this might not land in v3.0 but we have this planned.
@paulcpederson, would you like to send a PR to node-sass instead?

/cc @wesleytodd

paulcpederson:
@am11 yeah! I can definitely try! It would be much better to have this behavior in node-sass, always intended for scss-cli to be a band-aid measure, for sure.
 
Looking at #685 it looks like you started with this, but it was closed. What happened?

Also, what @patrickarlt is expecting, which I think is pretty common, is that if node-sass is pointed at a directory, it will recursively compile all sass files in said directory *except* for sass files beginning with an `_`. Obviously you could build this functionality yourself using grep or bash scripts, but IMO this is well within the scope of node-sass. So the feature requests are:

1. Compile a directory of sass files to css
2. Ignore sass files that begin with `_`

Currently in ruby sass this works with the `--watch` flag, but it also works with the (undocumented?) `--update` flag. So I can do something like `sass --update stylesheets:css` and it will recursively compile sass files in `stylesheets/` to `css/`. I've never been a huge fan of this syntax as it feels counter intuitive to use a flag like `--update` when aren't I always updating things when I compile *any* sass?

To my mind, there are a couple ways to accomplish this in node-sass: 

1. Using globs like `node-sass sass/**/*.scss -o css/ `
2. Automatically detecting a directory, like `node-sass sass/ -o css/`
[...]

PR #838, opened 2015-04-06
CLI: Adds support for multi-file compilation.

paulcpederson:
As per #828 - Adds the ability to compile multiple files from a directory.

Given this structure:

```
- sass/
| - main.scss
|  | - components/
|  |  | - _component1.scss
|  | - pages/
|  |  | - page1.scss
|  |  | - page2.scss
```

If you run: `node-sass sass/ -o build/`, you would get this output:

```
- sass/
- build/
| - main.css
|  | - pages/
|  |  | - page1.css
|  |  | - page2.css
```

Skips files that start with an `_`. Also respects the current `--recursive` flag. So given the same structure as above, if you ran `node-sass sass/ -o build/ --recursive false`, you would get:

```
- sass/
- build/
| - main.css
```

- Throws an error if you don't difine `--output`
- Throws an error if `--output` isn't a directory
- If `--output` is defined, but the folder doesn't exist, it will create it.

Very much looking for feedback. Things that feel specifically weird:

1. The fact that `done` is declared as a variable in the root scope of the file. (feels hacky)
2. The fact that this is waiting for each file to be finished before compiling the next one. (could probably be done in parallel?)

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -1,12 +1,16 @@
 #!/usr/bin/env node
 
 var Emitter = require('events').EventEmitter,
+    forEach = require('async-foreach').forEach,
     Gaze = require('gaze'),
     grapher = require('sass-graph'),
     meow = require('meow'),
+    util = require('util'),
     path = require('path'),
+    glob = require('glob'),
     render = require('../lib/render'),
-    stdin = require('get-stdin');
+    stdin = require('get-stdin'),
+    fs = require('fs');
 
 /**
  * Initialize CLI
@@ -88,14 +92,34 @@ var cli = meow({
 });
 
 /**
- * Check if file is a Sass file
+ * Is a Directory
  *
- * @param {String} file
+ * @param {String} filePath
+ * @returns {Boolean}
  * @api private
  */
 
-function isSassFile(file) {
-  return file.match(/\.(sass|scss)/);
+function isDirectory(filePath) {
+  var isDir = false;
+  try {
+    var absolutePath = path.resolve(process.cwd(), filePath);
+    isDir = fs.lstatSync(absolutePath).isDirectory();
+  } catch (e) {
+    isDir = e.code === 'ENOENT';
+  }
+  return isDir;
+}
+
+/**
+ * Get correct glob pattern
+ *
+ * @param {Object} options
+ * @returns {String}
+ * @api private
+ */
+
+function globPattern(options) {
+  return options.recursive ? '**/*.{sass,scss}' : '*.{sass,scss}';
 }
 
 /**
@@ -123,7 +147,7 @@ function getEmitter() {
   });
 
   emitter.on('done', function() {
-    if (!options.watch) {
+    if (!options.watch && !options.directory) {
       process.exit;
     }
   });
@@ -150,6 +174,13 @@ function getOptions(args, options) {
       [path.basename(options.src, path.extname(options.src)), '.css'].join(''));  // replace ext.
   }
 
+  if (options.directory) {
+    var sassDir = path.resolve(process.cwd(), options.directory);
+    var file = path.relative(sassDir, args[0]);
+    var cssDir = path.resolve(process.cwd(), options.output);
+    options.dest = path.join(cssDir, file).replace(path.extname(file), '.css');
+  }
+
   return options;
 }
 
@@ -162,8 +193,7 @@ function getOptions(args, options) {
  */
 
 function watch(options, emitter) {
-  var glob = options.recursive ? '**/*.{sass,scss}' : '*.{sass,scss}';
-  var src = isSassFile(options.src) ? options.src : path.join(options.src, glob);
+  var src = options.directory ? path.join(options.directory, globPattern(options)) : options.src;
   var graph = grapher.parseDir(path.resolve(path.dirname(src)), { loadPaths: options.includePath });
   var watch = [];
 
@@ -181,12 +211,10 @@ function watch(options, emitter) {
     graph.visitAncestors(file, function(parent) {
       files.push(parent);
     });
-
     files.forEach(function(file) {
-      if (path.basename(file)[0] === '_') return;
-      options = getOptions([path.resolve(file)], options);
-      emitter.emit('warn', '=> changed: ' + file);
-      render(options, emitter);
+      if (path.basename(file)[0] !== '_') {
+        renderFile(file, options, emitter);
+      }
     });
   });
 }
@@ -202,6 +230,15 @@ function watch(options, emitter) {
 function run(options, emitter) {
   if (!Array.isArray(options.includePath)) {
     options.includePath = [options.includePath];
+  }
+
+  if (options.directory) {
+    if (!options.output) {
+      emitter.emit('error', 'Specify an output directory when using multi-file compilation');
+    }
+    if (!isDirectory(options.output)) {
+      emitter.emit('error', util.format('Multi-file compilation: requires destination %s to be a directory.', options.output));
+    }
   }
 
   if (options.sourceMap) {
@@ -235,9 +272,53 @@ function run(options, emitter) {
 
   if (options.watch) {
     watch(options, emitter);
+  } else if (options.directory) {
+    renderDir(options, emitter);
   } else {
     render(options, emitter);
   }
+}
+
+/**
+ * Render a file
+ *
+ * @param {String} file
+ * @param {Object} options
+ * @param {Object} emitter
+ * @api private
+ */
+function renderFile(file, options, emitter) {
+  options = getOptions([path.resolve(file)], options);
+  if (options.watch) {
+    emitter.emit('warn', util.format('=> changed: %s', file));
+  }
+  render(options, emitter);
+}
+
+/**
+ * Render all sass files in a directory
+ *
+ * @param {Object} options
+ * @param {Object} emitter
+ * @api private
+ */
+function renderDir(options, emitter) {
+  var globPath = path.resolve(process.cwd(), options.directory, globPattern(options));
+  glob(globPath, {ignore: '**/_*'}, function(err, files) {
+    if(err) {
+      return emitter.emit('error', util.format('You do not have permission to access this path: %s.', err.path));
+    } else if(!files.length) {
+      return emitter.emit('error', 'No input file was found.');
+    }
+    forEach(files, function(subject) {
+      emitter.once('done', this.async());
+      renderFile(subject, options, emitter);
+    }, function(successful, arr) {
+      var outputDir = path.join(process.cwd(), options.output);
+      emitter.emit('warn', util.format('Wrote %s CSS files to %s', arr.lenth, outputDir));
+      process.exit;
+    });
+  });
 }
 
 /**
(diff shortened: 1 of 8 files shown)


bin/node-sass
```diff
@@ -99,6 +102,41 @@ function isSassFile(file) {
...
+  var isDir = false;
+  try {
+    var absolutePath = path.resolve(process.cwd(), filePath);
+    isDir = fs.lstatSync(absolutePath).isDirectory();
+  } catch (e) {
+    if (e.code === 'ENOENT') {
+      isDir = true;
+    } else {
+      isDir = false;
+    }
```
am11:
```javascript
catch (e) {
  isDir = e.code === 'ENOENT';
}
```

bin/node-sass
```diff
@@ -235,17 +286,69 @@ function run(options, emitter) {
...
+    if(err) {
+      emitter.emit('error', ['You do not have permission to access this path:', err.path, '.']);
+    } else if(!files.length) {
+      emitter.emit('error', 'No input file was found.');
+    }
+    forEach(files, function(file) {
+      done = this.async();
+      renderFile(file, options, emitter)
+    },
+    function(successful, arr) {
```
am11:
```javascript
  renderFile(file, options, emitter);
}, function(_, arr) {
```

bin/node-sass
```diff
@@ -235,17 +286,69 @@ function run(options, emitter) {
...
+function renderDir(options, emitter) {
+  var globPath = path.resolve(process.cwd(), options.src, globPattern(options));
+  glob(globPath, function(err, files) {
+    if(err) {
+      emitter.emit('error', ['You do not have permission to access this path:', err.path, '.']);
+    } else if(!files.length) {
+      emitter.emit('error', 'No input file was found.');
+    }
+    forEach(files, function(file) {
+      done = this.async();
```
am11: Just a convention I picked to refer first argument of all predicates as `subject`.

```
forEach(files, function(file) {
```

to

```
forEach(files, function(subject) {
```

:-)

bin/node-sass
```diff
@@ -204,6 +246,15 @@ function run(options, emitter) {
     options.includePath = [options.includePath];
   }
 
+  if(isDirectory(options.src)) {
+    if (!options.output) {
+      emitter.emit('error', 'Specify an output directory when using multi-file compilation');
+    }
+    if (!isDirectory(options.output)) {
+      emitter.emit('error', ['Multi-file compilation: requires destination', options.output, 'to be a directory.'].join(' '));
```
am11: Should we return error emission? Like so:

```javascript
return emitter.emit('error..
```

am11:
This is a great contribution to node-sass. Thank you from the whole Sass community! :sunglasses: :+1: 

Regarding parallel compilation, I second that will help the performance. Using `async-foreach` module, is it not happening already?

@kevva, @wesleytodd, please take a look.

paulcpederson:
Cool! All great suggestions (especially re: subject of predicates :+1: ), updating pr now.

bin/node-sass
wesleytodd: Also, I am not sure what the standard in `node-sass` is, but I prefer to use `util.format` for messages like this.   It is what node core does: https://github.com/joyent/node/blob/ef4344311e19a4f73c031508252b21712b22fe8a/lib/tls.js#L152

bin/node-sass
```diff
@@ -204,6 +246,15 @@ function run(options, emitter) {
     options.includePath = [options.includePath];
   }
 
+  if(isDirectory(options.src)) {
```
wesleytodd: Since below you are just relying on `options.directory` being set, seems like you could just do that here to.

bin/node-sass
paulcpederson: Yeah, saw this style elsewhere in the file. Happy to use util though :+1:

paulcpederson: Actually, should I just convert all of these to use util?

am11: :+1:
We should refactor the code with formatting error. I just followed @kevva's lead on that one :stuck_out_tongue_winking_eye: 
(we have used that array-join thingy a lot -- all over the place!)

paulcpederson:
Ok, I went ahead and replace some other instances of the `.join` style string interpolation with the `util.format` style mentioned by @wesleytodd in other parts of the cli as well.

bin/node-sass
```diff
@@ -235,17 +286,69 @@ function run(options, emitter) {
...
+
+/**
+ * Render all sass files in a directory
+ *
+ * @param {Object} options
+ * @param {Object} emitter
+ * @api private
+ */
+function renderDir(options, emitter) {
+  var globPath = path.resolve(process.cwd(), options.src, globPattern(options));
```
wesleytodd: For clarity, shouldn't `options.src` here use `options.directory` instead?  I know they are equal, but it might read more clearly for the next person who comes along.

paulcpederson: great idea. that would be rather confusing.

wesleytodd:
@paulcpederson I see what you mean about the `done` function, seems like this would be a good situation to use `EventEmitter.once`.  You could remove the disconnected calling of `done` by putting it right next to the related code:

```javascript
  forEach(files, function(file) {
      emitter.once('done', this.async());
      renderFile(file, options, emitter);  // Note you were missing a semicolon here :)
    },
```

I think you will still need the `!options.directory` line in the other event handler, but it gets rid of the need for the "global".

bin/node-sass
```diff
@@ -162,8 +207,7 @@ function getOptions(args, options) {
  */
 
 function watch(options, emitter) {
-  var glob = options.recursive ? '**/*.{sass,scss}' : '*.{sass,scss}';
-  var src = isSassFile(options.src) ? options.src : path.join(options.src, glob);
+  var src = isSassFile(options.src) ? options.src : path.join(options.src, globPattern(options));
```
wesleytodd: I might be wrong here, but doesn't your `options.directory` solve the same problem as this `isSassFile` call?  Maybe you can just use that here, and replace the second `options.src` with `options.directory`.  Also you can replace `path.resolve(path.dirname(src))` on the line below if we already know we have a directory.

paulcpederson: I think you are correct. let me look.

paulcpederson:
@wesleytodd good idea. let me try that out now.

wesleytodd:
Sorry for the load of comments..  Sometimes I feel like half my real job is peer review....my coworkers probably hate me :)

Seriously tho, this is a good PR.  I will most definitely use this feature once it is merged and working.  So thanks for the hard work!

@am11 & whoever else cares, I hope you don't mind my pitching in like this.  Let me know if I am going overboard as I really haven't been involved in the development of this project much before.  But I guess that is the deal with OSS right?  Anyone can say whatever they want....

paulcpederson:
I *love* getting comments like this. I feel like they make me a better developer! :smile_cat: 

Hold off on merging this though. I am finding some problems using this with the watcher. Fixing them now.

am11:
@wesleytodd, you are not stepping toes at all, but the opposite; a valued contribution! Feel free to continue.. :)

wesleytodd:
@am11 @paulcpederson Glad to hear it!  Peer reviews are like the way I have learned the most in my career.

paulcpederson:
@wesleytodd ok! updated everything. `emitter.once` was definitely the way to go. That let me clean up a lot of stuff!

wesleytodd:
Looks good to me!

bin/node-sass
```diff
@@ -204,6 +232,15 @@ function run(options, emitter) {
     options.includePath = [options.includePath];
   }
 
+  if (options.directory) {
+    if (!options.output) {
+      emitter.emit('error', 'Specify an output directory when using multi-file compilation');
+    }
+    if (!isDirectory(options.output)) {
+      emitter.emit('error', util.format('Multi-file compilation: requires destination %s to be a directory.', options.output));
+    }
```
am11: Can we add tests for these two conditions?
For example: [cli.js#L438-L448](https://github.com/sass/node-sass/blob/2d22d9c7129a8c45e261de9764efb5f2ee396c05/test/cli.js#L438-L448).

paulcpederson: Good idea.

test/cli.js
```diff
@@ -301,6 +302,86 @@ describe('cli', function() {
...
+
+      bin.once('close', function() {
+        var files = fs.readdirSync(dest);
+        assert.deepEqual(files, ['one.css', 'two.css']);
+        rimraf.sync(dest);
+        done();
+      });
+    });
+
+    it('should error if no output directory is provided', function(done) {
```
paulcpederson: @am11 ok, those errors are tested now.

paulcpederson:
@am11 ok. I managed to test for the error message. The data was coming out of stderr as a buffer, which included `\n` newlines. So I had to convert it to a string and test for the error message as a substring. Maybe I'm doing something wrong?

paulcpederson:
They do test for the message now, though, and the tests pass.

test/cli.js
```diff
@@ -301,6 +303,88 @@ describe('cli', function() {
...
+    });
+
+    it('should error if output directory is not a directory', function(done) {
+      var src = fixture('input-directory/sass');
+      var dest = fixture('input-directory/sass/one.scss');
+      var bin = spawn(cli, [src, '--output', dest]);
+      var msg = util.format('Multi-file compilation: requires destination %s to be a directory.', dest);
+
+      bin.stderr.once('data', function(data) {
+        assert(data.toString('utf8').indexOf(msg) > -1);
```
am11: Probably something like this:

```javascript
bin.stderr.setEncoding('utf8');
bin.stderr.once('data', function(data) {
  assert.equal(data, msg);
  done();
});
```

am11:
@paulcpederson, looks pretty good. I have left a comment, probably it would be better to mark the stderr stream as `utf8` prior to probing it.

test/cli.js
wesleytodd: This is actually a really brittle test.  Usually when I see myself doing something like this I try to take a step back and ask, "what am I really testing?"  Here the answer is that an error occurred.  We don't care what the error messages was in a strict sense.

With that in mind, it might be better to test that there WAS output, and that the files were NOT compiled.  That is the true test that will last longer than the next person who wants to change that error message.  What do you all think about that?

am11: Good idea. In that case we should check whether the result (compiled target) is produced in `bin.once('close', ..` instead of probing `bin.stderr`.

paulcpederson: I tend to agree with @wesleytodd . I will rewrite these to check that the compiled target doesn't exist after running and probably that it threw an error. Because if it silently just failed that would also be bad.

paulcpederson:
@am11 @wesleytodd Now the tests both check for two things:

1. Exited with an error code.
2. No css files were generated anywhere inside the fixture.

I think that's a pretty good halfway point between checking for the exact error message and just checking there was an error at all. What do you guys think?

wesleytodd:
@paulcpederson Yeah, thats awesome!!  This is a much more stable test.

am11:
Excellent work @paulcpederson. :+1:

PR #841, opened 2015-04-07
CLI: switch successful message event type

gibatronic:
hey guys. this pull request is meant to solve a simple issue:
send CLI succesful messages to stdout instead of stderr.

currently redirecting the stdout has no effect:

```shell
node-sass input.scss output.css > /dev/null
```

the above command still outputs:

```
Rendering Complete, saving .css file...
Wrote CSS to path/to/output.css
```

I could redirect the stderr, but then it would be impossible to know if something breaks.

what do you guys think?

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -138,10 +138,6 @@ function getEmitter() {
     }
   });
 
-  emitter.on('warn', function(data) {
-    console.warn(data);
-  });
-
   emitter.on('log', function(data) {
     console.log(data);
   });
@@ -290,7 +286,7 @@ function run(options, emitter) {
 function renderFile(file, options, emitter) {
   options = getOptions([path.resolve(file)], options);
   if (options.watch) {
-    emitter.emit('warn', util.format('=> changed: %s', file));
+    emitter.emit('log', util.format('=> changed: %s', file));
   }
   render(options, emitter);
 }
@@ -315,7 +311,7 @@ function renderDir(options, emitter) {
       renderFile(subject, options, emitter);
     }, function(successful, arr) {
       var outputDir = path.join(process.cwd(), options.output);
-      emitter.emit('warn', util.format('Wrote %s CSS files to %s', arr.lenth, outputDir));
+      emitter.emit('log', util.format('Wrote %s CSS files to %s', arr.lenth, outputDir));
       process.exit;
     });
   });
--- a/lib/render.js
+++ b/lib/render.js
@@ -60,7 +60,7 @@ module.exports = function(options, emitter) {
       return done();
     }
 
-    emitter.emit('warn', chalk.green('Rendering Complete, saving .css file...'));
+    emitter.emit('log', chalk.green('Rendering Complete, saving .css file...'));
 
     mkdirp(path.dirname(options.dest), function(err) {
       if (err) {
@@ -72,7 +72,7 @@ module.exports = function(options, emitter) {
           return emitter.emit('error', chalk.red(err));
         }
 
-        emitter.emit('warn', chalk.green('Wrote CSS to ' + options.dest));
+        emitter.emit('log', chalk.green('Wrote CSS to ' + options.dest));
         emitter.emit('write', err, options.dest, result.css.toString());
         done();
       });
@@ -86,7 +86,7 @@ module.exports = function(options, emitter) {
           return emitter.emit('error', chalk.red('Error' + err));
         }
 
-        emitter.emit('warn', chalk.green('Wrote Source Map to ' + options.sourceMap));
+        emitter.emit('log', chalk.green('Wrote Source Map to ' + options.sourceMap));
         emitter.emit('write-source-map', err, options.sourceMap, result.map);
         done();
       });
--- a/test/cli.js
+++ b/test/cli.js
@@ -169,15 +169,15 @@ describe('cli', function() {
       }, 100);
     });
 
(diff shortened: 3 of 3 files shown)


am11:
LGTM!

Can you rebase with master? I have just merged #838.

gibatronic:
done!

am11:
Thanks!

Going by this logic, we should replace the following instances as well:

* [bin/node-sass#L293](https://github.com/sass/node-sass/blob/cb9f55e76238cd4b8de3c3e0bc2adb9920b6d610/bin/node-sass#L293)
* [bin/node-sass#L318](https://github.com/sass/node-sass/blob/cb9f55e76238cd4b8de3c3e0bc2adb9920b6d610/bin/node-sass#L318)

Meaning we don't need emitter to `warn` at all? So lets remove:

* [bin/node-sass#L143](https://github.com/sass/node-sass/blob/cb9f55e76238cd4b8de3c3e0bc2adb9920b6d610/bin/node-sass#L141-L143)

as well and fix the related tests: probably somewhere `bin.stderr.once` would need to be replaced by `bin.stdout.once` in test/cli.js?


@wesleytodd, @paulcpederson, @kevva, thoughts?

gibatronic:
uhmm... right!

paulcpederson:
Using `warn` for a success message did feel a bit odd. Great idea!

gibatronic:
sorry, I don't understand why `should watch the full sass dep tree for a single file` has failed.

gibatronic:
oh I see... with `Node v0.12.2` and `io.js v1.6.4` everything is fine, the issue is with `Node v0.10.38`

gibatronic:
nope. even with `v0.10.38` everything went fine locally...

wesleytodd:
Ok, so after digging in here I see the problem.  This PR breaks standard unix programing pattern of `stdout` being for the IO of the actual program.  So the failing test are because you actually broke that feature.

This is actually why the logging is all supposed to go to `stderr`.  Most likely this PR needs to be denied.  The only way this might be acceptable is behind a flag.  Or maybe adding a `--silent` flag so you can suppress the logging to `stderr` if you have something that requires that.

For more reference: http://en.wikipedia.org/wiki/Standard_streams

> Standard output is the stream where a program writes its output data.

> Standard error is another output stream typically used by programs to output error messages or diagnostics.

wesleytodd:
@gibatronic Maybe you should post more about why you need this feature?  Then we could brainstorm something better to help?

gibatronic:
it's alright, it's probably not valuable enough for common usage.
I like the idea of a `silent` flag, but only to suppress `warn` events.
my need is to only show error events, the warning ones I would like to hid.
does it make sense for you guys?

am11:
Agree with @wesleytodd; this kind of output should not appear on stdout. So lets keep the `console.warn` and provide the said boolean flag. As for the flag name, how about `--silent-warning` or `--mute-warning`?

paulcpederson:
@wesleytodd +1 good sleuthing.

Having a flag to suppress `console.warn` messages would actually be really useful. Especially for people who are compiling lots of files. 

As for the name, I've seen `--quiet` used in a couple projects like [live-server](https://www.npmjs.com/package/live-server#usage-from-command-line), [supervisor](https://www.npmjs.com/package/supervisor#node-supervisor),and passed to [`git rm`](http://git-scm.com/docs/git-rm) (and probably other git commands). Not sure if that is descriptive/obvious enough, though...

am11:
> As for the name, I've seen  `--quiet`  

Note that we need to suppress warnings only and keep the errors on stderr.

wesleytodd:
`--quiet` fits better when we are not suppressing all output, IMHO.

gibatronic:
yea... `--silent-warnings` or `--mute-warnings` sounds great!

paulcpederson:
Good point. :+1:

gibatronic:
alright! I'll work on the `--silent-warnings` flag then!
thanks for your time guys, excellent thoughts.
soon I'll be making another pull request then.

wesleytodd:
Personally I would choose `--quite` because as @paulcpederson mentioned it is a commonly used flag in other projects.  `--silent-warnings` is a flag I have not seen used before, and although it makes perfect sense, I think there are benefits to following the herd in this.

paulcpederson:
A good example is `git rm --quiet`. Git still keeps errors on stderr, it just stops writing warnings like `rm 'thing'` when it removes things. That's what we're talking about here, right? Completely possible I don't understand the desired behavior correctly...

gibatronic:
not a problem, I'm fine with `--quiet`. I'm really up to you guys.
searching for it in Github I got this numbers:

 1. `--quiet` ≈ [1,191,999](https://github.com/search?q=--quiet&type=Code&utf8=%E2%9C%93)
 2. `--silent` ≈ [552,757](https://github.com/search?q=--silent&type=Code&utf8=%E2%9C%93)
 3. `--mute` ≈ [3,636](https://github.com/search?q=--mute&type=Code&utf8=%E2%9C%93)

am11:
`--quite` (with [unreserved alias `-q`](https://github.com/sass/node-sass/blob/bccad58ba20faa0d90934fca344bb44ead919530/bin/node-sass#L75-L81)) is fine and niftier than `--mute-warnings`, if it delivers the meanings. My suggestion was based on the verbosity. 

Hey, I am spoiled by windows powershell: `-ErrorAction SilentlyContinue`! :smile_cat:

PR #847, opened 2015-04-10
Fix #775: Introduce SASS_BINARY_SITE environment, refactor tests/api.js

### LINKED ISSUE #775: Impossible to mirror the libsass binaries

saper:
Factor out `tests/render.js` out of `tests/api.js`,
reducing boilerplate and allowing easier save and
restore of the environment.

Add tests for #775


am11:
The test for #775 should go in #775. You can combine the refactoring of test in one PR. Would be easy to review, monitor CI progress and since so forth.

saper:
This pull request also includes #847; I propose we discuss refactoring of tests here - it makes no sense to introduce another PR just for commit 565ece9 and commit 31321b9 serves as a justification why I needed to refactor this. Unfortunately, unlike in Gerrit Code Review, I cannot stack pull requests on top of another. If you think that `testes/render.js` is unproblematic, we can merge this and #847 closes. If not, let's discuss testing here and push #847 alone, as it's trivial.

xzyfer:
I'll look at this post #775.

lib/extensions.js
```diff
@@ -76,21 +76,40 @@ function getBinaryName() {
...
-         package.nodeSassConfig ? package.nodeSassConfig.binaryUrl : null ||
-         process.env.SASS_BINARY_URL ||
-         ['https://github.com/sass/node-sass/releases/download/v',
-          package.version, '/', sass.binaryName].join('');
+  var site = flags['--sass-binary-site'] || process.env.SASS_BINARY_SITE  ||
+             (package.nodeSassConfig ? package.nodeSassConfig.binarySite : null);
+  if (site) {
+		  return [site, 'v' + package.version, sass.binaryName].join('/');
+  } else {
+          return null;
```
xzyfer: This shoyldn't be possible. We should default to `'https://github.com/sass/node-sass/releases/download/v` if the users has give no proeference.

package.json
```diff
@@ -18,6 +18,9 @@
     "node": ">=0.10.0"
   },
   "main": "lib/index.js",
+  "nodeSassConfig": {
+    "binarySite1": "https://github.com/sass/node-sass/releases/download"
```
xzyfer: typo? `binarySite1`

xzyfer:
Superseded by https://github.com/sass/node-sass/pull/901.

PR #848, opened 2015-04-10
Show expected binary path in the error message if bindings not found

saper:


am11:
@kevva, you merged the test suite at some point into api.js and cli.js, when we used to have multiple files. Thoughts on this on?

saper:
The problem is that dependent commits have to be included so that dependent and includes first commit of #847

xzyfer:
:+1: scheduled for 3.0.0

PR #854, opened 2015-04-12
Close #713 (handle->flags & UV_CLOSING) assertion failure

### LINKED ISSUE #713: Assertion failed: (handle->flags & UV_CLOSING), function uv__finish_close, file ../deps/uv/src/unix/core.c, line 209.
I am randomly seeing this error when using the importer function. My intention is to treat @import like require(). Here is the offending code: 

```javascript
importer: function (url, prev, done) {
               if(imported[url]) {
                    done({contents: ''});
                } else {
                    imported[url] = true;
                    done({file: url});
                }
            }
````

saper:
It is not safe to delete uv_async_t object after
calling uv_close(). This object remains in the
closing queue and can be purged only in the final
callback.

### DIFF
--- a/src/callback_bridge.h
+++ b/src/callback_bridge.h
@@ -23,6 +23,7 @@ class CallbackBridge {
     // We will expose a bridge object to the JS callback that wraps this instance so we don't loose context.
     // This is the V8 constructor for such objects.
     static Handle<Function> get_wrapper_constructor();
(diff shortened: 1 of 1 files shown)


xzyfer:
If there are no immediate objections I'll be merging this shortly. If I never see this exception again it'll be too soon. Great work @saper !

PR #863, opened 2015-04-14
Fix Assertion failed: (handle->flags & UV_CLOSING)

saper:
Fix issue #713 for v2.1.1


xzyfer:
Due to the way we current do releases patching and doing another 2.x isn't likely to happen. Libsass 3.2.0 stable will be out this week, node-sass@3.0.0 will follow shortly after.

Thanks for your work @saper .

PR #874, opened 2015-04-20
Check libSass submodule when npm installing via git url

xzyfer:
This PR fixes npm installs via git urls erring when building the bindings.

When developing on node-sass I often npm install dev branches into my production code to test for regression or other issues. At the moment this require checking in the `src/libsass` directory, then rebasing that commit out when creating a PR.

This PR makes it so the libsass submodule is checked out as part of the postinstall process. The version checked out denoted by the existing `package.libsass` flag.

/cc @am11

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -3,6 +3,7 @@
  */
 
 var eol = require('os').EOL,
+    pkg = require('../package.json'),
     fs = require('fs'),
     mkdir = require('mkdirp'),
     path = require('path'),
@@ -46,6 +47,49 @@ function afterBuild(options) {
 }
 
 /**
+ * initSubmodules
+ *
+ * @param {Function} cb
+ * @api private
+ */
+
+function initSubmodules(cb) {
+  var errorMsg = '';
+  var git = spawn(['LIBSASS_GIT_VERSION=', pkg.libsass, ' ./scripts/git.sh'].join());
+  git.stderr.on('data', function(data) {
+    errorMsg += data.toString();
+  });
+  git.on('close', function(code) {
+    var error;
+    if (code !== 0) {
+      error = { message: errorMsg + 'Unable to checkout the libSass submodule' };
+    }
+    cb(error);
+  });
+}
+
+/**
+ * installGitDependencies
+ *
+ * @param {Function} cb
+ * @api private
+ */
+
+function installGitDependencies(cb) {
+  var libsassPath = './src/libsass';
+
+  if (fs.access) { // node 0.12+, iojs 1.0.0+
+    fs.access(libsassPath, fs.R_OK, function(err) {
+      err && err.code === 'ENOENT' ? initSubmodules(cb) : cb();
+    });
+  } else { // node < 0.12
+    fs.exists(libsassPath, function(exists) {
+      exists ? cb() : initSubmodules(cb);
+    });
+  }
+}
+
+/**
  * Build
  *
  * @param {Object} options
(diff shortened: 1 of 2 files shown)


saper:
Put some comments inline!

xzyfer:
@saper regarding your comments in the commit. The git code path is only hit when install node-sass via git url (which is currently broken) i.e.

```
npm install sass/node-sass#my/branch
```

Given this case `build.js` fails because gyp expect the libsass submodule to exists, which it does when installed via npm. If the `src/libsass` does not exist we can assume this was a git install not an npm one.

We cannot use a git submodule update because in npm does not maintain the git metadata when install via git. https://github.com/npm/npm/issues/6700#issuecomment-71302066

scripts/build.js
```diff
@@ -53,25 +97,33 @@ function afterBuild(options) {
...
+    proc.on('exit', function(errorCode) {
+      if (!errorCode) {
+        afterBuild(options);
 
-    console.error(errorCode === 127 ? 'node-gyp not found!' : 'Build failed');
+        return;
+      }
+
+      console.error(errorCode === 127 ? 'node-gyp not found!' : 'Build failed');
+      process.exit(1);
```
xzyfer: Currently if the postinstall script is run directly and building fails it exits with a `0`.

```
npm run postinstall
```

scripts/git.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+git clone --depth=1 git@github.com:sass/libsass.git ./src/libsass
+cd ./src/libsass
+git checkout $LIBSASS_GIT_VERSION
```
xzyfer: This is required be cause we have no way to know which commit the submodule is linked to when npm installs via git https://github.com/npm/npm/issues/6700#issuecomment-71302066

scripts/build.js
```diff
@@ -46,6 +47,49 @@ function afterBuild(options) {
...
 /**
+ * initSubmodules
+ *
+ * @param {Function} cb
+ * @api private
+ */
+
+function initSubmodules(cb) {
+  var errorMsg = '';
+  var git = spawn(['LIBSASS_GIT_VERSION=', pkg.libsass, ' ./scripts/git.sh'].join());
```
xzyfer: Make sure we checkout out libSass at the version we're reporting to user via `node-sass -v`.

This also gives us a central hook if we're experimenting with libSass branches etc..

saper:
I was wondering how do we know we checkout via git? Because `build.js` can be started in the usual `npm install` case in case the binary does not exist. Requiring `git` and SSH at this stage would be too much. I need to check what happens to `.gitmodules` file during npm install from git, maybe we can somehow salvage the problem. 

In any case we need not to do this when we want to have node-sass against the external library (like I do for FreeBSD binaries and also for development).

xzyfer:
> Because build.js can be started in the usual npm install case in case the binary does not exist

If the code is checked out via git or installed via npm (from npm) the `src/libsass` dir exists. This covers the use cases we expect from our users IMHO.

> I need to check what happens to .gitmodules file during npm install from git

npm/npm#6700 (comment)

> maybe we can somehow salvage the problem

I'm not following why we would care about `.gitmodules`, what use case are you addressing?

The `libsass` version in `package.json` will always be a tag that is the same commit as the submodule.
The exception being if you're developing on node-sass and have changed the version, in which I like that the `package.json` is a single point of truth. 

> In any case we need not to do this when we want to have node-sass against the external library

Do you need to run the build script in this case? What does the work flow look like? I wouldn't expect you to hit this code block unless you were doing an npm install on to a git backend.

Please correct me if I'm mistaken.

scripts/build.js
```diff
@@ -46,6 +47,49 @@ function afterBuild(options) {
...
+
+/**
+ * installGitDependencies
+ *
+ * @param {Function} cb
+ * @api private
+ */
+
+function installGitDependencies(cb) {
+  var libsassPath = './src/libsass';
```
xzyfer: If we don't have this directory we're almost certainly going to run into issues building node-sass.

If it does exist, which it does when installed via npm (from npm), we do nothing and continue on as before.

saper:
Ok, let's do that step by step:

My first failure is probably because I am running csh:

https://gist.github.com/5cdcd0f7133c7c7398b5

xzyfer:
Looks like an issue with my usage of [join()](https://github.com/sass/node-sass/pull/874/files#diff-ea1c4abd09127cfb80f898b2bf12b0a4R58) without a param. It's defaulting to a comma as [per the spec](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/join).

This wasn't happening for me but since it's the spec I'll patch it.

>LIBSASS_GIT_VERSION=,3.2.0-beta.5, ./scripts/git.sh

xzyfer:
I believe I have a fix for your issue in https://github.com/sass/node-sass/pull/875. Please confirm?

PR #876, opened 2015-04-20
Handle build script git operations in JS

xzyfer:
This PR is an iteration in the build script's git fallback introduced in #874.

We're now executing the git operations purely from JS - no more shell script - because we <3 Windows.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -47,6 +47,24 @@ function afterBuild(options) {
 }
 
 /**
+ * manageProcess
+ *
+ * @param {ChildProcess} proc
+ * @param {Function} cb
+ * @api private
+ */
+
+function manageProcess(proc, cb) {
+  var errorMsg = '';
+  proc.stderr.on('data', function(data) {
+    errorMsg += data.toString();
+  });
+  proc.on('close', function(code) {
+    cb(code === 0 ? null : { message: errorMsg });
+  });
+}
+
+/**
  * initSubmodules
  *
  * @param {Function} cb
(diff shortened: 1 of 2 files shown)


saper:
Thanks, this works now for me. I still would have preferred not to have this done in the `libsass_ext=[auto|yes]` case but I have no time to change this; let's see if the new beta builds on my FreeBSD binary building tool.

xzyfer:
I admit to not being completely familiar with all the build flags. Can you elaborate on the purpose of `libsass_ext` and why you would prefer this change?

saper:
On Tue, 21 Apr 2015, Michael Mifsud wrote:

> I admit to not being completely familiar with all the build flags. Can
> you elaborate on the purpose of `libsass_ext` and why you would prefer
> this change?

I have proposed this in https://github.com/sass/node-sass/pull/763

This way you can build with libsass you have already on your
system and not with the bundled one. I use this to use proper
autoconf configuration on the libsass; the library built this
way gets included in the FreeBSD library. For those builds
it is the best to have src/libsass empty to prevent
inclusion of possibly wrong headers.

For my automated builds I am using the following
options (${LOCALBASE} is usually /usr/local):

LIBSASS_EXT=yes LIBSASS_LIBRARY=${LOCALBASE}/lib/libsass.a
LIBSASS_CFLAGS=-I${LOCALBASE}/include

It is also possible to use LIBSASS_EXT=auto to just use
pkg-config (I use that sometimes for manual testing).
For quick manual testing of C++ bindings this is much
faster; I can also test the same bindings against
various minor version of libsass and check if something
breaks.

xzyfer:
I added the LIBSASS_EXT flag as one of the git bail out flags in https://github.com/sass/node-sass/pull/880

PR #879, opened 2015-04-21
Update sourcemap test to match updates in Libsass

xzyfer:
This PR updates a sourcemap spec that is now failing due to fixes in Libsass. Neither this or previous output is technically correct, but this is more correct in that it match Libsass' expected output.


xzyfer:
This PR will fail CI until https://github.com/sass/node-sass/pull/872 is merged.

PR #880, opened 2015-04-21
Respect the LIBSASS_EXT build flag when falling back to git

xzyfer:
This PR updates the build script to respect the LIBSASS_EXT build flag when falling back to git. Originally suggested in https://github.com/sass/node-sass/pull/876#issuecomment-94730030

@saper does this suit your usecase?

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -98,10 +98,12 @@ function initSubmodules(cb) {
  * @api private
  */
 
-function installGitDependencies(cb) {
+function installGitDependencies(options, cb) {
   var libsassPath = './src/libsass';
 
-  if (fs.access) { // node 0.12+, iojs 1.0.0+
+  if (process.env.LIBSASS_EXT || options.libsassExt) {
+    cb();
+  } else if (fs.access) { // node 0.12+, iojs 1.0.0+
     fs.access(libsassPath, fs.R_OK, function(err) {
       err && err.code === 'ENOENT' ? initSubmodules(cb) : cb();
     });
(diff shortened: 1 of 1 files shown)


xzyfer:
These failures are expected due to https://github.com/sass/node-sass/pull/879. Nothing to see here :simple_smile:

xzyfer:
@saper I think this is legit given the use case for `LIBSASS_EXT`. Any objections?

saper:
Yeah but to be fully kosher you need to take care of the command line option as well.

xzyfer:
I've had a look around and I can't a relevant cli flag. Am I missing something?

saper:
On Wed, 29 Apr 2015, Michael Mifsud wrote:

> I've had a look around and I can't a relevant cli flag. Am I missing something?

They are buried in

https://github.com/sass/node-sass/blob/master/scripts/build.js#L130-L131

Nice CLI option handling after 3.0.0 :)

xzyfer:
Yeah I see that, but it looks like it's actually pulling the value out of env when building the cli flags. is that not the case?

xzyfer:
I guess it's possible if

```js
options.args = ['--libsass_ext', ''/path/to/libsass']
```

saper:
On Wed, 29 Apr 2015, Michael Mifsud wrote:

> Yeah I see that, but it looks like it's actually pulling the value out of env when building the cli flags. is that not the case?

I should have bean clearer:

If you 'node script/install.js --libsass_ext=auto', `installGitDependencies` will not notice.

We could move installGitDependencies a bit down after constructing args first.

xzyfer:
Yeah cool/ Updated to handle the cli flag.

PR #901, opened 2015-04-28
Introduce SASS_BINARY_SITE environment and refactor API tests

### LINKED ISSUE #775: Impossible to mirror the libsass binaries

xzyfer:
This PR implements with https://github.com/sass/node-sass/pull/847 with a minor clean up, and merge conflicts resolved.

>Provide ability to locally mirror node-sass binaries for various versions and platforms.
>
>SASS_DIST_SITE needs to be an URL pointing to a collection of files organized like the Github repository.
>
>If SASS_DIST_SITE is set to
>
>http://myhost:8080/local/node-sass-bin
>
>then
>
>http://myhost:8080/local/node-sass-bin/v3.0.0-beta.4/freebsd-x64-14_binding.node
>
>should point to the FreeBSD 64 bit binary for node 0.12.0

Fixes #775.

***
**Breaking changes**

This PR removes the existing `binaryUrl` config in favour the more useful `binarySite` config.

/cc @saper


saper:
Thank you. This works well for me. If `package.json` is broken we get `TypeError: Cannot read property 'binarySite' of undefined` which also okay.

Tested with a local http repository.

PR #907, opened 2015-04-29
Set User-Agent: HTTP header

xzyfer:
This PR implements with #836 with a minor clean up, and merge conflicts resolved.

Upon installation when fetching the binary we'll use a node-sass specific `User-Agent` header.

/cc @saper

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -6,7 +6,8 @@ var fs = require('fs'),
     mkdir = require('mkdirp'),
     npmconf = require('npmconf'),
     path = require('path'),
-    request = require('request');
+    request = require('request'),
+    package = require('../package.json');
 
 require('../lib/extensions');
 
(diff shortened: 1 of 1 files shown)


saper:
Thanks! I was working on it... Just noticed an error handling here does not work (not related to HTTP header though).

What about removing `applyProxy`? I think proxies are handled by the `request` already?

xzyfer:
>I was working on it... 

Apologies. I had some free time and wanted to make some movement on 3.0.0.

>Just noticed an error handling here does not work (not related to HTTP header though).

I look forward to the PR ;)

>What about removing applyProxy? I think proxies are handled by the request already?

IIRC we've had a decent amount of proxy related issues. My faith in `request`s proxy handling isn't at an all time high. I think we re-asses this down the road.

PR #909, opened 2015-04-29
Improve binary file download error handling

saper:
Do not throw exception on invalid/mangled URL
Provide consistent error message with the URL
on all failures.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -21,32 +21,39 @@ require('../lib/extensions');
  */
 
 function download(url, dest, cb) {
-  var returnError = function(err) {
-    cb(typeof err.message === 'string' ? err.message : err);
+  var reportError = function(err) {
+    cb(['Cannot download "', url, '": ',
+      typeof err.message === 'string' ? err.message : err].join(''));
+  };
+  var successful = function(response) {
+    return response.statusCode >= 200 && response.statusCode < 300;
   };
 
-  if (url) {
-    applyProxy({ rejectUnauthorized: false }, function(options) {
-      options.headers = {
-        'User-Agent': [
-          'node/', process.version, ' ',
-          'node-sass-installer/', package.version
-        ].join('')
-      };
-      request.get(url, options).on('response', function(response) {
-        if (response.statusCode < 200 || response.statusCode >= 300) {
-          returnError(['Can not download file from:', url].join());
-          return;
+  applyProxy({ rejectUnauthorized: false }, function(options) {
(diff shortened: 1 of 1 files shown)


scripts/install.js
```diff
@@ -21,32 +21,37 @@ require('../lib/extensions');
  */
 
 function download(url, dest, cb) {
-  var returnError = function(err) {
-    cb(typeof err.message === 'string' ? err.message : err);
+  var reportError = function(err) {
+    cb(['Cannot download "', url, '": ',
+      typeof err.message === 'string' ? err.message : err].join(''));
```
xzyfer: Would this not cause an error if `err.message` was undefined? i.e. if `err` was a string

saper: On Mon, 4 May 2015, Michael Mifsud wrote:

> > @@ -21,32 +21,37 @@ require('../lib/extensions');
> >   */
> >  
> >  function download(url, dest, cb) {
> > -  var returnError = function(err) {
> > -    cb(typeof err.message === 'string' ? err.message : err);
> > +  var reportError = function(err) {
> > +    cb(['Cannot download "', url, '": ',
> > +      typeof err.message === 'string' ? err.message : err].join(''));
> 
> Would this not cause an error if `err.message` was undefined? i.e. if `err` was a string

````
m> node
> z = "aaaa"
'aaaa'
> typeof z.message === 'string'
false
````

Comment (xzyfer) [inline scripts/install.js]: I stand corrected :+1:

File: scripts/install.js [hunk 2]
```diff
@@ -21,32 +21,37 @@ require('../lib/extensions');
...
-          return;
+  applyProxy({ rejectUnauthorized: false }, function(options) {
+    options.headers = {
+      'User-Agent': [
+        'node/', process.version, ' ',
+        'node-sass-installer/', package.version
+      ].join('')
+    };
+    request(url, options, function(err, response, body) {
+      if (err) {
```
xzyfer: nitpick. nested `if` are hard to grok

```js
if (err) {
  reportError(err);
} else if (!successful(response)) {
  reportError(body);
} else {
  cb();
}
```

saper:
nitpick fixed. I have also added exception wrapper since it turns out https://github.com/request/request/issues/1318#issuecomment-97374120 that `request` may still throw some exceptions and not pass them to us via the callback or `on('error')`.

Also will not print full error body, it was very short (`NotFound`) for `http-server`, but some sites have much more elaborate 404 messages.

PR #910, opened 2015-04-29
Rework test coverage facility w/ istanbul

saper:
1) Use istanbul instead of jscoverage
2) generate nice HTML report with branches
3) include 'spec.js' tests in the coverage reporting - hits/line ratio on index.js ~300
4) show little coverage summary in the console output
5) do not use spawn and shell to launch mocha and coveralls clients - wire reporters directly via JavaScript

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   "devDependencies": {
     "coveralls": "^2.11.2",
     "cross-spawn": "^1.0.3",
-    "jscoverage": "^0.5.9",
+    "istanbul": "^0.3.13",
     "jshint": "^2.8.0",
     "mocha": "^2.2.5",
     "mocha-lcov-reporter": "^0.0.2",
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ build
 lib-cov
 node_modules
 vendor
+test/html-report
+test/lcov-report
+test/lcov.info
+coverage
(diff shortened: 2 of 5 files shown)


saper:
Well, it looks like travis-ci now testing [with io.js 3.0.0](https://travis-ci.org/sass/node-sass/jobs/74660239) as the "latest".

saper:
@xzyfer @am11 what do you think about this? I like not having to invoke shell to do coverage testing (although the resulting JS file is not the simplest).

package.json
```diff
@@ -64,7 +64,7 @@
   "devDependencies": {
     "coveralls": "^2.11.2",
     "cross-spawn": "^1.0.3",
-    "jscoverage": "^0.5.9",
+    "istanbul": "^0.3.13",
```
am11: nit: latest version is `0.3.18`.

am11:
LGTM. Lets see if CI agrees. :)

> Well, it looks like travis-ci now testing with io.js 3.0.0 as the "latest".

As per [docs](http://www.appveyor.com/docs/lang/nodejs-iojs), seems like we need to change this line: https://github.com/sass/node-sass/blob/f95af4/appveyor.yml#L12 to `- nodejs_version: 3` in order to have same affect in AppVeyor.

xzyfer:
I'm fine with this.

PR #915, opened 2015-05-01
Fix #912: Use runtime libsass version

### LINKED ISSUE #912: Report runtime libsass version
Currently we give simply what we have in package.info:

````
function getVersionInfo() {
  return [
           ['node-sass', package.version, '(Wrapper)', '[JavaScript]'].join('\t'),
           ['libsass  ', package.libsass, '(Sass Compiler)', '[C/C++]'].join('\t'),
  ].join(eol);
}
````

Because I often link to the libsass shared library or (for the official FreeBSD binary) I include the externally-built `libsass.a` it would be great to know which version we are running at the moment.

Also what about adding a build metadata (+FreeBSD etc.) to the version number for the official binaries?

### LINKED ISSUE #913: .renderSync err is returning a string (not an Object), .render not getting called at all
Node version `v0.12.2`, node-sass `v2.1.1`

```
try {
      data = sass.renderSync({data: data.toString(), outputStyle: 'compressed'}).css;
    } catch(e) {
      callback(e, data);
      return;
    }
```

In this case, data is a buffer.
e will be a JSON-parseable string and not an object (expected with e.message). This is non-standard.

Additionally, I'm having problems executing
```
sass.render({data: data.toString(), outputStyle: 'compressed'}, function(err, result) {
  callback(err, result && result.css);
});
```

On error or success.

The contents of `data.toString()` are as follows:

```
.main-splash-container {

  position: relative;
  display: table;
  width: 100%;
  height: 100%;
  top: 0;
  background: #999;
  background-image: linear-gradient(to top, rgba(0,0,0,0.6), transparent), url('static/images/backgrounds/header_hero_image.jpg');
  background-size: cover;
  background-position: 50% 0%;

  .main-splash-content {

    display: table-cell;
    width:100%;
    height: 85vh;
    height: calc(100vh - 116px);
    min-height: 380px;
    padding: 40px;
    text-align: center;
    vertical-align: middle;

    .main-splash-header {
      color: #FFF;
      font-weight: bold;
      font-size: 30px;
    }
    .space-categories {
      text-decoration: underline;
    }

    .main-splash-subheader {
      color: #FFF;
      font-size: 21px;
      font-weight: 400;
    }
  }
}
```

### PR DESCRIPTION
**Pull request by saper**
Return real libsass version
information - in when a shared library
is used this version may be changed
even without recompilation

libsass version in the package.json
file is now used only for checking
out the bundled libsass library
from the git submodule

PR: #912

### DIFF
--- a/src/binding.cpp
+++ b/src/binding.cpp
@@ -302,11 +302,17 @@ NAN_METHOD(render_file_sync) {
   NanReturnValue(NanNew<Boolean>(result == 0));
 }
 
+NAN_METHOD(libsass_version) {
+  NanScope();
+  NanReturnValue(NanNew<String>(libsass_version()));
+}
+
 void RegisterModule(v8::Handle<v8::Object> target) {
   NODE_SET_METHOD(target, "render", render);
   NODE_SET_METHOD(target, "renderSync", render_sync);
   NODE_SET_METHOD(target, "renderFile", render_file);
   NODE_SET_METHOD(target, "renderFileSync", render_file_sync);
+  NODE_SET_METHOD(target, "libsassVersion", libsass_version);
   SassTypes::Factory::initExports(target);
 }
 
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ console.log(result.stats);
 
 ### Version information (>= v2.0.0)
 
-Both `node-sass` and `libsass` version info is now present in `package.json` and is exposed via `info` method:
+Both `node-sass` and `libsass` version info is now exposed via the `info` method:
 
 ```javascript
 var sass = require('node-sass');
@@ -368,6 +368,8 @@ console.log(sass.info);
   libsass         3.1.0   (Sass Compiler) [C/C++]
 */
 ```
+
+Since node-sass >=v3.0.0 libsass version is determined at run time.
 
 ## Integrations
 
--- a/binding.gyp
+++ b/binding.gyp
@@ -22,7 +22,7 @@
         '<!(node -e "require(\'nan\')")',
       ],
       'conditions': [
-        ['libsass_ext == ""', {
+        ['libsass_ext == "" or libsass_ext == "no"', {
           'dependencies': [
             'src/libsass.gyp:libsass',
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: src/binding.cpp [hunk 1]
```diff
@@ -302,11 +302,17 @@ NAN_METHOD(render_file_sync) {
   NanReturnValue(NanNew<Boolean>(result == 0));
 }
 
+NAN_METHOD(libsass_version) {
+  NanScope();
+  NanReturnValue(NanNew<String>(libsass_version()));
```
xzyfer: What does this defer to, is this information exposed by Libsass?

saper: Yes, via the `libsass_version()` function. But now I see in the failing test Appveyor that it has always been set to `[NA]` on Windows and probably also on Unix (when compiled from the submodule) in the past.

xzyfer: Yeah cool that's what I figured but I missed it in a quick grep of libsass. I see it now :+1:

saper:
appveyor \o/

saper:
Just a heads-up: if you use `node-sass` JavaScript code after this change got merged and still have an old binary you will get:

````
/home/saper/node_modules/node-sass/lib/index.js:27
           ['libsass  ', binding.libsassVersion(), '(Sass Compiler)', '[C/C++]
                                 ^
TypeError: undefined is not a function
    at getVersionInfo (/home/saper/node_modules/node-sass/lib/index.js:27:34)
    at Object.<anonymous> (/home/saper/node_modules/node-sass/lib/index.js:418:28)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/saper/sw/nodes-sass-problems/924/node-sass-issue-924-example.js:2:12)
    at Module._compile (module.js:460:26)
````

The solution is to recompile binary or download 3.0.0 ones.

Issue #870, opened 2015-04-17
"Module did not self-register"on Windows 8.1, iojs 1.7.1


doodledood:
Tried to build a project using webpack, that uses babel-core which uses node-sass and got the following error: "Module did not self-register".

Unable to solve this no matter what I try.
Was reproduced on couple of machines

saper:
Which version of node-sass?

saper:
Looks like it is a problem with io.js https://github.com/iojs/io.js/issues/751 the workaround is to use `nvm` to switch between node and io.js

doodledood:
v2.1.1

xzyfer:
iojs 0.71 isn't supported as it's a old version. We currently actively support iojs 1.0+.

doodledood:
I am sorry, I completely messed up the version. I was using v1.7.1 not 0.17.1...

xzyfer:
I see. This currently an issue with the node-sass 2.x branch. It only supports iojs 1.0-1.6 (since 1.7 didn't exist at the time).
As @saper suggests you can use nvm to run multiple node/iojs versions.


Alternatively you can use the latest beta. It has a much more robust way of handling new runtime versions and shouldn't suffer from these issues in the future.

iam4x:
That's a good news @xzyfer ! Is there's an open issue I can susribe to know when the new version is out?

xzyfer:
There is not. I have however labelled this issued as `vNext` and re-opened it. I'll update this issue the next stable is released. For now the currently beta release should solve your issue. You can install it with `npm install node-sass@beta`.

doodledood:
@xzyfer Thanks for your response.
I tried using the latest beta (5) but it still gives me the same error. 
I will try reverting back the iojs version.

xzyfer:
Interesting. Could you please show the full output of a fresh `npm install node-sass@beta`.

doodledood:
> 
> > node-sass@3.0.0-beta.5 install D:\Projects\test\node_modules\node-sass
> > node scripts/install.js
> 
> Binary downloaded and installed at D:\Projects\test\node_modules\node-sass\vendor\win32-x64-43\binding.node
> 
> > node-sass@3.0.0-beta.5 postinstall D:\Projects\test\node_modules\node-sass
> > node scripts/build.js
> 
> ` D:\Projects\test\node_modules\node-sass\vendor\win32-x64-43\binding.node ` exists. 
>  testing binary.
> Problem with the binary.
[...]

PR #926, opened 2015-05-06
build: Adds win_delay_hook so iojs runs with alias

am11:
Before building, first apply the following patch
to pangyp: rvagg/pangyp#5 (until it is merged).

Issue URL: #870.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -2,6 +2,7 @@
   'targets': [
     {
       'target_name': 'binding',
+      'win_delay_load_hook': 'true',
       'sources': [
         'src/binding.cpp',
         'src/create_string.cpp',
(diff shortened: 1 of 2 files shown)


xzyfer:
Hey @am11, do we need to ship this for 3.0.0?

am11:
I think yes as this change has no side effect.
I have changed the pangyp version to my master, until they merge that PR.

am11:
This actually fixes that module registration bug on Windows. :)

xzyfer:
Argh alright. I guess I can revoke the tag since we haven't created the release :) /cc @saper if you figure out what's up with the FreeBSD watcher we can ship it in 3.0.0.

saper:
This change has no effect for us until we inherit objects from v8 or something like that, right?

xzyfer:
I believe it's related to node-sass install script failing with windows on iojs

xzyfer:
This doesn't affect the 3.0.0 node-sass-binaries so I'll ship this and re-tag v3.0.0.

PR #928, opened 2015-05-06
--help fix: mentioning options also in stdin/stdout syntax

akauppi:
When doing `node_modules/.bin/node-sass --help`, the "Usage" text didn't mention options in the stdin/stdout syntax, but they are there. The example uses them.

This adds a mention of options to the help text.


am11:
Tervetola Asko! :tada:

PR #936, opened 2015-05-07
Remove use of 'package' as a variable name in install.js

ajp427:
`package` is a reserved word in strict mode. Therefore, node-sass cannot be packaged by a bundler that forces strict mode.

This pull request removes the `package` keyword, and renames the variable to `pkg`, as it is in `scripts/build.js`.


xzyfer:
Thanks @ajp427 !

PR #940, opened 2015-05-08
Update README.md

antleblanc:
Libsass now allow us to add the missing compact and expanded output styles.


xzyfer:
Thanks @antleblanc

Issue #933, opened 2015-05-07
installation problem


Soularix:
When I try to instal node-sass it shows up :
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\\\node.exe" "C:\\Program Files\\nodejs
\\node_modules\\npm\\bin\\npm-cli.js" "install" "node-sass"
npm ERR! node v0.12.2
npm ERR! npm  v2.7.4

npm ERR! Cannot read property 'trim' of undefined
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     C:\Users\Souhaib\npm-debug.log

xzyfer:
Hi @Soularix, please follow the steps in out troubleshooting guide and produce the output.

https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md#windows-1

Soularix:
I followed the steps but I still get the same error :
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\\\node.exe" "C:\\Program Files\\nodejs
\\node_modules\\npm\\bin\\npm-cli.js" "install" "node-sass"
npm ERR! node v0.12.2
npm ERR! npm  v2.7.4

npm ERR! Cannot read property 'trim' of undefined
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
-
npm ERR! Please include the following file with any support request:
npm ERR!     C:\temp1\npm-debug.log

C:\temp1>cd node_modules\node-sass\vendor
The system cannot find the path specified.

xzyfer:
That's fine. However we need to output from those commands before we can offer any help.

Soularix:
what commands ?

xzyfer:
The commands shown on our troubleshooting guide https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md#windows-1

Please follow **every** single step and paste the output here.

xzyfer:
This is an error in npm. It's cause by `git` not being installed. This related to https://github.com/sass/node-sass/pull/931. 

@am11 @saper it would seem having a git backed dep is show stopper for Windows. We need a solution here ASAP. #931 is failing CI hard. I suggest we revert the pangyp changes until it's fixed upstream.

[...]

PR #931, opened 2015-05-07
Ensure am11/pangyp not fetched via SSH

saper:
Pin am11/pangyp to the particular commit

Should fix the problem mentioned in:
https://github.com/sass/node-sass/commit/579baf322ca48f73b04a009b70220a6c0a3cbc15#commitcomment-11082041

PR #942, opened 2015-05-09
Revert "build: Adds win_delay_hook so iojs runs with alias."

xzyfer:
This PR reverts commit 579baf322ca48f73b04a009b70220a6c0a3cbc15.

This is causing a lot of issues for people trying to install node-sass on windows. Two main culprits are proxy blocking git ports, or the local system not having git installed.

Fixes https://github.com/sass/node-sass/issues/939.
Fixes https://github.com/sass/node-sass/issues/933.
Closed https://github.com/sass/node-sass/pull/931.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "mkdirp": "^0.5.0",
     "nan": "^1.8.4",
     "npmconf": "^2.1.1",
-    "pangyp": "am11/pangyp",
+    "pangyp": "^2.1.0",
     "request": "^2.55.0",
     "sass-graph": "^2.0.0"
   },


saper:
As an alternative, what about adding iojs.exe or atomwhateceritscallednow.exe binaries to the matrix?

xzyfer:
The troubleshooting guide has an issue for the iojs.exe issue. I don't think there is an alternative since Windows installs atm are broken.

xzyfer:
Reverting this makes them less broken.

binding.gyp
```diff
@@ -2,7 +2,6 @@
   'targets': [
     {
       'target_name': 'binding',
-      'win_delay_load_hook': 'true',
```
am11: Keep this line

xzyfer: :+1:

am11:
Windows install are not broken.
We don't need alternative solution to the problem with tons of people discussed in io.js issue tracker and devised this working tangible solution.
Secondly, until we hear back from @rvagg regarding https://github.com/rvagg/pangyp/pull/5, we can revert back pangyp to npm version. For the release, I will build the binaries locally with the delay load fix. So the issue is very much contained and effect only those who use io.js on windows && manually building node-sass binary.

xzyfer:
More specifically, they're broken for Windows users without git installed.

Issue #914, opened 2015-04-30
Feature Reqeust: Store the new `.file` path if an importer returns both `.file` and  `.contents` properties


jameslnewell:
If an importer returns both `.file` and `.contents` properties then use the `.contents` property as the source but store/replace the original `.file` path with the `.file` path returned by the importer. So that, if the `.contents` contained any import statements, the next call to an importer has the correct "resolved" path of the `prev` file.

This is required to [make npm style imports](https://www.npmjs.com/package/sass-composer) work properly. When an importer returns `.contents` for an import with a `file` param using a module name e.g. `sass-grid` and the imported contents contains more import statements e.g. `sass-named-breakpoints`, when the importer is called for the new import statement with a `prev` argument of `sass-grid` which lives somewhere in node_modules, you don't have a usable `prev` path to start looking for the new import contents.

Let me know if I need to explain it clearer.

saper:
Sounds not very difficult, can you post a complete example/test case of how it should work?

saper:
( possibly without composer, just plain node-sass pls)

jameslnewell:
An example of what I'd like  (obviously won't work):

    var sass = require('node-sass');

    var i = 0;
    sass.render(
      {
        file: './index.scss',
        importer: function(file, prev, cb) {

          if (i++ === 0) {
            console.log(prev); //"./index.scss"
            cb({
              file:     './node_modules/sass-grid/mixins.scss",
              contents: '@import "sass-named-breakpoints";'
            });
          } else {
            console.log(prev); //"./node_modules/sass-grid/mixins.scss", not "sass-named-breakpoints"
            cb({
              file:     './node_modules/sass-grid/mixins.scss",
              contents: '.foobar {};' //....could possibly have more imports
            });
          }

        }
      },
      function(err, result) {
        if (err) {
          console.log(err);
        } else {
          console.log(result.css.toString());
        }
      }
    );

The `.file`, `.data` properties to the `.render()` method should probably work the same way for consistency.

mattfysh:
Let me check to see if I understand correctly. You want to be able to call `cb({file: '/a.scss', contents: 'import "b";'})` -- and then when the custom importer is called to resolve "b", the second parameter `prev` comes through as "a.scss".

We've actually managed to have this working for the last week or so, and I think the trick was to reorder the object properties as {contents, file} instead of {file, contents} - let me know if that does the trick!

jameslnewell:
[...]

PR #956, opened 2015-05-13
Fix #924: Empty data string is not a file name

### LINKED ISSUE #924: Calling render without file or data produces a cryptic error message
On master, and doing `sass.render({contents: input}, callback)`.

When `input` is an empty string, I get the error `File context created without an input path`.

saper:


test/api.js
```diff
@@ -76,6 +76,15 @@ describe('api', function() {
       });
     });
 
+    it('should NOT compile empty data string', function(done) {
+      sass.render({
+        data: ''
+      }, function(error) {
+        assert.equal(error.message, 'Data context created with empty source string');
```
xzyfer: IMO this error message is the real problem. We need to validate better in the CLI and return and error message that helps the user.

saper: I am kind of ambivalent about this:

- we are pretty thin wrapper here and I wouldn't really introduce own `node-sass` error messages if `libsass` already handles the situation
- I am not sure that having two kinds of contexts (data and file) are good design. They are actually the same structures differing just by one extra field may be very error prone. A data context with `source_string` set to zero is indistinguishable from the file context - we have to use two sets of functions to handle the distinction. We also have ambiguous situation when both `file` and `data` are set - there is a special meaning for this situation as well (see #914).

We need some work on the libsass side to have this ironed out in my opinion.

xzyfer: > we are pretty thin wrapper here

This is no longer the case in 3.x now that we have custom functions and types. We need to stop thinking this way.

>I wouldn't really introduce own node-sass error messages if libsass already handles the situation

We should serve our users as best we can. If Libsass creates a poor user experience we should fill those gaps until they're resolved upstream. It's important to understand that Libsass has different goals to us for the time being.

> I am not sure that having two kinds of contexts (data and file) is a good design

I don't disagree here. I think we should still take the two inputs (maybe more in the future?) but having two separate properties for this doesn't serve a benefit to the user IMHO.

>We need some work on the libsass side to have this ironed out in my opinion.

I agree that Libsass needs to provide a better developer experience.

Long story short I emphatically disagree that this is sane behaviour.

Any message we show to the user should clearly describe what went wrong and how to resolve it. IMHO `Data context created with empty source string` does neither.

saper:
`renderSync({})` hangs - to be investigated :) volunteers needed!

saper:
Once tests pass I think this is ready for review. I was afraid there was another binding bug hiding, but fortunately it was just a missing `done()` in the mocha script. @xzyfer

lib/index.js
```diff
@@ -351,7 +352,13 @@ module.exports.render = function(options, cb) {
     });
   }
 
-  return options.data ? binding.render(options) : binding.renderFile(options);
+  if (options.hasOwnProperty('data')) {
+    binding.render(options);
+  } else if (options.hasOwnProperty('file')) {
+    binding.renderFile(options);
+  } else {
+    cb({status: 3, message: 'No input specified: provide a file name or a source string to process' });
```
xzyfer: :heart_eyes:

xzyfer:
:+1:

lib/index.js
```diff
@@ -351,7 +352,13 @@ module.exports.render = function(options, cb) {
     });
   }
 
-  return options.data ? binding.render(options) : binding.renderFile(options);
+  if (options.hasOwnProperty('data')) {
```
xzyfer: IMHO these checks are too lax. 

```
if (options.hasOwnProperty('data') && options.data) {
  binding.render(options);
} else if (options.hasOwnProperty('file') && options.file) {
  binding.renderFile(options);
} else {
 ...
}
```

A use should _never_ see

> Data context created with empty source string

lib/index.js
```diff
@@ -398,7 +405,15 @@ module.exports.renderSync = function(options) {
     });
   }
 
-  var status = options.data ? binding.renderSync(options) : binding.renderFileSync(options);
+  var status;
+  if (options.hasOwnProperty('data')) {
```
xzyfer: ditto

xzyfer:
The new error message is big improvement. However I'm still argue that under no circumstance should a user receive this error message if we can be prevented (which in this case it can)

>Data context created with empty source string

saper:
In that case, I think `if (options.data) { ... ` and `if (options.file) { ... ` are enough?

xzyfer:
Agreed

saper:
in that case the last commit should be ok

PR #992, opened 2015-06-09
Bump Libsass to 3.2.5

xzyfer:
This PR bumps Libsass to 3.2.5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "node-sass",
   "version": "3.1.2",
-  "libsass": "3.2.4",
+  "libsass": "3.2.5",
   "description": "Wrapper around libsass",
   "license": "MIT",
   "homepage": "https://github.com/sass/node-sass",


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "node-sass",
   "version": "3.1.2",
-  "libsass": "3.2.4",
+  "libsass": "3.2.5",
   "description": "Wrapper around libsass",
   "license": "MIT",
   "homepage": "https://github.com/sass/node-sass",
```
am11: It is probably a good idea to keep bumping the dependencies versions with every release (even the minor/patch too). I know that while installation, npm takes `~` into account and installs the latest patch, but still it feels right to keep everything to its bleeding edge. :)

Here is the list of changes can be made according to https://david-dm.org/sass/node-sass:

```
--- deps ---
glob: 5.0.10
mkdirp: 0.5.1
npmconf: 2.1.2
pangyp: 2.2.1
request: 2.57.0

--- dev deps ---
cross-spawn: 0.4.1
jshint: 2.8.0
mocha: 2.2.5
rimraf: 2.4.0
```

xzyfer: Yep this is something we should do regularly.

Issue #894, opened 2015-04-26
The last `.contents` property returned by an importer is always used


jameslnewell:
When an importer function returns the `.contents` string, node-sass always uses the last returned value is used.

The following example imports a file called `foobar` twice with a different `.contents` returned each time. Node-sass always uses the last `.contents` for both imports resulting in `.barfoo` being repeated twice.

    var sass = require('node-sass');
    
    var i = 0;
    
    sass.render(
      {
        data: '@import "foobar"; @import "foobar";',
        importer: function(file, prev, cb) {
          console.log(file, prev);
    
          if (i++ === 0) {
            cb({contents: '.foobar {color: red;}'});
          } else {
            cb({contents: '.barfoo {color: blue;}'});
          }
    
        }
      },
      function(err, result) {
        if (err) {
          console.log(err);
        } else {
          console.log(result.css.toString());
        }
      }
    );

Note: Using a `.file` result for one import and a `.contents` result for the other import does not result in the same issue.

I'm using node-sass `3.0.0-beta.7`, libsass `3.2.0-beta.6` and running on OS X 10.10.3.

xzyfer:
I haven't had a chance to look into this yet.

Could this be a race condition? Can you confirm both callbacks are being
called.

saper:
confirmed here.

xzyfer:
Thank @jameslnewell and @saper. I'm scheduling this for v3.next.

saper:
Early research indicates this is a bug in `libsass`. Will follow up.

xzyfer:
After some digging I've confirmed this actually not the case. Custom importers are working as expected.

The test cases provided illustrates a feature of LibSass which is caching of import resolution. Custom importers must be idempotent, meaning that given the same input they **must** return the same output. Given this restriction the result of the importer is cached within LibSass.

saper:
@xzyfer I have a question:

 https://github.com/sass/libsass/wiki/API-Sass-Importer says

> Imports have to be relative from the parent import context. Therefore we need to pass this information to the importer callback. This is currently done by passing the complete import string/path of the previous import context.

If the imports are idempotent, why do we have `prev` parameter then?

xzyfer:
I expect only @mgreter knows the answer to that question.

saper:
Ruby has two methods (http://sass-lang.com/documentation/Sass/Importers/Base.html):

````
(Sass::Engine?) find(uri, options) 
[...]

PR #998, opened 2015-06-10
Adding a test for importer callback order

beatak:
This pull request is to add a test case for https://github.com/sass/libsass/issues/1262 and this test is currently failing.  I have chatted with @xzyfer and he thinks it's a bug.

Besides the main point (the callback order is not DFS), I also found that the importer won't callback for the nested nodes.

This test add `index.scss` and tree is as follows:
```
                   +-----+            
                   |index|            
                   +--+--+            
                      |               
         +-------------------------+  
        +-+        +------+       +-+ 
        |a|        |common|       |b| 
        +++        +--+---+       +++ 
         |            |            |  
   +-----+--+     +---+----+       |  
+------+  +--+  +----+ +------+   ++-+
|common|  |a1|  |vars| |struct|   |b1|
+--+---+  +--+  +----+ +------+   +--+
   |                                  
  +--------+                          
+----+ +------+                       
|vars| |struct|                       
+----+ +------+                       
```

Actually, `node-sass` calls nodes in the order of `[a, common, b, common, a1, vars, struct, b1]` instead of `[a, common, vars, struct, a1, common, vars, struct, b, b1]`  Not only the order is wrong, but also `vars` and `struct` didn't get called twice.


xzyfer:
Great, I'll dig into this with you tomorrow. I looked into it briefly for https://github.com/sass/libsass/issues/1262 but was not able to reproduce as per https://github.com/sass/libsass/issues/1262#issuecomment-110981379.

xzyfer:
I wonder if you're running into https://github.com/sass/node-sass/issues/894

xzyfer:
I've been able to confirm this is an issue with custom importers only. Libsass is resolving `@import`s depth-first, but custom importers are being call breadth-first.

xzyfer:
Superseded by #1420

Issue #997, opened 2015-06-10
Synchronous custom importer not providing done


DominicBarnard:
As I understand from the docs, a custom importer should be able to handle both `render` and `renderSync` by calling `done()` asynchronously.  So a single importer can load files with the async `readFile()` and be handled correctly by both `render` and `renderSync`.

If this is correct then I'm seeing a bug as when I run the following code, `done` is showing up as `undefined`.  This is what I believe is causing errors in the current implementation of [eyeglass](https://github.com/sass-eyeglass/eyeglass) when run with `renderSync` which relies on calling `done`.  

``` js
sass.renderSync({
    file: 'sass/test.scss',
    importer: function (uri, prev, done) {
        console.log(done); \\ logs undefined
    }
}, function(err, result) {
    console.log('successfully rendered');
});
```

DominicBarnard:
Related to https://github.com/sass/node-sass/issues/849, https://github.com/sass-eyeglass/eyeglass/pull/30

DominicBarnard:
[Some failing tests](https://github.com/DominicBarnard/node-sass/commit/6c35e735375a10ca506b357442a88536fe10db90)

saper:
Looks like we pass the callback bridge as the third parameter (from C++, not to your JS callback), but the values are not used. Need to investigate.

DominicBarnard:
So far I've tried https://github.com/DominicBarnard/node-sass/commit/812c5b88860f34fd4bc8ce128dc68c7c24c131b1 but my new tests are still failing :/
`Error: file to import not found or unreadable: foo
Current dir: `

saper:
You can cheat it this way https://github.com/saper/node-sass/commit/1417658f81ee54a560ddf9f36e8fab2b5cee2d2d

DominicBarnard:
ok so that passes my tests now.  But I expect that will break when an importer actually uses `readFile` to async load a file?

saper:
On Wed, 10 Jun 2015, Dominic Barnard wrote:

> ok so that passes my tests now.  But I expect that will break when an importer actually uses `readFile` to async load a file?

Your importer might return before your callback gets a chance to return something meaningful... Needs
to be tested, it's too late for me to guess

DominicBarnard:
Yes [this test](https://github.com/DominicBarnard/node-sass/commit/11a707bc2d5614e47b5872ecdc1ee6d1d385b7df) is failing.

DominicBarnard:
[...]

Issue #857, opened 2015-04-13
Render with custom import hangs


xzyfer:
Given the following `fs.readFile` never executes it's callback and process hang indefinitely.

```scss
// entry.scss
@import "_partial";
```
```scss
// _partial.scss
foo{}
```
```js
var fs = require('fs');
var sass = require('node-sass');

var files = [
    'entry.scss',
    'entry.scss',
    'entry.scss',
    'entry.scss',
];

files.forEach(function (file) {
    sass.render({
        file: file,
        importer: function (uri, prev, done) {
            var filename = process.cwd() + '/_partial.scss';
            console.log('fs.readFile:', filename);
            fs.readFile(filename, function(err, data) {
                console.log('I am never executed');
                done({ contents: data });
            });
        }
    }, function(err, result) {
        console.log('successfully rendered');
    });
});
```

Outputs this and hangs

```
➜  node-sass-test  node index.js
fs.readFile: /Users/michael/node-sass-test/_partial.scss
fs.readFile: /Users/michael/node-sass-test/_partial.scss
fs.readFile: /Users/michael/node-sass-test/_partial.scss
fs.readFile: /Users/michael/node-sass-test/_partial.scss
```

This example works fine if:
- I only have 3 files
- I use `renderSync`
- I use `fs.readFileSync`

This is potentially related to https://github.com/sass/node-sass/issues/794.

saper:
I think I have it identified. Workaround: `env UV_THREADPOOL_SIZE=5 node indes.js` :)

xzyfer:
Nice, I'll dig into it. I've just started reading the node addon docs :)

xzyfer:
Looks like that's the legit solution haha nice one.

Ideally we could catch when the queue was full and handle it more gracefully.

saper:
I have a fix in my head just need to go via the usual write-compile-test-repeat cycle :)
Good news maybe we can even simplify code this time

xzyfer:
I'd love to see it. I've been reading over the relevant docs but I can't a way to increase the thread pool during execution or stop the queue from being flooded.

xzyfer:
This seems to suggest it can be done in runtime if it's early enough

https://github.com/joyent/libuv/issues/649#issuecomment-16980782
https://github.com/mapbox/mapbox-studio/issues/16#issuecomment-17034522

saper:
this is a bug. it just explains why 4 (the default) seems to the magic number.

xzyfer:
[...]

PR #1000, opened 2015-06-11
renderSync vs async custom importers

DominicBarnard:
## Original problem

A callback function was not being passed on to custom importers when using renderSync.  Also renderSync is not set up to stop and wait for an importer to complete async operation before moving on with the render.

The original issue https://github.com/sass/node-sass/issues/997
## Proposal
- Ensure `done` is always supplied to importers and handle firing the callback if it is called synchronously (doesn't actually have to wait for anything) 
- Emit an error if an importer doesn't return (or syncly callback) any results when in `renderSync` mode (assume it tried to do something it's waiting for). 
- Supply a `mode` flag indicating either `'sync'` or `'async'` to importers on the `options` object to allow importers to decide how to execute in current context.

Under this scheme, [eyeglass's importer could look something like this](https://github.com/DominicBarnard/eyeglass/blob/1cd44eb60ef4ef30abe26393fa106530246a67b1/lib/module_importer.js) (providing it's important for it to read files asyncly).  Basically always calling `done()` to complete and deciding elsewhere in the file to use `readFile()` or `readFileSync`.
## Alternatives
- Maybe use something like [deasync](https://www.npmjs.com/package/deasync) to hold execution and wait for async importers to return.
- I'm missing some way to make async work?
- Maybe there is no reason to do anything async in an importer?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -219,7 +219,7 @@ function establishImporter(target, options) {
     if (result) {
       done(result === module.exports.NULL ? null : result);
     }
-  }
+  };
 }
 
 /**


lib/index.js
```diff
@@ -201,6 +201,53 @@ function getOptions(options, cb) {
...
+      resultCb = data;
+    }
+
+    var result = target.call(options.context, file, prev, done);
+    result = (typeof resultCb !== 'undefined') ? resultCb : result;
+
+    if(typeof result !== 'undefined') {
+      return result === module.exports.NULL ? null : result;
+    }
+    return new Error('no value returned by importer: possibly due to async operation');
```
xzyfer: I'm not a fan of this error message. Given the use case for this message to appear there isn't much the user can do about it.

We should provide enough context for the user to action on this error message. At a minimum we should should supply the file that failed to import.

xzyfer:
All in all I like where this going. Great work @DominicBarnard 

Next steps I see
- the error handling needs work. You can see my feelings about developer UX in https://github.com/sass/node-sass/pull/956#discussion_r33430645
- the git history needs some love. I think you can just squash this PR to a single commit.

xzyfer:
@am11 @saper I like your thoughts on this PR

xzyfer:
> Maybe there is no reason to do anything async in an importer

I have thought about this as well. We currently have issues with async importers due to node's threadpool limit https://github.com/sass/node-sass/issues/857. Going sync has some performance implications but it entirely works around this issue.

Maybe @chriseppstein has some practical thoughts here from his eyeglass work.

am11:
The idea of `renderSync` is to `return` instead of `done`. If you want to wait, you need `render` (async). If you have other use-case, which requires synchronous op, you use renderSync through and through.

IMHO, we shouldn't mix the expected behavior of `renderSync` to accommodate callback, when we have async functionality available.

DominicBarnard:
@am11 I agree that using renderSync through and through is preferred, which works fine when you are creating your own custom importer to work with your own project.  The concern is the added complexity of handling both sync and async scenarios when creating 3rd party importers like [eyeglass](https://github.com/sass-eyeglass/eyeglass).

For example [here is the eyeglass importer when it originally had to handle this with both return and done callback](https://github.com/sass-eyeglass/eyeglass/blob/1450c24aae65753906ca96de38b09afdfd9c5ed1/lib/module_importer.js#L142). 

Notice there is a lot of 

```js
if(done){
  ...
}else{
  return ...
}
```

You can compare that to [my suggested version of the same importer implemented against this PR](https://github.com/DominicBarnard/eyeglass/blob/renderSync-imports/lib/module_importer.js#L126).
There you can see that the importer function can simply always exit with callback and use `this.options.mode` to differentiate at the point of choosing between sync and async ops.

I suppose the difference is a matter of preferred style.  Personally i tend towards offering flexibility to the user.

am11:
@DominicBarnard, I totally get the sentiment "why not *that* option as well". 

Please see: https://github.com/sass/node-sass/issues/530#issuecomment-67771373

The reality is, I initially implemented the custom importers with unified sync and asyn signatures (and a unified base), but after spending days investigating the weird/undefined behaviors it turned out the rabbit hole goes much deeper. v8 inherently binds with libuv in a way that you cannot await the synchronous call (not with uv or C++ mutices). Someone told me that it is not possible on libuv's IRC channel, but I still managed to learn it the hard way. :)

With that said, yes we can pass the `done` cb to `renderSync` at the risk of getting misused by the consumer. Since we have discussed the pros and cons, I will leave the final decision to @xzyfer.

maoberlehner:
What is the current status of this issue? I created some custom importers (most notably: https://github.com/maoberlehner/node-sass-magic-importer) and noticed the problem today. Should I rewrite my custom importer packages to work synchronously or wait for a fix in node-sass?

saper:
@maoberlehner I don't believe this is going to be changed. Just use `render()` for the async callbacks and `renderSync()` if you expect the renderer to wait. 

I think we should change the way the C++ bridge behaves. There is too much JavaScript wrapping the C++ bridge already, I think we should simplify it and by the way provide a better importer interface.

It is unlikely I'll find time to work on the bridge until the end of this year, though.

PR #1006, opened 2015-06-20
Support libsass string quotes in the c/js api bindings.

chriseppstein:
This code compiles but the tests fail and I'm not sure why -- I think it might be a libsass bug. I could use some help debugging it.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -1228,6 +1228,29 @@ describe('api', function() {
       done();
     });
 
+    it('should handle quoted strings', function(done) {
+      var result = sass.renderSync({
+        data: '.swapped { result: swap-quotes("asdf", qwerty); }',
+        functions: {
+          'swap-quotes($s1, $s2)': function(s1, s2) {
+            s1.setQuoted(!s1.isQuoted());
+            s2.setQuoted(!s2.isQuoted());
+            console.log("s1: " + s1.isQuoted());
+            console.log("s2: " + s2.isQuoted());
+            var list = new sass.types.List(2);
+            list.setValue(0, s1);
+            list.setValue(1, s2);
+            list.setSeparator(true);
+
+            return list;
+          }
+        }
+      });
+
+      assert.equal(result.css.toString().trim(), '.swapped { result: asdf, "qwerty"; }');
+      done();
+    });
+
     it('should let custom function invoke sass types constructors without the `new` keyword', function(done) {
       var result = sass.renderSync({
         data: 'div { color: foo(); }',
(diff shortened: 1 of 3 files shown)


chriseppstein:
@matryo @mgreter @xzyfer @saper any ideas about how to make this work? It's blocking us here at work I can't wait for the big refactor.

saper:
One reason why it fails because quoted flag gets discarded from time to time...

xzyfer:
I'm jumping on a plane to SF in a couple hours, I'll take a look at a
sensible hour today.

Regards,
Michael

saper:
Well, basically quoted string support in `libsass` for custom functions is not implemented...

test/api.js
```diff
@@ -1228,6 +1228,29 @@ describe('api', function() {
...
+            list.setValue(0, s1);
+            list.setValue(1, s2);
+            list.setSeparator(true);
+
+            return list;
+          }
+        }
+      });
+
+      assert.equal(result.css.toString().trim(), '.swapped { result: asdf, "qwerty"; }');
```
saper: Is this result: `.swapped {\n  result: asdf, "qwerty"; }` acceptable?

saper:
C SASS API kind of expects that a quoted string will be, ... well quoted.

So it's not just a flag to be set, one actually needs to convert the `mystring` to `"mystring"\ or `'mystring'`.

saper:
I have prepared an early set of changes to libsass to make quoted strings work:

1) https://github.com/chriseppstein/node-sass/pull/1 modifies the test to provide an actual quoted string from the client (based on how currently API is constructed),

2) https://github.com/saper/libsass/tree/clonefix fixes passing of quoted strings via the C API in libsass.

test/api.js
chriseppstein: Yes, that seems right.

saper:
@chriseppstein Can you try applying https://github.com/sass/libsass/pull/1289 to libsass? May not fix all quoted-vs-unquoted string issues, but certainly lets us pass them over to node and back.

mgreter:
@chriseppstein @saper IMO the merged PR on libsass side done by @saper does the correct thing. I can also give you a little background info why we have two string types (ATM). Strings that appear in sass are not always unquoted. There are "instances" that will always remain static (String_Constant) up to the output (preserving white-space and "comments" too). That's probably why you always want to deal with String_Quoted on the C-API side (looks like it simply never got implemented).

String_Quoted should work pretty straight forward from there on. You pass the "parsed" string to the constructor and it will be unquoted internally (storing the unquoted value and the quotemark flag). As a side note, I'm getting more confident that we could use a bool flag for the quotemark. We previousely had wrongly parsed some String_Constants as String_Quotes, which made it necessary to preserve the actual quotemark to pass the specs. Now we parse more of them correctly as String_Constant and should be able to get rid of explicitly storing the quotemark char, and instead always rely on autoquote rules.

saper:
Ah now I start to understand. So maybe "String_Quoted" should not be derived class of String_Constant? Are they in some circumstances interchangeable? Because in that case we should use common base class (I can see there is "String")... need to read more...

PR #1022, opened 2015-07-02
Correct typeof reference in README

sharkbrainguy:
JavaScript's typeof operator returns "string" not "String" for string values and objects


keithamus:
:+1:

Issue #1030, opened 2015-07-11
Cannot read property 'binarySite' of undefined in getBinaryUrl() -> node-sass/lib/extensions.js:107:32


wwi-johnbuhay:
Here are the gists that includes the steps from the troubleshooting guide:

https://gist.github.com/wwi-johnbuhay/6fb5e7fe4518dca28548

saper:
Thank you. Did you have any previous version of node-sass installed before? Can you check the contents of npm cache? I think it's a bug we have but I'd like to investigate further. How does your cached `package.json` look like in the NPM cache?

wwi-johnbuhay:
Thanks for the help,

I did not have a previous version of node-sass before, to provide this debugging info I made sure to start from a clean environment.

I've added the contents of the npm cached package.json here:
https://gist.github.com/wwi-johnbuhay/6fb5e7fe4518dca28548#file-npm_node-sass_3-2-0_package_package-json

wwi-johnbuhay:
I just compared the package.json between the problematic one above and the one that installs successfully in the node:latest docker container, which i did not configure to use our artifactory registry.

The missing lines in the problematic package.json are:

"gitHead": "6343cea64ed6d33e2f37ff388888f6a3a189c8a8",
    "gypfile": true,
...
"libsass": "3.2.5",
...
"nodeSassConfig": {
        "binarySite": "https://github.com/sass/node-sass/releases/download"
    },

saper:
That's interesting. It's `nodeSassConfig` that is missing indeed. Is your custom registry not keeping custom properties stored?

saper:
Can you try `npm install npm-registry` and try this script? https://gist.github.com/578d44b6e02753e0008e

wwi-johnbuhay:
It appears that way, not keeping custom properties stored. I will reach out to them about it after this.

Here is the gist using the artifactory registry https://gist.github.com/wwi-johnbuhay/6fb5e7fe4518dca28548#file-npm-registry

saper:
Maybe it's not "The Most Powerful Repository Around" :) Do hey have a public bug tracker?

The second case is strange - I just removed a whole `nodeSassConfig` stanza from `package.json` and that's what I got:

````
$ git diff -- package.json
diff --git a/package.json b/package.json
index f1dc6dc..9befb1b 100644
--- a/package.json
+++ b/package.json
@@ -18,9 +18,6 @@
     "node": ">=0.10.0"
   },
[...]

PR #1032, opened 2015-07-11
Provide nodeSassConfig hardcoded defaults

saper:
Some private npm repositiories seem not to carry
custom package.json properties around.

Fixes: #1030

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -104,7 +104,9 @@ function getBinaryName() {
 function getBinaryUrl() {
   var site = flags['--sass-binary-site'] ||
              process.env.SASS_BINARY_SITE  ||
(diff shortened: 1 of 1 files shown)


xzyfer:
Although this does fix #1030, I'd like to get to the root cause of the issue. As per https://github.com/sass/node-sass/issues/1030#issuecomment-120795170.

saper:
We don't know; might be some replication issue. We already provide hardcoded default for the other variable, so I think it is okay to provide fallback here as well.

saper:
#1030 has been fully clarified, but I think we can add this (in fact the fallback was there before)

xzyfer:
It bums me out that this is needed but ok

PR #1040, opened 2015-07-21
Import libsass 3.3 code

saper:
From now on, use includes from
src/libsass/include and the sources from
src/libsass/src

This is to be included when merging newer libsass.

PR #1054, opened 2015-08-05
Update bindings for io.js 3.0.0

### LINKED ISSUE #1053: Support iojs v3
iojs v3 was released today, this means we need to update to nan v2 which has brought some massive breaking changes.

saper:
## Update to io.js 3.0.0, use nan 2.0

The code has been updated to meet
the requirements of the updated V8 engine
as provided by io.js.

In some cases additional `EscapableHandleScope`
has been provided.

`SassValueWrapper<T>::New` has been changed:
in case of instance construction via
the function call `sass.types.Color(0,0,0)` 
(actually incorrect JavaScript...)
instead of `new sass.types.Color(0,0,0)`
the `new` part is no longer executed by
falling through. This should not be necessary
because the constructor is called by
the `NewInstance` call anyway.

This fixes the following error:

    FATAL ERROR: v8::Object::SetAlignedPointerInInternalField() Internal field out of bounds


saper:
\o/

saper:
It looks like this change is ready now for wider testing and review. 

The coverage is [down by one line in index.js](https://coveralls.io/builds/3252667/source?filename=lib%2Findex.js#L216) which is interesting, because we no longer throw errors
as bare strings. This can be the consequence of the change in the `SassValueWrapper<T>::New`
to what I now believe is the correct behaviour. 

@matryo you might want to have a look at this (I put my comments inline).

xzyfer:
Amazing work @saper. I've just come back from my trip. Testing this on OSX now.

xzyfer:
Things look good on OSX. I'll give Windows a whirl and start prepping a new release.

src/binding.cpp
```diff
@@ -37,12 +37,12 @@ union Sass_Value* sass_custom_function(const union Sass_Value* s_args, Sass_Func
...
 
-void ExtractOptions(Local<Object> options, void* cptr, sass_context_wrapper* ctx_w, bool is_file, bool is_sync) {
-  NanScope();
+void ExtractOptions(v8::Local<v8::Object> options, void* cptr, sass_context_wrapper* ctx_w, bool is_file, bool is_sync) {
+  Nan::HandleScope scope;
 
   struct Sass_Context* ctx;
 
-  NanAssignPersistent(ctx_w->result, options->Get(NanNew("result"))->ToObject());
+  ctx_w->result.Reset(Nan::Get(options, Nan::New("result").ToLocalChecked()).ToLocalChecked()->ToObject());
```
kkoopa: `ctx_w->result.Reset(Nan::To<v8::Object>(Nan::Get(options, Nan::New("result").ToLocalChecked()).ToLocalChecked()).ToLocalChecked());`

src/binding.cpp
```diff
@@ -61,52 +61,52 @@ void ExtractOptions(Local<Object> options, void* cptr, sass_context_wrapper* ctx
...
   }
 
   if (!is_file) {
-    ctx_w->file = create_string(options->Get(NanNew("file")));
+    ctx_w->file = create_string(Nan::Get(options, Nan::New("file").ToLocalChecked()));
     sass_option_set_input_path(sass_options, ctx_w->file);
   }
 
-  int indent_len = options->Get(NanNew("indentWidth"))->Int32Value();
+  int indent_len = Nan::Get(options, Nan::New("indentWidth").ToLocalChecked()).ToLocalChecked()->Int32Value();
```
kkoopa: `int indent_len = Nan::To<int32_t>(Nan::Get(options, Nan::New("indentWidth").ToLocalChecked()).ToLocalChecked()).FromJust();`

src/binding.cpp
```diff
@@ -61,52 +61,52 @@ void ExtractOptions(Local<Object> options, void* cptr, sass_context_wrapper* ctx
...
 
   ctx_w->indent = (char*)malloc(indent_len + 1);
 
   strcpy(ctx_w->indent, std::string(
     indent_len,
-    options->Get(NanNew("indentType"))->Int32Value() == 1 ? '\t' : ' '
+    Nan::Get(options, Nan::New("indentType").ToLocalChecked()).ToLocalChecked()->Int32Value() == 1 ? '\t' : ' '
```
kkoopa: ditto

src/binding.cpp
```diff
@@ -61,52 +61,52 @@ void ExtractOptions(Local<Object> options, void* cptr, sass_context_wrapper* ctx
...
-  sass_option_set_source_map_embed(sass_options, options->Get(NanNew("sourceMapEmbed"))->BooleanValue());
-  sass_option_set_source_map_contents(sass_options, options->Get(NanNew("sourceMapContents"))->BooleanValue());
+  sass_option_set_output_style(sass_options, (Sass_Output_Style)Nan::Get(options, Nan::New("style").ToLocalChecked()).ToLocalChecked()->Int32Value());
+  sass_option_set_is_indented_syntax_src(sass_options, Nan::Get(options, Nan::New("indentedSyntax").ToLocalChecked()).ToLocalChecked()->BooleanValue());
+  sass_option_set_source_comments(sass_options, Nan::Get(options, Nan::New("sourceComments").ToLocalChecked()).ToLocalChecked()->BooleanValue());
+  sass_option_set_omit_source_map_url(sass_options, Nan::Get(options, Nan::New("omitSourceMapUrl").ToLocalChecked()).ToLocalChecked()->BooleanValue());
+  sass_option_set_source_map_embed(sass_options, Nan::Get(options, Nan::New("sourceMapEmbed").ToLocalChecked()).ToLocalChecked()->BooleanValue());
+  sass_option_set_source_map_contents(sass_options, Nan::Get(options, Nan::New("sourceMapContents").ToLocalChecked()).ToLocalChecked()->BooleanValue());
```
kkoopa: I'll stop commenting on these now.

saper: On Sun, 9 Aug 2015, Benjamin Byholm wrote:

> >  
> >    sass_option_set_output_path(sass_options, ctx_w->out_file);
> > -  sass_option_set_output_style(sass_options, (Sass_Output_Style)options->Get(NanNew("style"))->Int32Value());
> > -  sass_option_set_is_indented_syntax_src(sass_options, options->Get(NanNew("indentedSyntax"))->BooleanValue());
> > -  sass_option_set_source_comments(sass_options, options->Get(NanNew("sourceComments"))->BooleanValue());
> > -  sass_option_set_omit_source_map_url(sass_options, options->Get(NanNew("omitSourceMapUrl"))->BooleanValue());
> > -  sass_option_set_source_map_embed(sass_options, options->Get(NanNew("sourceMapEmbed"))->BooleanValue());
> > -  sass_option_set_source_map_contents(sass_options, options->Get(NanNew("sourceMapContents"))->BooleanValue());
> > +  sass_option_set_output_style(sass_options, (Sass_Output_Style)Nan::Get(options, Nan::New("style").ToLocalChecked()).ToLocalChecked()->Int32Value());
> > +  sass_option_set_is_indented_syntax_src(sass_options, Nan::Get(options, Nan::New("indentedSyntax").ToLocalChecked()).ToLocalChecked()->BooleanValue());
> > +  sass_option_set_source_comments(sass_options, Nan::Get(options, Nan::New("sourceComments").ToLocalChecked()).ToLocalChecked()->BooleanValue());
> > +  sass_option_set_omit_source_map_url(sass_options, Nan::Get(options, Nan::New("omitSourceMapUrl").ToLocalChecked()).ToLocalChecked()->BooleanValue());
> > +  sass_option_set_source_map_embed(sass_options, Nan::Get(options, Nan::New("sourceMapEmbed").ToLocalChecked()).ToLocalChecked()->BooleanValue());
> > +  sass_option_set_source_map_contents(sass_options, Nan::Get(options, Nan::New("sourceMapContents").ToLocalChecked()).ToLocalChecked()->BooleanValue());
> 
> I'll stop commenting on these now.

Thanks, will update all of this accordingly.

src/binding.cpp
```diff
@@ -241,14 +243,13 @@ NAN_METHOD(render) {
 
   assert(status == 0);
 
-  NanReturnUndefined();
+  return;
```
kkoopa: unnecessary

src/binding.cpp
```diff
@@ -261,14 +262,13 @@ NAN_METHOD(render_sync) {
 
   sass_free_context_wrapper(ctx_w);
 
-  NanReturnValue(NanNew<Boolean>(result == 0));
+  info.GetReturnValue().Set(Nan::New<v8::Boolean>(result == 0));
```
kkoopa: can do `info.GetReturnValue().Set(result == 0);` directly

src/binding.cpp
```diff
@@ -278,14 +278,13 @@ NAN_METHOD(render_file) {
 
   assert(status == 0);
 
-  NanReturnUndefined();
+  return;
```
kkoopa: unnecessary

src/binding.cpp
```diff
@@ -298,20 +297,19 @@ NAN_METHOD(render_file_sync) {
...
+  info.GetReturnValue().Set(Nan::New<v8::String>(libsass_version()).ToLocalChecked());
 }
 
-void RegisterModule(v8::Handle<v8::Object> target) {
-  NODE_SET_METHOD(target, "render", render);
-  NODE_SET_METHOD(target, "renderSync", render_sync);
-  NODE_SET_METHOD(target, "renderFile", render_file);
-  NODE_SET_METHOD(target, "renderFileSync", render_file_sync);
-  NODE_SET_METHOD(target, "libsassVersion", libsass_version);
+void RegisterModule(v8::Local<v8::Object> target) {
```
kkoopa: `NAN_MODULE_INIT(RegisterModule)` is better.

src/binding.cpp
saper: On Sun, 9 Aug 2015, Benjamin Byholm wrote:

> > @@ -261,14 +262,13 @@ NAN_METHOD(render_sync) {
> >  
> >    sass_free_context_wrapper(ctx_w);
> >  
> > -  NanReturnValue(NanNew<Boolean>(result == 0));
> > +  info.GetReturnValue().Set(Nan::New<v8::Boolean>(result == 0));
> 
> can do `info.GetReturnValue().Set(result == 0);` directly

Is it Nan::New never necessary in those cases? (also for strings etc.?)
In that case we have plenty of these, will remove.

Thanks for taking time to review this!

src/create_string.cpp
```diff
@@ -3,12 +3,18 @@
...
+  if (maybevalue.ToLocal(&value)) {
+    if (value->IsNull() || !value->IsString()) {
+      return 0;
+    }
+  } else {
     return 0;
   }
 
-  String::Utf8Value string(value);
+  v8::String::Utf8Value string(value);
```
kkoopa: `Nan::Utf8String` is better, since it handles invalid UTF-8 in a sensible way.

src/sass_types/boolean.cpp
```diff
@@ -15,59 +13,63 @@ namespace SassTypes
...
 
-      tpl->SetClassName(NanNew("SassBoolean"));
+      tpl->SetClassName(Nan::New("SassBoolean").ToLocalChecked());
       tpl->InstanceTemplate()->SetInternalFieldCount(1);
-      tpl->PrototypeTemplate()->Set(NanNew("getValue"), NanNew<FunctionTemplate>(GetValue)->GetFunction());
+      Nan::SetPrototypeTemplate(tpl, "getValue", Nan::GetFunction(Nan::New<v8::FunctionTemplate>(GetValue)).ToLocalChecked());
 
-      NanAssignPersistent(constructor, tpl->GetFunction());
+      constructor.Reset(Nan::GetFunction(tpl).ToLocalChecked());
+      conslocal = Nan::New(constructor);
```
kkoopa:
```c++
conslocal = Nan::GetFunction(tpl).ToLocalChecked();
constructor.Reset(conslocal);
```
is better

src/sass_types/boolean.cpp
```diff
@@ -15,59 +13,63 @@ namespace SassTypes
...
   Sass_Value* Boolean::get_sass_value() {
     return sass_make_boolean(value);
   }
 
-  Local<Object> Boolean::get_js_object() {
-    return NanNew(this->js_object);
+  v8::Local<v8::Object> Boolean::get_js_object() {
+    Nan::EscapableHandleScope scope;
+    return scope.Escape(Nan::New(this->js_object));
   }
```
kkoopa: Since this only creates one new handle, you can safely omit the scope if you inline it.
```c++
inline v8::Local<v8::Object> Boolean::get_js_object() {
  return Nan::New(this->js_object);
}
```

src/sass_types/color.cpp
```diff
@@ -52,86 +50,82 @@ namespace SassTypes
...
-    NanScope();
-    NanReturnValue(NanNew(sass_color_get_a(unwrap(args.This())->value)));
+    info.GetReturnValue().Set(Nan::New(sass_color_get_a(unwrap(info.This())->value)));
   }
 
   NAN_METHOD(Color::SetR) {
-    if (args.Length() != 1) {
-      return NanThrowError(NanNew("Expected just one argument"));
+    if (info.Length() != 1) {
+      return Nan::ThrowError(Nan::New("Expected just one argument").ToLocalChecked());
```
kkoopa: You can do `return Nan::ThrowError("Expected just one argument");` directly, but even better would be to throw the correct subtype where appropriate: Error, RangeError, ReferenceError, SyntaxError, TypeError

src/sass_types/factory.cpp
```diff
@@ -45,26 +43,27 @@ namespace SassTypes
     }
   }
 
-  void Factory::initExports(Handle<Object> exports) {
-    Local<Object> types = NanNew<Object>();
-    exports->Set(NanNew("types"), types);
+  void Factory::initExports(v8::Local<v8::Object> exports) {
+    Nan::EscapableHandleScope scope;
```
kkoopa: Its return type is void, a normal HandleScope will suffice.

src/binding.cpp
kkoopa: Not strings, apart from the empty string. https://github.com/nodejs/nan/blob/master/doc/methods.md#api_nan_return_value

src/sass_types/factory.cpp
saper: > > -  void Factory::initExports(Handle<Object> exports) {
> > -    Local<Object> types = NanNew<Object>();
> > -    exports->Set(NanNew("types"), types);
> > +  void Factory::initExports(v8::Local<v8::Object> exports) {
> > +    Nan::EscapableHandleScope scope;
> 
> Its return type is void, a normal HandleScope will suffice.

Re-read "V8 Embedders Guide" which says "you cannot return a local handle
directly from a function that declares a handle scope". 

Thank you very much!

kkoopa: Should I re-read it, or did you re-read it? Either way, the function does not return anything, so the ability to return handles from the scope is useless.

kkoopa: oh, and this could also use `NAN_MODULE_INIT` if it's being fed from `NODE_MODULE`.

saper: > > -  void Factory::initExports(Handle<Object> exports) {
> > -    Local<Object> types = NanNew<Object>();
> > -    exports->Set(NanNew("types"), types);
> > +  void Factory::initExports(v8::Local<v8::Object> exports) {
> > +    Nan::EscapableHandleScope scope;
> 
> Should I re-read it, or did you re-read it? Either way, the function
> does not return anything, so the ability to return handles from the
> scope is useless.

Of course you are correct! It was my misunderstanding.

saper: On Sun, 9 Aug 2015, Benjamin Byholm wrote:

> > -  void Factory::initExports(Handle<Object> exports) {
> > -    Local<Object> types = NanNew<Object>();
> > -    exports->Set(NanNew("types"), types);
> > +  void Factory::initExports(v8::Local<v8::Object> exports) {
> > +    Nan::EscapableHandleScope scope;
> 
> oh, and this could also use `NAN_MODULE_INIT` if it's being fed from `NODE_MODULE`.

Yes it is! Will rename the parameter.

src/binding.cpp
```diff
@@ -61,52 +61,52 @@ void ExtractOptions(Local<Object> options, void* cptr, sass_context_wrapper* ctx
...
   sass_option_set_indent(sass_options, ctx_w->indent);
   sass_option_set_linefeed(sass_options, ctx_w->linefeed);
 
-  Local<Value> importer_callback = options->Get(NanNew("importer"));
+  v8::Local<v8::Value> importer_callback = Nan::Get(options, Nan::New("importer").ToLocalChecked()).ToLocalChecked();
 
   if (importer_callback->IsFunction()) {
-    Local<Function> importer = Local<Function>::Cast(importer_callback);
-    auto bridge = std::make_shared<CustomImporterBridge>(new NanCallback(importer), ctx_w->is_sync);
+    v8::Local<v8::Function> importer = v8::Local<v8::Function>::Cast(importer_callback);
```
kkoopa: Since you have already determined that `importer_callback` is a `v8::Function`, it is safe to use `v8::Local<v8::Function> importer = importer_callback.As<v8::Function>();`, it skips some type checks and therefore runs infinitesimally faster. The same goes for the other casts.

src/sass_types/factory.cpp
saper: Scratch that, it is called from `RegisterModule` in `src/binding.cpp`. But given it is passed `node::ADDON_REGISTER_FUNCTION_ARGS_TYPE` parameter from Nan, I can (and probably should) use that macro.

src/sass_types/boolean.cpp
```diff
@@ -15,59 +13,63 @@ namespace SassTypes
...
+      Nan::SetPrototypeTemplate(tpl, "getValue", Nan::GetFunction(Nan::New<v8::FunctionTemplate>(GetValue)).ToLocalChecked());
 
-      NanAssignPersistent(constructor, tpl->GetFunction());
+      constructor.Reset(Nan::GetFunction(tpl).ToLocalChecked());
+      conslocal = Nan::New(constructor);
 
-      NanAssignPersistent(get_singleton(false).js_object, NanNew(constructor)->NewInstance());
-      NanSetInternalFieldPointer(NanNew(get_singleton(false).js_object), 0, &get_singleton(false));
-      NanNew(constructor)->Set(NanNew("FALSE"), NanNew(get_singleton(false).js_object));
+      get_singleton(false).js_object.Reset(conslocal->NewInstance());
```
kkoopa: It might be that this should be `get_singleton(false).js_object.Reset(Nan::NewInstance(conslocal));`.

src/create_string.cpp
```diff
@@ -3,12 +3,18 @@
...
   char *str = (char *)malloc(string.length() + 1);
```
kkoopa: And it would be better if you made it return a pointer to a heap allocated `Nan::Utf8String`, which can be dereferenced for access to the underlying `char *` and allows you to do `delete` when appropriate. This way you avoid the second memory allocation and string copy.

src/create_string.cpp
```diff
@@ -3,12 +3,18 @@
...
-  if (value->IsNull() || !value->IsString()) {
+char* create_string(Nan::MaybeLocal<v8::Value> maybevalue) {
+  v8::Local<v8::Value> value;
+  
+  if (maybevalue.ToLocal(&value)) {
+    if (value->IsNull() || !value->IsString()) {
+      return 0;
+    }
+  } else {
     return 0;
```
kkoopa: Could use the `Maybe` API instead of dangerous null pointers. This would be `return Nan::Nothing();`, the other case `return Nan::Just(str);`

src/sass_types/null.cpp
```diff
@@ -15,45 +13,49 @@ namespace SassTypes
...
-      Local<FunctionTemplate> tpl = NanNew<FunctionTemplate>(New);
+      v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(New);
 
-      tpl->SetClassName(NanNew("SassNull"));
+      tpl->SetClassName(Nan::New("SassNull").ToLocalChecked());
       tpl->InstanceTemplate()->SetInternalFieldCount(1);
 
-      NanAssignPersistent(constructor, tpl->GetFunction());
+      constructor.Reset(Nan::GetFunction(tpl).ToLocalChecked());
+      conslocal = Nan::New(constructor);
```
kkoopa: Again,
```c++
conslocal = Nan::GetFunction(tpl).ToLocalChecked();
constructor.Reset(conslocal);
```
is more efficient.

src/sass_types/null.cpp
```diff
@@ -15,45 +13,49 @@ namespace SassTypes
...
 
-      NanAssignPersistent(get_singleton().js_object, NanNew(constructor)->NewInstance());
-      NanSetInternalFieldPointer(NanNew(get_singleton().js_object), 0, &get_singleton());
-      NanNew(constructor)->Set(NanNew("NULL"), NanNew(get_singleton().js_object));
+      get_singleton().js_object.Reset(conslocal->NewInstance());
+      Nan::SetInternalFieldPointer(Nan::New(get_singleton().js_object), 0, &get_singleton());
```
kkoopa: What's this doing? Is it taking the address of an rvalue?

src/sass_types/null.cpp
```diff
@@ -15,45 +13,49 @@ namespace SassTypes
...
   }
 
   Sass_Value* Null::get_sass_value() {
     return sass_make_null();
   }
 
-  Local<Object> Null::get_js_object() {
-    return NanNew(this->js_object);
+  v8::Local<v8::Object> Null::get_js_object() {
+    return Nan::New(this->js_object);
```
kkoopa: Should be inlined

src/sass_types/factory.cpp
kkoopa: Yes. Alternatively, you can use `Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE` directly, if you feel so inclined.

src/create_string.cpp
saper: > > -char* create_string(Local<Value> value) {
> > -  if (value->IsNull() || !value->IsString()) {
> > +char* create_string(Nan::MaybeLocal<v8::Value> maybevalue) {
> > +  v8::Local<v8::Value> value;
> > +  
> > +  if (maybevalue.ToLocal(&value)) {
> > +    if (value->IsNull() || !value->IsString()) {
> > +      return 0;
> > +    }
> > +  } else {
> >      return 0;
> 
> Could use the `Maybe` API instead of dangerous null pointers. This
> would be `return Nan::Nothing();`, the other case `return
> Nan::Just(str);`

Unfortunately we are leaving a nice V8 world here:

create_string() pointers are being directly fed into C structures.
In fact this function exists only because the libsass C API insists
that almost all (char *) pointers should be heap allocated, because it
says it frees them (I think this API is incorrect, but that's another story).

But ~Utf8String() may crash if someone free()'s that underlying string, am I correct?
And if the string fits n str_st_ it can't be free()d.

src/sass_types/null.cpp
saper: On Sun, 9 Aug 2015, Benjamin Byholm wrote:

> > +      Nan::SetInternalFieldPointer(Nan::New(get_singleton().js_object), 0, &get_singleton());
> 
> What's this doing? Is it taking the address of an rvalue?

get_singleton() returns Null& so that forms an "lvalue expression"
that can be &

In boolean.cpp we do

&get_singleton(false) and &get_singleton(true)

src/create_string.cpp
kkoopa: That is correct.

src/sass_types/null.cpp
saper: I can't inline this since this is (currently) a [public interface mandated by the superclass](https://github.com/sass/node-sass/blob/34db870992bcae17de30dd0e1f0fd8ddb4bcc363/src/sass_types/value.h#L15):

    virtual v8::Local<v8::Object> SassTypes::Value::get_js_object()

kkoopa: Should use an escapable scope then.

saper: I am confused :)

Let me follow the call chain:

[`CallbackBridge<>::operator()(std::vector<void*> argv)`](https://github.com/saper/node-sass/blob/a1395134907c6167a71487f5937869f7ca7d21b2/src/callback_bridge.h#L80-L91) calls `pre_process_args()` in two ways:

1) directly in the sync case - we run in the V8 thread, so we create `EscapableHandleScope` and call `pre_process_args()`.  (Note: Now I think `HandleScope` should be enough here).
2) indirectly in the async case - we schedule [`CallbackBridge<>::ReturnCallback`](https://github.com/saper/node-sass/blob/a1395134907c6167a71487f5937869f7ca7d21b2/src/callback_bridge.h#L120-L138) which is a `NAN_METHOD` and has implicit `HandleScope`.

One of those methods, [`CustomFunctionBridge::pre_process_args()`](https://github.com/saper/node-sass/blob/a1395134907c6167a71487f5937869f7ca7d21b2/src/custom_function_bridge.cpp#L14-L22), calls `get_js_object()` so I assume it is all in the scope opened as described above?

`get_js_object()` is also called from some types `New()` methods (all defined via `NAN_METHOD()`), where it is wrapped in their implicit `HandleScope`.

If the above is correct, I might not need `EscapableScope` in the `get_js_object()`, right? Can I use calling function `(Escapable)HandleScope` instead?

kkoopa: You need an `EscapableHandleScope` in 1), the answer is 'yes' to everything else.

src/binding.cpp
saper: @kkoopa regarding the proper application of `Maybe<>` types:

Can you have a look at https://github.com/saper/node-sass/commit/df65d0aece948f90972d2286717e88b002f41579 - I have restructured the `Color` class construction to use `Maybe<>` types and not use `IsNumber()` or something like that.

Looks like a pretty extreme functional style - is this a proper way to go?

kkoopa: It looks alright, as far as I can see.

src/sass_types/null.cpp
saper: I'm lost! Do you mean I need `EscapableHandleScope` here:

````cpp
template <typename T, typename L>
T CallbackBridge<T, L>::operator()(std::vector<void*> argv) {
  // argv.push_back(wrapper);
  if (this->is_sync) {
    Nan::EscapableHandleScope scope;
    std::vector<v8::Local<v8::Value>> argv_v8 = pre_process_args(argv);
    argv_v8.push_back(Nan::New(wrapper));

    return this->post_process_return_value(
      scope.Escape(this->callback->Call(argv_v8.size(), &argv_v8[0]))
    );
  }
  /* ... */
````

I am not returning `Local<>` handle in this method...?

kkoopa: You need it there because you create the handle scope in the scope of the if-block. It does not exist outside.

kkoopa: Although, yes, since you don't return it, a normal `HandleScope` ought to work.

saper: Sorry to be a PITA, I'm learning this way...

In the [current code](https://github.com/saper/node-sass/blob/a1395134907c6167a71487f5937869f7ca7d21b2/src/callback_bridge.h#L80-L91) we have only few `NewNan`s, I guess it works only because we are lucky:

````cpp
if (this->is_sync) {
    std::vector<Handle<Value>> argv_v8 = pre_process_args(argv);
    argv_v8.push_back(NanNew(wrapper));

    return this->post_process_return_value(
      NanNew<Value>(this->callback->Call(argv_v8.size(), &argv_v8[0]))
    );
  }
````

Thank you again!

src/binding.cpp
saper: Just learned the hard way that `Nan::To` tries really hard to convert the values and is no replacement for `->IsNumber()` check and friends. Some string treated with converter to `<v8::Boolean>` or `<bool>` will be simply false.

xzyfer:
@saper thanks again for your work here. Please let me know when you think this is ready to ship (after squashing the git history of cause ;) )

@kkoopa we really appreciated your help here!

xzyfer:
@saper do you think you'll have to complete the revisions soon or should I fork this branch?

saper:
> @saper do you think you'll have to complete the revisions soon or should I fork this branch?

Forking is always good :)

The good news is that I believe that currently the branch is correct as far as functionality goes,
we are just tidying up the nan-related things. And maybe
(should I say Nan::Maybe<>? :) making things nicer. That's why I dedicded to have separate
commits on top of the original patch - they are self-contained and all tests pass in between.
Should we accidentally break something (missing Nan::New somewhere or similar), it would be
easy to fix.

We definitely need more tests for SassTypes! Like https://github.com/sass/node-sass/pull/1058
and also I am uncomfortable with the fact that CLI testing discovers bugs in the JS/C++ bridge
https://github.com/sass/node-sass/issues/1053#issuecomment-128311740
We should have tests before @chriseppstein decided to use our lists or maps for anything
usefel! :)

Right now I am working on ->ToXXX() stuff, but probably I'll push some other low-hanging
fruit first before. I think we can move from ->ToXXX to typecasts in some cases.

So right now I'd like to do it like this:

1) push more easy nitpicks
2) push changes in scope handling where still needed
3) push ->ToXXX/typecase work

xzyfer:
The first priority is to get into a state where we can ship a workable iojs 3.0.0 binary :) 
Last I checked this worked with OSX. When you're comfortable with the state of iojs 3.0.0 support I'm keen to build some binaries, test it, and ship.

At which point there is no rush to get the v8 binding code cleaned up, and no versioning issues since this completely internal.

With the latest LibSass beta about to drop I'll need to update LibSass binding code as well.

****

I wasn't sure if https://github.com/sass/node-sass/pull/1058 was dependant on this PR. That's the only reason I haven't shipped it.

xzyfer:
As for your final points regarding test. I completely agree. Custom functions, and by extension the Sass types are experimental features developed by a non-core developer. This is largely accountable for the lack of test general lack of attention.

saper:
On Tue, 11 Aug 2015, Michael Mifsud wrote:

> With the latest LibSass beta about to drop I'll need to update LibSass binding code as well.

Like this? https://github.com/sass/node-sass/pull/1040

I think we should definitely do a release with io.js 3.0 but with libsass 3.2.5.

Otherwise it'll be difficult to quickly sort out which one broke what in the reports...

xzyfer:
>I think we should definitely do a release with io.js 3.0 but with libsass 3.2.5.

Totally agree. That'll be 3.1.0 when this is ready to go.

xzyfer:
>Like this? https://github.com/sass/node-sass/pull/1040

Nice I missed this.

saper:
> >Like this? https://github.com/sass/node-sass/pull/1040
> 
> Nice I missed this.

While here; do you think we should add io.js 2.x to the build
matrix?

xzyfer:
>While here; do you think we should add io.js 2.x to the build matrix?

Assuming you mean travis ci yes. Also iojs 1.x, and 1.0.x.. essentially anything we're currently building binaries for.

xzyfer:
>While here;

For the record you can get me our Slack pretty much anytime. If I'm not there I'll get back to when I am. I'm online there most of the day.

xzyfer:
@saper code cosmetics aside is there any reason we can't ship this as is?

src/create_string.cpp
saper: I'll revisit it later - we might need to review `create_string()` anyway to better comply with libsass API. 

For now I have more issues to worry than V8 sending me wrong UTF-8 :)

saper:
So, this is it, I think.

I have addressed almost all comments from @kkoopa (big thanks again!). In case I decided to go in the other direction - I have replied to the code review comment in this thread.

Every commit in this pull request is standalone and should be passing tests (bar whimsical AppVeyor behaviour), so I will not be squashing them. If we will find problems, I want to have one of those commits clearly identified.

I hope I will soon stop whispering `<v8::Local<v8::Value>>` in my dreams :)

xzyfer:
:heart_eyes:

saper:
Looks like AppVeyor has some trouble installing node v0.10.40.

dukex:
Thank you @saper, @xzyfer and @kkoopa <3 <3

PR #1058, opened 2015-08-10
Additional tests for SassTypes::Color

saper:
A byproduct of #1054

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -769,6 +769,50 @@ describe('api', function() {
       });
     });
 
+    describe('should properly bubble up errors from sass color constructor', function() {
+      it('four booleans', function(done) {
+        sass.render({
+          data: 'div { color: foo(); }',
+          functions: {
+            'foo()': function() {
+              return new sass.types.Color(false, false, false, false);
+            }
+          }
+        }, function(error) {
+          assert.ok(/Constructor arguments should be numbers exclusively/.test(error.message));
+          done();
+        });
+      });
+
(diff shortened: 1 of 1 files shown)


xzyfer:
Ship when ready! :tada:

saper:
> Ship when ready! :tada:

Not sure I like the indented layout of output. Opinions?

We need moar of these! List! Maps! Crazy constructors!

If I do conversion in the binding code, 
"new sass.types.Boolean('string')" brings false,
if I do casting - it errors. I think the latter is better?

xzyfer:
>Not sure I like the indented layout of output. Opinions?

I don't have my dev laptop with me atm. 

However generally speaking I care about the layout of the test themselves with regards to readability (which I think is perfectly fine). IMHO the output doesn't matter since you only care about the tests that fail :)

xzyfer:
>If I do conversion in the binding code,
>"new sass.types.Boolean('string')" brings false,
>if I do casting - it errors. I think the latter is better?

I'm not following. Does this relate directly to this PR. I'd rather we didn't change the implementation of the Sass types atm.

saper:
> >If I do conversion in the binding code,
> >"new sass.types.Boolean('string')" brings false,
> >if I do casting - it errors. I think the latter is better?
> 
> I'm not following. Does this relate directly to this PR. I'd rather we didn't change the implementation of the Sass types atm.

Good, I'll keep that in mind! Once I figure out how it works :>

Issue #1061, opened 2015-08-11
npm@3: Error: Cannot find module '/tmp/node_modules/node-sass/node_modules/pangyp/bin/node-gyp


saper:
npm 3 has a new toy called auto-flattening, i.e all dependencies will be installed in the same level `node_modules` if possible. This breaks our assumption about how to find `node-gyp`.

Workaround: make sure you are using npm 2.x series.

yairEO:
I apparently got stuck with this error after upgrading my Node to the latest (on windows) `v0.12.7`
`NPM -v` is `2.11.3` ... 

I hope this will be solved quickly..

saper:
This error is about npm 3.0 and we have a fix in #1063. Can you go through our [TROUBLESHOOTING.md](https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md) guide and post the output to the new issue?

busches:
Using this with `npm@3.3` resolved the `Error: Cannot find module 'X:\node_modules\node-sass\node_modules\pangyp\bin\node-gyp'.` I was getting.

PR #1063, opened 2015-08-11
Locate pangyp via module API

saper:
Might fix https://github.com/sass/node-sass/issues/1061

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -128,7 +128,7 @@ function build(options) {
       process.exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


xzyfer:
Seems legit. I'll confirm it doesn't break npm@2 and ship

saper:
It should also fix cases when pangyp is reachable globally with npm@2

xzyfer:
According to [the docs](https://nodejs.org/api/globals.html#globals_require_resolve) this should work as expected :+1:

PR #1081, opened 2015-08-17
Added note to indentedSyntax in read.me 

dkebler:
Added note to indentedSyntax in read.me to clarify that node-sass can compile mixed libraries even with this set to false.


saper:
Thank you very much :tada: !

Issue #1090, opened 2015-08-20
sass.types.Color(2,3) causes "Empty MaybeLocal" abort


saper:
After #1058 (which is testing this) when we try to execute this:

````js
var sass = require('./');
        sass.render({
        data: 'div { color: foo(); }',
          functions: {
            'foo()': function() {
              return sass.types.Color(2,3);
            }
          }
});
````

We get an abort: `FATAL ERROR: v8::ToLocalChecked Empty MaybeLocal.` because `Color::construct()` throws a C++ exception.

PR #1092, opened 2015-08-20
Handle MaybeLocal value from the constructor

saper:
In case of exception occuring during construction
of a Sass type instance just return Undefined value,
as the delayed JS exception will pop up later.

The following code:

    var sass = require('./');
        sass.render({
        data: 'div { color: foo(); }',
          functions: {
            'foo()': function() {
              return sass.types.Color(2,3);
            }
          }
    });

now results in an libsass error

     Constructor should be invoked with either 0, 1, 3 or 4 arguments

instead of V8 abort

     FATAL ERROR: v8::ToLocalChecked Empty MaybeLocal

Fixes https://github.com/sass/node-sass/issues/1090

### DIFF
--- a/src/sass_types/sass_value_wrapper.h
+++ b/src/sass_types/sass_value_wrapper.h
@@ -106,8 +106,12 @@ namespace SassTypes
       }
     } else {
       v8::Local<v8::Function> cons = T::get_constructor();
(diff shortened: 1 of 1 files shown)


saper:
Let's keep tests clean.

Issue #56, opened 2013-02-03
Automate building of binaries


andrew:
We need to make it easier to rebuild binaries for each platform, perhaps using vagrant boxes and puppet.

saper:
Both Travis CI and Appveyor can push artifacts to a GitHub release: http://www.appveyor.com/docs/deployment/github I think this is the simplest way to do it out of CI, once we decide robots are in charge :)

xzyfer:
I have looked into the Travis CI approach in the past. It'll work for oxs
but IIRC it's not compatible with hoops we jump for old centos support in
the Linux builds. Also Travis is using docker so we can't use our docker
script because they don't allow docker in docker for security reasons.

saper:
I am thinking about simple

````
deploy:
  provider: releases
  api_key: $GITHUB
  file: XXXX
````

setup that just works, no docker script required. Just upload the generated binding file.

Funny Linux setup won't work (unless you get the whole process working for Travis - also possible).
FreeBSD won't work as well - but those can still be build manually and added to the release.

Regarding old CentOS support - I am ready to re-visit https://github.com/sass/libsass/pull/995 and https://github.com/sass/libsass/pull/996 to make sure libsass standard autoconf setup works for old CentOS. Need only somebody to test it for me. Then no patching should be necessary - and we could use stock `libsass`. The only problem is that we might need `autoconf` setup but I see two solutions there:

1) Make `node-sass` setup using externally-built libsass with `autoconf`
2) Define a set of `config.h` constants in the `libsass.gyp` file once old CentOS is detected.

What is the distinguishing feature of the "old CentOS"? Old compiler? Which version it is? gcc 4.7 ?

xzyfer:
> Then no patching should be necessary - and we could use stock libsass

We don't patch LibSass anymore. @am11 put together a nice Docker setup that builds against an old gcc which seems to have solved our problem - https://gist.github.com/xzyfer/f5623546cb73b601b9b8

> FreeBSD won't work as well

@am11 mentioned this in chat, does this help with automating FreeBSD binaries?
>also to run the docker in travis, found: https://github.com/rocker-org/rocker/blob/56cb59cc262e9d2c02f9e11cfacab65f0b659c5f/.travis.yml#L10-L24 via http://stackoverflow.com/a/28223092.

saper:
So it is only a matter of having CentOS 5 `libc` and `libstc++.so` from their g++ (I think it is 4.8.2 now) around, good.

[...]

PR #1097, opened 2015-08-21
CI: Adds explicit runtime versions in job matrix

am11:
* CI: Adds explicit runtime versions in job matrix.
  Updates appveyor.yml.

  Related: #1096
* CI: Stores build artifact (binding.node) for job. 
  * The artifact will be available under job details in `Artifacts` tab
  * URL format: `https://ci.appveyor.com/project/sass/node-sass/build/<build-id>/job/<job-id>/artifacts`
* Package: Updates to latest versions.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,13 +7,11 @@ environment:
   matrix:
     # node.js
     - nodejs_version: 0.10
-    - nodejs_version: 0    # will fetch the latest node.js version
+    - nodejs_version: 0.12
     # io.js
-    - nodejs_version: 1    # will fetch the latest io.js version
-
-matrix:
-  allow_failures:
     - nodejs_version: 1
+    - nodejs_version: 2
+    - nodejs_version: 3
 
 install:
   - ps: Install-Product node $env:nodejs_version
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,3 +21,12 @@ install:
   - npm install --msvs_version=2013
 
 test_script: npm test
+
+on_success:
+# Save artifact with full qualified names of binding.node
+# (which we use in node-sass-binaries repo)
+- ps: Get-ChildItem .\vendor\**\*.node | % `
+      {
+        Push-AppveyorArtifact $_.FullName -FileName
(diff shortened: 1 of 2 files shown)


am11:
/cc @xzyfer, @saper, @nschonni

This is related to #56.

The artifacts are publically accessible. To wit: https://ci.appveyor.com/project/sass/node-sass/build/1180/job/1sjh8wsp5rfb1n95/artifacts, which means no more Windows machine required for releases.

Random thoughts:

We can configure the TravisCI do the same for Mac and Linux. As I have understood it, TravisCI doesn't provide out of the box storage but encourages users to configure an external storage like S3 bucket (which makes sense IMO, as this approach promises storage permanency).

After that is done, we can actually make our install script serve the binary from the CI server directly. For instance:
https://ci.appveyor.com/api/buildjobs/1sjh8wsp5rfb1n95/artifacts/win32-ia32-14_binding.node. If we come across issue of "artifact expiry" (which I am not sure whether or not release artifacts get deleted from AppVeyor cloud after some time), we can push the artifact to the same instance of S3 bucket which will contain FreeBSD, Mac and Linux binaries (this can be script within `appveyor.yml`).

To consume the binaries from CI artifacts directly, we can probably define a new item `"binaries": { }` in `package.json`, which will contain the URL (or the unique part of the URL) for each combination. For instance:

```json
"binaries": {
  "linux32-ia32-44": "<s3-url>",
  "win32-ia32-14": "1sjh8wsp5rfb1n95"
}
```

If this is the direction we want to take (as opposed to configuring node-pre-gyp strategy; which IMO possesses some technical debt as their advertised approach of building binaries etc. is slightly incompatible with our architecture), I can close #56 and open a new issue with the proposed strategy; in order to make CI do all the heavy-lifting for us in future.

saper:
Replied in #56

saper:
What do you mean by saying node-pre-gyp advertised strategy is not compatible with us?

am11:
That we don't need to make changes to configure node-pre-gyp in our setup, as CIs are flexible enough.

Issue #1077, opened 2015-08-17
"Build succeeded but target not found": Debug/Release should be autodetected


saper:
I am currently running `node-sass` with io.js 3.0 in the debug mode, therefore `node-gyp` prepares the build in the `build/Debug` directory.

I need to use the following workaround to avoid the `Build succeeded but target not found` problem at the end:

````
diff --git a/scripts/build.js b/scripts/build.js
index 4e726e1..990ba85 100644
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -18,9 +18,9 @@ require('../lib/extensions');
  * @api private
  */
 
-function afterBuild(options) {
+function afterBuild() {
   var install = process.sass.binaryPath;
-  var target = path.join(__dirname, '..', 'build', options.debug ? 'Debug' : 'Release', 'binding.node');
+  var target = path.join(__dirname, '..', 'build', process.config.target_defaults.default_configuration, 'binding.node');
 
   mkdir(path.dirname(install), function(err) {
     if (err && err.code !== 'EEXIST') {
@@ -141,7 +141,7 @@ function build(options) {
 
     proc.on('exit', function(errorCode) {
       if (!errorCode) {
-        afterBuild(options);
+        afterBuild();
 
         return;
       }
````

This is not a 100% correct solution since we might be using `--node-dir` or be otherwise cross-compiling node-sass to some other node engine other than the one currently using. We should ideally get this info back from `node-gyp` somehow.

Once this is in place we can remove the `-d` option.

PR #1101, opened 2015-08-24
Use process.config.target_defaults.default_configuration as the build type default

saper:
`node-gyp` is using `process.config.target_defaults.default_configuration`
to check which configuration is about to be used, but this
can be overriden by particular engine's `config.gypi` file.
    
Since we do not support such a cross compilation anyway,
use this variable in place of a hardcoded default.
    
Instead of introducing another option to force the use
of 'Release' we should read and pass options to node-gyp.
This is not implemented here.

Fixes https://github.com/sass/node-sass/issues/1077

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -20,7 +20,9 @@ require('../lib/extensions');
 
 function afterBuild(options) {
   var install = process.sass.binaryPath;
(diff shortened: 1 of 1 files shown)


xzyfer:
:+1:

PR #1108, opened 2015-08-26
Workaround the segfault that occurs when used with fsevents

xzyfer:
`fsevents@0.3.8` introduces a segfault that manifests when used with node-sass https://github.com/strongloop/fsevents/issues/82

`fsevents` is widely popular. As a mitigation we've added `fsevents@0.3.7` as an `optionalDependency`. The intention here is the prevent users install `node-sass@3.3.x` and `fsevents@0.3.8` by using npm's deduping behaviour.

Adding `optionalDependency` will cause npm to dedupe the package and install the most suitable version. Since we've pinned to a specific version ours will always be the most suitable.

This is certainly cause us problems in the long run and should be reverted as soon as issue fixed upstream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,9 @@
     "request": "^2.61.0",
     "sass-graph": "^2.0.1"
   },
+  "optionalDependencies": {
+    "fsevents": "0.3.7"
+  },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


saper:
We should no longer need this with #1110 and #1113.

Issue #1107, opened 2015-08-26
3.3.0 segmentation fault 11 on OSX


bestander:
Node quits with "segmentation fault 11" when node-sass is required.
I am using gulp-sass that depends on node-sass.

Mac OS X 10.10.5 
node -v
v2.3.0
npm -v
2.12.1

bestander:
Reverting to 3.2.0 fixes the bug

eddiemoore:
:+1: I'm having the exact same issue
```shell
node -v
# v2.5.0
npm -v
# 2.13.2
```

I'm using it through gulp-sass though, so can't really revert.

xzyfer:
My guess is that you need to do a delete your `node_modules` and do a clean install. Is sounds like your system has kept the v3.2.0 rather than downloading the updated v3.3.0 binary.

bestander:
Nope, 3.3.0 is definitely causing the problem on my PC. 
Reverting 3.3.0 with 3.2.0 manually fixes the problem.

xzyfer:
@bestander please follow our [troubleshooting guide](https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md#linuxosx) and create a gist with all the output.

We're unable to reproduce this issue.

yizhan:
I deleted my node_modules, and did a clean install. It shows segmentation fault 11 now.

xzyfer:
If someone could produce a scss sample that produce that the segfault that would greatly help.

bestander:
Sure will do.  
I am using gulp-sass, so maybe the reason is that latest gulp-sass is not compatible with node-sass 3.3.0

enriquecaballero:
Same here. Downgrading to 3.2.0 fixes the Segmentation Fault: 11.

bestander:
Hm, in an isolated case it does not throw the error:
```
var sass = require('gulp-sass');
var gulp = require('gulp');

gulp.task('test', function () {

});
```

But in our production system with a large gulp script just having 
```
var sass = require('gulp-sass');
```
Throws node with error.

```
npm -v
2.12.1
node -v
v2.3.0
node -p process.versions
{ http_parser: '2.5.0',
  node: '2.3.0',
  v8: '4.2.77.20',
  uv: '1.6.1',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  modules: '44',
  openssl: '1.0.2c' }
node -p process.platform
darwin
node -p process.arch
x64
```

[...]

PR #1109, opened 2015-08-26
Fix iojs compatibility by using fsevents 0.3.8

gravityrail:
This makes things work with iojs

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "sass-graph": "^2.0.1"
   },
   "optionalDependencies": {
-    "fsevents": "0.3.7"
+    "fsevents": "0.3.8"
   },
   "devDependencies": {
     "coveralls": "^2.11.4",


saper:
I am afraid this will not work - see https://github.com/sass/node-sass/issues/1107 and https://github.com/strongloop/fsevents/issues/82. Latest node-sass and fsevents do not seem to get along together.

xzyfer:
Thanks @gravityrail but as per https://github.com/sass/node-sass/issues/1107#issuecomment-134867296 `fsevents@0.3.8` introduces a segfault which breaks Node Sass. Locking to `fsevents@0.3.7` as an attempt to prevent npm install the broken the version https://github.com/sass/node-sass/pull/1108.

Yes. Upgrading to `fsevents@0.3.8` will give you iojs v3 compatibility. However it'll mean you cannot use Node Sass.

**TLDR** don't use iojs v3 for anything that matters yet.

PR #1110, opened 2015-08-27
Use libstdc++.so.6.0.9 to workaround static init problems

saper:
libsass needs to be linked against libc++ from clang
on MacOSX to use advanced C++11 functions.

The binding code is not that picky and we can try
to link against whatever is the default.
Mixing C++ runtime libraries is still very bad, but
we should be able to avoid crash when being
used with another node module using nan.h
and its GetWrapper() static cache:

https://github.com/strongloop/fsevents/issues/82

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -57,8 +57,7 @@
         ['OS=="mac"', {
           'xcode_settings': {
             'OTHER_CPLUSPLUSFLAGS': [
-              '-std=c++11',
-              '-stdlib=libc++'
+              '-std=c++11'
             ],
             'OTHER_LDFLAGS': [],
             'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
--- a/binding.gyp
+++ b/binding.gyp
@@ -62,7 +62,6 @@
             ],
             'OTHER_LDFLAGS': [],
             'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
-            'GCC_ENABLE_CPP_RTTI': 'YES',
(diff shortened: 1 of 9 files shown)


xzyfer:
IIRC this would ideally be superseded by https://github.com/nodejs/nan/pull/431?

saper:
@xzyfer can you check if this compiles, and if it does, does it crash with stock nan 2.0.7 (without https://github.com/nodejs/nan/pull/431 applied)

xzyfer:
This appears to work @saper

saper:
Yay! :happy:

xzyfer:
Do you have objections to the nan patch? Doesn't look like we can avoid it

saper:
As to relation to https://github.com/nodejs/nan/pull/431: nan use of static revealed an incompatibility between C++ libraries or a linker/compiler bug. This can be to a large extent avoided by this patch here. There is a possibility that some other binary component (not necessarily nan) introduces a feature that breaks again, so we should stay on a safe side.

xzyfer:
I'm happy for you to ship this. I'm for about 16hrs. Feel free to merge and
start building freebsd binaries.

saper:
So we are about to get 3.3.2 soon :)

saper:
@xzyfer would be cool if you were able to check 60e0b25566bc88bd406a803efbdfe971e3dc5a86 (without nan patch) if it fixes the issue. We should probably get MacOSX CI :)

xzyfer:
We've tried to get OSX CI but we were denied by the Travis team. I'll chase them up again.

Issue #1048, opened 2015-08-02
node process hangs on unrelated error when using custom importers


mattfysh:
The error when thrown doesn't cause the node process to exit
SSCCE:
```javascript
var opts = {data: '@import "x"', importer: function() {}};
require('node-sass').render(opts, function() {});
throw new Error();
```
We were able to work around this using renderSync but our builds are a little slower. Any ideas what could be causing this, or if there's a better workaround available? thanks.

mattfysh:
Found a way to continue using render (and keep build performance) using an uncaughtException listener:
```javascript
var opts = {data: '@import "x"', importer: function() {}};
require('node-sass').render(opts, function() {});
process.on('uncaughtException', function(e) {
	console.error(e.stack);
	process.kill(process.pid, 'SIGINT');
});
throw new Error('test msg');
```
Not the cleanest solution... are there any plans to fix this behavior for node-sass? I noticed #857 about the process hanging with the UV_THREADPOOL_SIZE the only solution given for now.. thanks

xzyfer:
Thanks for this @mattfysh. This is not related to #857. We should catch this error and return a useful error message.

saper:
Yes, we schedule an event and wait on the lock the event handler returns.
The problems is it never returns, because it never gets called - the event loop gets stopped before it happens.

saper:
So, during the next rewrite of the binding code this will be taken into account. For now there is no easy solution (except for `process.kill`). Sorry for that!

PR #1112, opened 2015-08-29
[WIP] Timeout custom JS after 2000ms

saper:
Do not wait for the JavaScript code
indefinitely - abort execution
attempt after 2000ms.

JavaScript code might never get called
if no loop events will be processed
(https://github.com/sass/node-sass/issues/1048)
or might block forever if there are
no more worker threads available
(https://github.com/sass/node-sass/issues/857).


wmertens:
Nice! I think 2s might be too short on slow systems though… how about 20s? It's not like this will be encountered often…

saper:
On Sun, 14 Feb 2016, Wout Mertens wrote:

> Nice! I think 2s might be too short on slow systems though… how about 20s? It's not like this will be encountered often…

Feel free to change it as you wish. It's an experimental patch and it's kind of wrong, but it might
help to diagnose the problem (I doubt it will make compilation possible - it might only help to pinpoint
the offending include).

xzyfer:
@saper what's the status of this?

saper:
This is just a workaround for people experiencing #1048 or #857, it still merges with the master. I'd leave it open here.

PR #1114, opened 2015-08-29
Revert "Workaround the segfault that occurs when used with fsevents"

saper:
This reverts commit 3166b613fd17f1182e99a9e46fd9e355d3dfb8e1.

Not needed any more after update to nan 2.0.8 and
enabling the use of libstdc++.so.6 for the binding.


saper:
(Reverts #1108)

PR #1117, opened 2015-08-31
Update to the new Travis CI infrastructure.

xzyfer:
http://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: node_js
 
-env: SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
+env: SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true CC=gcc-4.7 CXX=g++-4.7 LINK=gcc-4.7 LINKXX=g++-4.7
 
 compiler: gcc
+
+sudo: false
 
 node_js:
   - "0.10"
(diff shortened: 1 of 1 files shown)


saper:
Er, looks like we are not getting the compiler we want. My suggestion for priority: low :)

saper:
According to [C++0x/C++11 Support in GCC table] [template alias](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2258.pdf) should be supported in g++ 4.7. Which compiler are we getting in the end?

Interestingly, [libsass seems to compile just fine with the default gcc 4.6.3](https://travis-ci.org/sass/libsass/jobs/78113138). So the new V8 is quite picky about the compiler. Can we ask for g++ 4.7 only for io.js 3 in the new infrastructure? (besides it does not seem to work)

xzyfer:
It looks like I can install gcc/gcc 4.7 but I can't alias gcc/g++ to the newer versions without `sudo`.

I agree this isn't a high priority. I just wanted to take a run at it before going to bed. The slow Travis times are also starting to bug me.

saper:
That should be relatively easy, then. If the compilers are named `gcc47` and `g++47` we just need to call `/usr/bin/env CC=gcc47 CXX=g++48 LINK=gcc47 LINKXX=g++47 npm install`

xzyfer:
Looks like this works now. I've some intermittent Travis failures which don't instill me with confidence.

PR #1128, opened 2015-09-08
Replace pangyp with node-gyp

### LINKED ISSUE #1124: Move back to node-gyp
`node-gyp@3` has been released with iojs support - https://github.com/nodejs/node-gyp/pull/711

Since iojs support was the primary reason for moving to `pangyp` it's time to start looking at moving back. 

From the `pangyp` readme
>A (hopefully temporary) fork of TooTallNate/node-gyp adding io.js support

xzyfer:
We originally adopted pangyp as a tempory work around for iojs support. This and more has been added to node-gyp and as such pangyp has been deprecated.

As far as I can tell this is all that's needed. My guess is we'll need to generated the bindings so lets ship this when we ship the node@4 bindings.

Closes #1124


saper:
This is not strictly necessary for 4.x if on supplies node headers to compile. I have some issues recently with gyp variables, looks like a variant of node js/nodejs#26. Will be tracking this down and posting upstream.

PR #1130, opened 2015-09-08
Make Travis CI to report failures on Node.js 4.0

gdi2290:


.travis.yml
```diff
@@ -13,6 +13,7 @@ node_js:
   - "iojs-v1"
   - "iojs-v2"
   - "iojs-v3"
+  - "4.0"
```
mzgol: It should be `4`, not `4.0`. Node will surely get a lot of minor updates before it goes LTS and node-sass should test on the latest of them.

This is analogous to how in the above lines you can see `iojs-v3`, not `iojs-v3.0`.

gdi2290: good point, one sec I'll update the pr

gdi2290:
@mzgol  I pushed up the change thanks for your feedback

mzgol:
:+1:

xzyfer:
Thanks @gdi2290 and @mzgol. We'll merge this when we ship Node.js 4 support

PR #1132, opened 2015-09-09
Add Node.js 4 to AppVeyor

xzyfer:
This PR adds Node.js 4 support to AppVeyor matching the Travis config (https://github.com/sass/node-sass/pull/1130)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,8 @@ environment:
     - nodejs_version: 1
     - nodejs_version: 2
     - nodejs_version: 3
+    # node
+    - nodejs_version: 4
 
 install:
   - ps: Install-Product node $env:nodejs_version


xzyfer:
We expect this to fail without https://github.com/sass/node-sass/pull/1128

mzgol:
This suggest you're adding io.js 4 because of the comment above.

xzyfer:
@mzgol true. Fixed in 05b7c9c

mzgol:
node -> node.js to be consistent with the first comment?

Sorry for nitpicking. This is not terribly important. :D

xzyfer:
@mzgol I recall there is was meant to a distinction between Node, io.js and Node.js because of trademark issues. Can't remember the specifics. I'll look into it.

mzgol:
I meant that the first comment says "node.js" and the third just "node". The official name is Node.js in both cases.

xzyfer:
@mzgol https://github.com/sass/node-sass/commit/c5a16a9f10b63e88050740cb0ba863b1fe2abc69 :)

Issue #1131, opened 2015-09-08
Custom function returning sass.types.List(1) containg bare JS Object causes a crash


saper:
To reproduce:

````
var sass = require('node-sass');
      sass.render({
        data: 'div { color: foo(); }',
        functions: {
          'foo()': function() {
            var out = new sass.types.List(1);
            out.setValue(0, new Object());
            return new out;
          }
        }
      }, function(error) {
        console.log(error);
        done();
      });
````

Backtrace:

````
#4  0x0000000802dbe761 in SassTypes::Factory::unwrap (obj=...) at ../src/sass_types/factory.cpp:68
#5  0x0000000802dc0ff5 in SassTypes::List::SetValue (info=...) at ../src/sass_types/list.cpp:73
#6  0x0000000802dac93c in Nan::imp::FunctionCallbackWrapper (info=...) at ../node_modules/nan/nan_callbacks_12_inl.h:174
#7  0x00000000007f7f45 in v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&))
    ()
````

Fix on the way!

saper:
Closed via #1133

Issue #1127, opened 2015-09-08
Uncaught std::invalid_argument: Constructor arguments should be numbers exclusively


xzyfer:
Originally mentioned in https://github.com/sass/node-sass/issues/1124#issuecomment-138547391
Error produced running `npm test` with v3.3.2 on iojs 3.2.0 on Darwin.

```
      ✓ should always map null, true and false to the same (immutable) object
      should properly bubble up errors from sass color constructor
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: Constructor arguments should be numbers exclusively.
[1]    40186 abort      npm test
```

```
$ otool -L  /Users/michael/.nvm/versions/io.js/v3.2.0/bin/node
/Users/michael/.nvm/versions/io.js/v3.2.0/bin/node:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1153.18.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
	/usr/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 104.1.0)
	/usr/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 283.0.0)
```

Build log - https://gist.github.com/xzyfer/61d161a8b56ec0f86b8e

saper:
Which commit are you at?

xzyfer:
That log was from the `v3.3.2` tag, but is see same issue on master edbf497

saper:
Can you also `ldd <your-binding-file>` ?

saper:
And ... also I'd like to have your `/Users/michael/Projects/Sass/node-sass/build/config.gypi`

mzgol:
I have exactly the same error on OS X 10.11 (public beta 6) with Node.js 4.0.0 final.

My `./build/config.gypi`:
```
# Do not edit. File was generated by pangyp's "configure" step
{
  "target_defaults": {
    "cflags": [],
    "default_configuration": "Release",
    "defines": [],
    "include_dirs": [],
    "libraries": []
  },
  "variables": {
    "asan": 0,
    "host_arch": "x64",
    "icu_data_file": "icudt55l.dat",
    "icu_data_in": "../../deps/icu/source/data/in/icudt55l.dat",
    "icu_endianness": "l",
    "icu_gyp_path": "tools/icu/icu-generic.gyp",
[...]

PR #1133, opened 2015-09-09
Do not use C++ exceptions in the binding code

saper:
We are running into trouble with C++ runtime
libraries (notably by mixing libstdc++ and libc++
on MacOSX). Reduce the risk by adhering to Google
Style Guide and stop using C++ exceptions at all.

libsass should offer us a pure C interface
and promises to not throw any exceptions as well.

While here, add additional checks for list and
map values.

Fixes: https://github.com/sass/node-sass/issues/1131
Closes: https://github.com/sass/node-sass/issues/1127

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -60,22 +60,12 @@
               '-std=c++11'
             ],
             'OTHER_LDFLAGS': [],
-            'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
+            'GCC_ENABLE_CPP_EXCEPTIONS': 'NO',
             'MACOSX_DEPLOYMENT_TARGET': '10.7'
-          }
-        }],
-        ['OS=="win"', {
-          'msvs_settings': {
-            'VCCLCompilerTool': {
-              'AdditionalOptions': [
-                '/EHsc'
-              ]
-            }
           }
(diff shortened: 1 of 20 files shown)


saper:
@mgol @xzyfer let me know how far are your Macintoshes with this patch.

xzyfer:
>libsass should offer us a pure C interface and promises to not throw any exceptions as well

@saper I believe we're making steps in this direction in 3.3 - https://github.com/sass/libsass/pull/1544

saper:
We don't use `sass_value_op` in node-sass.

xzyfer:
>@xzyfer let me know how far are your Macintoshes with this patch.

943 passing (21s) :100:

saper:
Can you show me your full compiler flags now?

xzyfer:
https://gist.github.com/xzyfer/81e89cee8124f2480c81

xzyfer:
>We don't use sass_value_op in node-sass.

Yep I just meant we agree with the sentiment and we're working on it. Would be worth opening a LibSass issue to track for 3.4.

xzyfer:
:+1: Ship it when you're comfortable @saper

PR #1137, opened 2015-09-10
Run OSX builds in Travis CI

xzyfer:
I've been following https://github.com/travis-ci/travis-ci/issues/2311 and it seems like getting a node project to build on OSX should be doable with some persistence.

~~I'm mostly just PR'ing this so Travis will build it. Feel free to ignore for the time being. This could be a complete failure.~~

It took a bit of doing but this works! I'm only building OSX for 0.12 and 4 for now to keep builds quicky-ish.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,22 @@
-language: node_js
-
-env: SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true CC=gcc-4.7 CXX=g++-4.7 LINK=gcc-4.7 LINKXX=g++-4.7
-
+language: cpp
 compiler: gcc
-
 sudo: false
 
-node_js:
-  - "0.10"
-  - "0.12"
-  - "iojs-v1.0"
-  - "iojs-v1"
-  - "iojs-v2"
-  - "iojs-v3"
-  - "4"
+os:
+  - linux
+  - osx
+
+env:
+  global:
+    - SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
+  matrix:
+    - export NODE_VERSION="0.10"
+    - export NODE_VERSION="0.12"
+    - export NODE_VERSION="iojs-v1.0"
+    - export NODE_VERSION="iojs-v1"
+    - export NODE_VERSION="iojs-v2"
+    - export NODE_VERSION="iojs-v3"
+    - export NODE_VERSION="4"
 
 matrix:
   fast_finish: true
@@ -27,10 +30,29 @@ addons:
       - g++-4.7
 
 before_install:
-  - g++ --version
   - git submodule update --init --recursive
+  - git clone https://github.com/creationix/nvm.git ./.nvm
+  - source ./.nvm/nvm.sh
+  - nvm install $NODE_VERSION
+  - nvm use $NODE_VERSION
+  - npm config set python `which python`
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+      export CC="gcc-4.7";
+      export CXX="g++-4.7";
+      export LINK="gcc-4.7";
+      export LINKXX="g++-4.7";
+    fi
+  - gcc --version
+  - g++ --version
 
-after_success: npm run-script coverage
+script:
+  - npm install
+  - npm test
+
+after_success:
+  - if [ $TRAVIS_OS_NAME == "linux" ] && [ $NODE_VERSION == "4" ]; then
+      npm run-script coverage;
+    fi
 
 cache:
   directories:


xzyfer:
Thoughts @am11 @saper ?

saper:
There is nothing to think about :) only to act! I think we should do in this direction - and maybe whenever a tag is pushed we should be prepared to do a full run on OS X to create & deploy binaries.

One question - why do you need to install python from homebrew? something is not working?

am11:
This is brilliant! :+1: 
IMO, let OSX also run the entire matrix. The docker infrastructure is already delivering significant speed.

xzyfer:
> whenever a tag is pushed we should be prepared to do a full run on OS X to create & deploy binaries

That's the plan. Ideally I want to wire this and your AppVeyor work into node-pre-gyp

> let OSX also run the entire matrix

This as frustratingly slow. I double the number of simultaneous builds in Travis but didn't help too much.
TBH iojs 1,2 are on their way out IMHO once people the update Node 4.

xzyfer:
>why do you need to install python from homebrew

node-gyp wants python2 in the path. This isn't present by default but `brew` puts it the right place. The docs suggest python2 is install on the machines so maybe there's an env I can use to switch where node-gyp is looking.

xzyfer:
>node-gyp wants python2 in the path
>The docs suggest python2 is install on the machines so maybe there's an env I can use to switch where node-gyp is looking.

Turns out this is possible with

```
npm config set python `which python`
```

No more `brew`

xzyfer:
> This as frustratingly slow

Not running `brew` along with double the concurrent builds has made running the whole matrix feasible fast. :+1:

xzyfer:
screen shot 2015-09-11 at 12 06 32 pm

😍😍😍😍

am11:
> double the concurrent builds has made running the whole matrix feasible fast.

Wow! This is already in excellent shape. Lets merge it then! Further improvements can continue to follow over time. :smile:

saper:
I think that eventual migration to `node-pre-gyp` (as a replacement for our `scripts/*` infrastructure) is a different (larger?) project.

From the context I presume you would like to use `node-pre-gyp publish` feature? Somehow I feel we can be further ahead with GitHub releases and the built-in deployment feature of CIs.

I'd love to use `node-pre-gyp` to kill our own lifecycle scripts, but I don't want to add another custom script and another resource (S3) to have deployment of the binaries done.

One thing to decide before we go ahead with either solution is that should agree first how to trigger the release build - is having a separate `release` branch an option? New release would be commited to `master` (with the release notes as the commit comment) and then pushed to the `release` branch with a tag. The AppVeyor script from https://github.com/sass/node-sass/pull/1135 is ready to do it right away.

The only reason we might need a separate branch is http://bit.ly/twoBuilds issue - with a separate branch I can configure package building on commit push instead of tag push.

xzyfer:
>I think that eventual migration to node-pre-gyp [..] is a different (larger?) project.

I agree. Just something I want us to keep in mind.

>From the context I presume you would like to use `node-pre-gyp publish` feature? Somehow I feel we can be further ahead with GitHub releases and the built-in deployment feature of CIs.

IIRC `node-pre-gyp` backs on the built-in features of each CI platform. It's largely just a single point of config which is used to delegate to each CI platforms implementation of "release/publish artefacts to X"

> is having a separate `release` branch an option?

I have previous suggested we (libsass) adopt a `stable` branch to allow backporting features to old releases - https://github.com/sass/libsass/issues/796. This is not dissimilar from the `release` branch.

Personally this is something I would prefer to avoid.

> The only reason we might need a separate branch is http://bit.ly/twoBuilds issue 

Does this not help? https://help.appveyor.com/discussions/questions/261-only-deploy-when-theres-a-new-tag#comment_37132177

saper:
On Mon, 14 Sep 2015, Michael Mifsud wrote:

> >From the context I presume you would like to use `node-pre-gyp
> >publish` feature? Somehow I feel we can be further ahead with GitHub
> >releases and the built-in deployment feature of CIs.
> 
> IIRC `node-pre-gyp` backs on the built-in features of each CI
> platform. It's largely just a single point of config which is used
> to delegate to each CI platforms implementation of "release/publish
> artefacts to X"

node-pre-gyp does not seem to care which CI platform
is being used, it just takes two variables from the
build environment and uses them as the S3 keys,
and then uses AWS SDK to push to S3.
So, no magic there.

> > is having a separate `release` branch an option?
> 
> I have previous suggested we (libsass) adopt a `stable` branch to allow backporting features to old releases - https://github.com/sass/libsass/issues/796. This is not dissimilar from the `release` branch.
> 
> Personally this is something I would prefer to avoid.

Me too. But for now(?) we could just make "release" to fully mirror
"master" - we push current master to release only to trigger
the build. This branch would be nothing special and can be deleted
after the build is triggered.

> > The only reason we might need a separate branch is http://bit.ly/twoBuilds issue 
> 
> Does this not help? https://help.appveyor.com/discussions/questions/261-only-deploy-when-theres-a-new-tag#comment_37132177

Nope, see my comment below ... it triggers a while build on the release
branch (say, "master") commit and then does deployment only when there
is a tag. It is hard to push a tag and a commit at the same time
so that appveyor does not start two builds:

* commit-only -> builds 32- & 64-bit but does not deploy
* tag-only -> builds 32-bit only, does not publish artifacts and does not deploy;
        deployment rule is never reached.

I know it sounds complicated, I hope Fedor somehow responds to that...

Issue #1143, opened 2015-09-14
Clearly document our environment variables


xzyfer:
Inspired by #1106 (and others I can't remember atm).

We have a handful of really useful environment variables, with subtly different behaviour. These are currently poorly, or undocumented which leads to people not knowing they exist, or how they work.

**We need to document theses**

Off the top of my head we ENV variables for installing/downloading the binaries, as well compiling the binary in dev mode. Below is some, but not all, of the ENV variables that directly affect node-sass.

```
SASS_BINARY_NAME
SASS_BINARY_SITE
SASS_BINARY_PATH
```

am11:
:+1: 
Incidentally, should we add two more columns to list down their package.json cofig and CLI argument counterparts (where applicable)?

PR #1145, opened 2015-09-15
Use sass binary url from .npmrc if available

fh1ch:
This PR adds the possibility to configure the sass-binary URL in the .npmrc configuration file (e.g.  `sass_cdnurl=http://example.com/sass-binaries`). This is also used the same way in [PhantomJS](https://github.com/Medium/phantomjs/blob/master/install.js#L23) by pointing the download URL to a mirror without having to set an environment variable manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,6 +509,28 @@ Also, note `--importer` takes the (absolute or relative to pwd) path to a js fil
 The `--source-map` option accepts a boolean value, in which case it replaces destination extension with `.css.map`. It also accepts path to `.map` file and even path to the desired directory. 
 When compiling a directory `--source-map` can either be a boolean value or a directory.
 
+## Binary configuration parameters
+
+node-sass supports different configuration parameters to change settings related to the sass binary such as binary name, binary path or alternative download path. Following parameters are supported by node-sass:
+
+Variable name    | .npmrc parameter | Process argument   | Value
+-----------------|------------------|--------------------|------
+SASS_BINARY_NAME | sass_binary_name | --sass-binary-name | path
+SASS_BINARY_SITE | sass_binary_site | --sass-binary-site | URL
+SASS_BINARY_PATH | sass_binary_path | --sass-binary-path | path
+
+These parameters can be used as environment variable:
+
+* E.g. `export SASS_BINARY_SITE=http://example.com/`
+
+As local or global [.npmrc](https://docs.npmjs.com/misc/config) configuration file:
+
+* E.g. `sass_binary_site=http://example.com/`
+
+As a process argument:
+
+* E.g. `npm install node-sass --SASS_BINARY_SITE=http://example.com/`
+
 ## Post-install Build
 
 Install runs only two Mocha tests to see if your machine can use the pre-built [libsass] which will save some time during install. If any tests fail it will build from source.
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -48,7 +48,8 @@ function getRuntimeInfo() {
 
 /**
  * Get binary name.
- * If environment variable SASS_BINARY_NAME or
+ * If environment variable SASS_BINARY_NAME,
+ * .npmrc variable sass_binary_name or
  * process argument --binary-name is provided,
  * return it as is, otherwise make default binary
  * name: {platform}-{arch}-{v8 version}.node
@@ -63,6 +64,8 @@ function getBinaryName() {
     binaryName = flags['--sass-binary-name'];
   } else if (process.env.SASS_BINARY_NAME) {
     binaryName = process.env.SASS_BINARY_NAME;
+  } else if (process.env.npm_config_sass_binary_name) {
+    binaryName = process.env.npm_config_sass_binary_name;
   } else if (pkg.nodeSassConfig && pkg.nodeSassConfig.binaryName) {
     binaryName = pkg.nodeSassConfig.binaryName;
   } else {
@@ -80,7 +83,8 @@ function getBinaryName() {
  * site on GitHub.
  *
  * The default URL can be overriden using
- * the environment variable SASS_BINARY_SITE
+ * the environment variable SASS_BINARY_SITE,
+* .npmrc variable sass_binary_site or
  * or a command line option --sass-binary-site:
  *
  *   node scripts/install.js --sass-binary-site http://example.com/
@@ -104,6 +108,7 @@ function getBinaryName() {
 function getBinaryUrl() {
   var site = flags['--sass-binary-site'] ||
              process.env.SASS_BINARY_SITE  ||
+             process.env.npm_config_sass_binary_site ||
              (pkg.nodeSassConfig && pkg.nodeSassConfig.binarySite) ||
              'https://github.com/sass/node-sass/releases/download';
 
@@ -121,7 +126,8 @@ sass.runtime = getRuntimeInfo();
 
 /**
  * Get binary path.
- * If environment variable SASS_BINARY_PATH or
+ * If environment variable SASS_BINARY_PATH,
+ * .npmrc variable sass_binary_path or
  * process argument --sass-binary-path is provided,
  * select it by appending binary name, otherwise
  * make default binary path using binary name.
@@ -140,6 +146,8 @@ sass.getBinaryPath = function(throwIfNotExists) {
     binaryPath = flags['--sass-binary-path'];
   } else if (process.env.SASS_BINARY_PATH) {
     binaryPath = process.env.SASS_BINARY_PATH;
+  } else if (process.env.npm_config_sass_binary_path) {
+    binaryPath = process.env.npm_config_sass_binary_path;
   } else if (pkg.nodeSassConfig && pkg.nodeSassConfig.binaryPath) {
     binaryPath = pkg.nodeSassConfig.binaryPath;
   } else {
--- a/test/runtime.js
+++ b/test/runtime.js
@@ -24,176 +24,117 @@ describe('runtime parameters', function() {
       require(extensionsPath);
     });
 
-    describe('in package.json', function() {
-      it('should use the binary path', function() {
-        require.cache[packagePath].exports.nodeSassConfig = { binaryPath: 'ccc' };
-        require(extensionsPath);
+    describe('configuration precedence should be respected', function() {
 
-        assert.equal(process.sass.binaryPath, 'ccc');
+      describe('SASS_BINARY_NAME', function() {
+        beforeEach(function() {
+          process.argv.push('--sass-binary-name', 'aaa');
+          process.env.SASS_BINARY_NAME = 'bbb';
+          process.env.npm_config_sass_binary_name = 'ccc';
+          require.cache[packagePath].exports.nodeSassConfig = { binaryName: 'ddd' };
+        });
+
+        it('command line argument', function() {
+          require(extensionsPath);
+          assert.equal(process.sass.binaryName, 'aaa_binding.node');
+        });
+
+        it('environment variable', function() {
+          process.argv = [];
+          require(extensionsPath);
+          assert.equal(process.sass.binaryName, 'bbb_binding.node');
+        });
+
+        it('npm config variable', function() {
+          process.argv = [];
+          process.env.SASS_BINARY_NAME = null;
+          require(extensionsPath);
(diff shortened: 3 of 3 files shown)


huerlisi:
Cool, that looks like something we've been looking for!

saper:
That's cool, but let's get at least naming consistent. #1143 is a nice attempt to document those.

Also in the implementation we should just wrap the names of all parameters with `npm_config_` and looks them up in the environment.

fh1ch:
Totally agree. I hope `npm_config_sass_binary_site` is ok as name. I'll also take care of the documentation and add it to this PR.

The second point I didn't really get. What's your idea there?

xzyfer:
Firstly, thanks for your contribution @fh1ch, we appreciated.

Secondly, some house keep. I the future it's preferred you open an issue to discuss new features before implementing them. Odds are something similar has previously been discussed, or the team may already have plans in the area of interest.

With regards to configuration we do have plans for standardising on `nodeSassConfig` in `package.json` to address concerns we have about code portability. In this case however it doesn't preclude your proposed addition.

****

Specifically to this PR we currently support 3 ENV variable that affect the binary download and/or resolution process, `BINARY_PATH`, `BINARY_SITE`, and `BINARY_NAME`. This PR should add support for all these variables, as well test coverage for these new options.

saper:
I :cupid pull requests :) 

@fh1ch check this out https://github.com/sass/node-sass/blob/master/scripts/build.js#L133-L136
- we just transform the same string into the command line option and the environment variable name.

One just needs probably to stick ' || process.env['npm_config_' + subject]' somewhere to make it work as you propose.

xzyfer:
>we should just wrap the names of all parameters with npm_config_ and look them up in the environment.

@saper I also do not understand what we're accomplishing here. I assume you're referring to the following:

>Any environment variables that start with npm_config_ will be interpreted as a configuration parameter. For example, putting npm_config_foo=bar in your environment will set the foo configuration parameter to bar. Any environment configurations that are not given a value will be given the value of true. Config values are case-insensitive, so NPM_CONFIG_FOO=bar will work the same.
>Source - https://docs.npmjs.com/misc/config#environment-variables

This however isn't adding `.npmrc` support which I think is worth while.

We already support `.npmrc` for [proxy settings](https://github.com/sass/node-sass/blob/master/scripts/install.js#L72-L74) so extending to the binary resolution make sense.

xzyfer:
> One just needs probably to stick `|| process.env['npm_config_' + subject]` somewhere to make it work as you propose

I'm not sure I follow this reasoning. We already have support for the `--sass-binary-site` and `sass-binary-path` flags. The `npmconf` module should handle the npm config resolution process i.e. `npm_conf_foo` vs `foo` on `.npmrc`

fh1ch:
I applied the .npmrc configuration possibility also to the other environment variables as proposed by @xzyfer. The're now also covered by documentation and runtime tests.

test/runtime.js
```diff
@@ -173,22 +206,44 @@ describe('runtime parameters', function() {
         assert.equal(process.sass.binaryUrl.substr(0, URL.length), URL);
       });
     });
-    describe('checking if environment variable overrides package.json', function() {
+    describe('checking if environment variable overrides .npmrc configuration variable', function() {
```
xzyfer: This test should be added in addition to the existing `package.json` tests. The test you're removing here is still important.

lib/extensions.js
```diff
@@ -63,6 +64,8 @@ function getBinaryName() {
     binaryName = flags['--sass-binary-name'];
   } else if (process.env.SASS_BINARY_NAME) {
     binaryName = process.env.SASS_BINARY_NAME;
+  } else if (process.env.npm_config_sass_binary_name) {
+    binaryName = process.env.npm_config_sass_binary_name;
```
xzyfer: I don't think these `npm_config_` work as expected. You should use `npmconf` instead.

fh1ch: I only know `npmconf` as NPM module and not as an env namespace. I still think `npm_config_` is the right thing :smile:. As per NPM documentation:
> Any environment variables that start with npm_config_ will be interpreted as a configuration parameter.

This mechanism works bidirectional and npm exposes all configuration parameter into this env namespace which will be then passed to all child processes (like in the phantom.js installation). Parameters could be set in a local/global .npmrc file or passed directly as argument.

I also tested it locally using a .npmrc file in the project with an alternative download path and it worked like a charm.

test/runtime.js
fh1ch: I tried to avoid redundancy there. The test is still existing but tests that .npmrc parameters can override package.json parameters ( see `checking if .npmrc configuration variable overrides package.json`). And `checking if environment variable overrides .npmrc configuration variable` checks if env parameters can override .npmrc parameters. This covers the whole chain form a hierarchical point of view. 

I can of course add the original tests again if you like.

lib/extensions.js
xzyfer: > This mechanism works bidirectional and npm exposes all configuration parameter into this env namespace

Do you have a link to the documentation for this. I was under the impression it was a one way street. ENV overwrote .npmrc but not the other way around. Happy to be wrong.

test/runtime.js
xzyfer: I don't think either approach is correct. We no have 4 ways to set these configurations. Testing they get set is important, testing their precedence is most important. The better tests would be asserting, per variable, that precedence is respected.

```
describe('configuration precedence should be respected', () => {
  describe('SASS_BINARY_PATH', () => {
    beforeEach(() => {
      process.env.SASS_BINARY_PATH = 'foo';
      process.env.npm_config_sass_binary_path = 'bar';
      require.cache[packagePath].exports.nodeSassConfig = { binaryPath: 'baz' };
      require(extensionsPath);
    });
    it('environment variable', () => {
      assert.equal(process.sass.binaryPath, 'baz');
    });
    it('npm config', () => {
      process.env.npm_config_sass_binary_path = null;
      assert.equal(process.sass.binaryPath, 'bar');
    });
    it('package.json', () => {
      process.env.npm_config_sass_binary_path = null;
      require.cache[packagePath].exports.nodeSassConfig = {};
      assert.equal(process.sass.binaryPath, 'foo');
    });
  })
})
```

lib/extensions.js
fh1ch: I guess that this is a bit the problem with the NPM documentation. It is not clearly written in the config part [1] that it works bidirectional. In fact NPM exposes **all** parameters (configuration, metadata...) to the env and is mentioned quiet hidden in the documentation like [2] and [3]. This is actually a quiet handy mechanism for configuration and also used by other projects [4].

It might make sense enhance this in the upstream npm documentation.

[1] https://docs.npmjs.com/misc/config#environment-variables
[2] https://docs.npmjs.com/misc/scripts#current-lifecycle-event
[3] https://docs.npmjs.com/files/package.json#config
[4] https://github.com/Medium/phantomjs/blob/master/install.js#L23

test/runtime.js
fh1ch: You're right, this is a much better solution. I will start refactoring all configuration unit tests.

xzyfer: You're awesome!

This should also reduce some bloat from these specs! \o/

lib/extensions.js
xzyfer: I did not know this! Thanks for dropping some knowledge on me

fh1ch: You're very welcome :smile: Thank you for this awesome project!

test/runtime.js
fh1ch: The unit tests are now implemented in the way you suggested. It's now much smaller and cleaner without decreasing the coverage. I'm hoping your happy with the current solution. Just tell me if there is anything else to do.

xzyfer:
Amazing work! :heart_eyes:

fh1ch:
Thank you very much, I'm really humbled :heart_eyes: I'm glad if I can help to bring this awesome project a bit forward.

xzyfer:
If you're comfortable with git rebase I'd ask you to please squash these commits together

- `Use sass binary url from .npmrc if available`
- `Adapted all variables to support aligned .npmrc parameters`

and 

- `Added runtime tests for .npmrc parameters`
- `Refactored runtime unit test to follow new configuration schema`

****

I the end there would be 3 commits

- A commit adding .npmrc support
- A commit updating the tests
- A commit updating the docs

README.md
```diff
@@ -509,6 +509,28 @@ Also, note `--importer` takes the (absolute or relative to pwd) path to a js fil
 The `--source-map` option accepts a boolean value, in which case it replaces destination extension with `.css.map`. It also accepts path to `.map` file and even path to the desired directory. 
 When compiling a directory `--source-map` can either be a boolean value or a directory.
 
+## Binary configuration parameters
+
+node-sass supports different configuration parameters to change settings related to the sass binary such as binary name, binary path or alternative download path. Following parameters are supported by node-sass:
+
+Variable name    | .npmrc parameter | Process argument   | Value
+-----------------|------------------|--------------------|------
+SASS_BINARY_NAME | sass_binary_name | --binary-name      | path
```
saper: I think it's `--sass-binary-name`. At least I hope so :)

saper:
On Wed, 16 Sep 2015, Michael Mifsud wrote:

> 
> - `Use sass binary url from .npmrc if available`
> - `Adapted all variables to support aligned .npmrc parameters`
> 
> and 
> 
> - `Added runtime tests for .npmrc parameters`
> - `Refactored runtime unit test to follow new configuration schema`
> 
> ****
> 
> I the end there would be 3 commits
> 
> - A commit adding .npmrc support
> - A commit updating the tests

I don't mind having those two together, so that we are 100% green
between commits.

README.md
fh1ch: Yes, you're right. Thank you very much for the input. I will correct and push it when I'm rebasing the stuff.

fh1ch:
No problem, I will quickly rebase that stuff as you two suggest it.

xzyfer:
Presumably the existing test should still pass with the new functionality. 

To be honest I'm fine squashing this all into 1 commit. It's all tightly related.

fh1ch:
There are now only 3 commits. Should a rebase them all into 1?
Would make sense since Travis striked again :smile:

saper:
On Wed, 16 Sep 2015, Fabio Huser wrote:

> There are now only 3 commits. Should a rebase them all into 1?
> Would make sense since Travis striked again :smile: 

I am just trying to quickly fix some bug in Travis tests
so I'd love more testing :)

fh1ch:
> On Wed, 16 Sep 2015, Fabio Huser wrote:
 There are now only 3 commits. Should a rebase them all into 1?
 Would make sense since Travis striked again :smile:
I am just trying to quickly fix some bug in Travis tests
so I'd love more testing :)

Good to now :smile: Are these 3 commits now ok for you or should I squash them all together?

saper:
On Wed, 16 Sep 2015, Fabio Huser wrote:

> > On Wed, 16 Sep 2015, Fabio Huser wrote:
>  There are now only 3 commits. Should a rebase them all into 1?
>  Would make sense since Travis striked again :smile:
> I am just trying to quickly fix some bug in Travis tests
> so I'd love more testing :)
> 
> Good to now :smile: Are these 3 commits now ok for you or should I squash them all together?

As they say those days, LGTM!

fh1ch:
Awesome! Just tell me if I can help with something.

xzyfer:
@fh1ch this is all good! Our plan is to ship in 3.4.0 hopefully next week.

The reason for 3.4.0 is that this is feature and by semver it should be a minor release.
The reason for not releasing 3.4.0 immediately is that we want to 3.4.0 to have LibSass 3.3.0-stable. 

This is largely to help communicate to people which version of node-sass has which version of LibSass.
Also because LibSass follows pre 1.0.0 semver atm so 3.3 is actually a major version bump from 3.2.

We really appreciate your work and we're super keen to get it our the wild ASAP. We expect to have 3.4.0-beta out this week!

fh1ch:
@xzyfer: Good to know, thanks :smile: It doesn't hurries for me so it's not that important in which release it gets. But it's nice to have libSass-3.3.0 in.

Thank you very much guys, it was a pleasure for me contributing to this project.

PR #1147, opened 2015-09-16
Don't compile delay hook into libsass

saper:
Windows delay hook feature is only needed
on the main module, linked into the shared
library. Don't compile and include
it second time in the bundled libsass.

Related: https://github.com/nodejs/node-gyp/issues/732

### DIFF
--- a/src/libsass.gyp
+++ b/src/libsass.gyp
@@ -2,6 +2,7 @@
   'targets': [
     {
       'target_name': 'libsass',
+      'win_delay_load_hook': 'false',
       'type': 'static_library',
       'defines': [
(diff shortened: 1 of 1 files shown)


saper:
@am11 it seems that the delay hook gets compiled by default into every gyp target. We certainly don't need in the libsass part, we only need it in the main module.

Can you check if we still can link at runtime to `iojs.exe` and `node.exe` without problems after this change?

am11:
Tested with iojs v3.3.1 and its alias `node` and both compile Sass code with same binary (generated by `iojs scripts/build -f`).
LGTM. :+1: :shipit:

PR #1151, opened 2015-09-17
Add a hint for Chinese installers

pmq20:
Fix dlmanning/gulp-sass#342


saper:
Is the proxy mandatory? Why is this problem popping out?

pmq20:
@saper The Internet censorship causes a lot of websites abroad inaccessible, this one included.

saper:
@pmq20 I have re-worded it a bit (it's not only for China, also for many Western corporations) in https://github.com/pmq20/node-sass/pull/1 - I hope you like it this way!

pmq20:
@saper I like it, thanks!

PR #1163, opened 2015-09-21
Add CLI option for watcher polling interval

### LINKED ISSUE #801: Support for polling interval with --watch
If using `node-sass --watch` in an environment where Gaze falls back to polling (e.g. networked file system) the polling can cause unnecessary CPU-load.

Would you consider some method for passing [the `interval` option](https://www.npmjs.com/package/gaze#properties) to Gaze?

xzyfer:
In some environments the fs watcher falls back to polling for
changes. Both our current watcher and the proposed new watcher
both had support for user-defined polling intervals. I see no
reason not to add this for users who want extra control.

Fixes #801.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -47,6 +47,7 @@ var cli = meow({
     '  --source-map-root          Base path, will be emitted in source-map as is',
     '  --include-path             Path to look for imported files',
     '  --follow                   Follow symlinked directories',
+    '  --poll-interval            The filesystem polling interval in ms when the watcher has to poll',
     '  --precision                The amount of precision allowed in decimal numbers',
     '  --importer                 Path to .js file containing custom importer',
     '  --functions                Path to .js file containing custom functions',
@@ -240,7 +241,11 @@ function watch(options, emitter) {
     watch.push(i);
   }
 
-  var gaze = new Gaze();
+  if (options.pollInterval && typeof options.pollInterval !== 'number') {
+    emitter.emit('error', 'The --poll-interval must be an integer.');
+  }
+
+  var gaze = new Gaze(null, { interval: options.pollInterval });
   gaze.add(watch);
   gaze.on('error', emitter.emit.bind(emitter, 'error'));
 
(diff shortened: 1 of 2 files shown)


xzyfer:
Feels @saper ?

saper:
Everybody should have noticed by now that I avoid watcher issues like hell. Our watcher tests are still wrong and we need to fix them.

On a serious note, I happen to debug NFS performance issues for living and I'd like to know more about #801.

xzyfer:
>Doesn't feel like `--interval` is a right way to do it.

Could you elaborate? Poll interval control is pretty common to FS watchers.

I personally think the default 500ms is perfectly reasonable. I'm convinced we need to ability to change it.

xzyfer:
To support that feeling, https://github.com/sass/node-sass/issues/801 been around for about 6 months with other supports. Also Ruby Sass does not support this via their CLI.

saper:
I wonder what kind of system is this when 500ms is inducing too much load. Maybe we are ineffective in some other, nasty way and we can avoid that. Like we don't use `stat()` where we should.

xzyfer:
I don't think it's load. More likely it's simply an annoying delay. We frontend developers are very touch about things feeling instant. Considering we can get sub-second compilations, wait 500ms for the watcher to kick in can feel like an eternity. Speaking from experience.

saper:
#801 says something about CPU load. So in reality you are eager to consider `node-sass --ultra-productive-developer`, shortened to `node-sass -U` ?

saper:
I think I'll abstain on this one. I know something about kernel polling mechanisms like `kqueue` and `epoll`, I've heard there are issues with Mac OS X watcher API (the main reason so many 'watcher' packages exist in node I think) but I have little clue how what we are doing right now in JavaScript translates to a real **something being done** on the system that might affect system performance if polling too often. Until I study the APIs and carry out some NFS tests the answer from me will be: "I don't know".

xzyfer:
Closing this. It appears to be an isolated case.

saper:
:tada

Issue #1026, opened 2015-07-08
CLI --source-map option broken


johnpolacek:
In a new directory, create a test.scss file with one line of css, save then in CLI run:
```node-sass test.scss test.css --source-map```

You will get the following error:
```
var sourceMapIsDirectory = options.sourceMapOriginal.indexOf('.map', optio
TypeError: Object true has no method 'indexOf'
```

**Comment by am11** [comment]
This was a breaking change in v3, see https://github.com/sass/node-sass/pull/964.

In your case, use `node-sass test.scss test.css --source-map true` or ` node-sass test.scss test.css --source-map custom-path/to/whatever.map`

**Comment by nhunzaker** [comment]
I just hit this as well and the error message is very cryptic. What do you think about improving the error here if a value isn't present?

**Comment by glassdimly** [comment]
I am of the opinion that node-sass --help ought to note that --source-map needs a boolean parameter. Otherwise most people would reasonably assume that this was just a standard unix parameter that did not need a boolean. It should not be necessary to google this bug or read the source: that's what the --help or man page is for.

**Comment by sant123** [comment]
This error is still happening... I have this

```json
  "scripts": {
    "ts" : "node_modules\\.bin\\tsc -w -p .",
    "sass_env" : "SET node-sass=node_modules\\.bin\\node-sass",
    "sass_compile" : "npm run sass_env && node-sass styles -o styles --output-style compressed --source-map ",
    "sass" : "npm run sass_compile && node-sass styles -wo styles --output-style compressed --source-map "
  }
```

When I execute ```npm run sass``` I get the following error:

``` 
D:\santiagoah\informacion\Desktop\Visor de documentos>npm run sass                                                                                                 
                                                                                                                                                                   
> visor-de-documentos@1.0.0 sass D:\santiagoah\informacion\Desktop\Visor de documentos                                                                             
> npm run sass_compile && node-sass styles -wo styles --output-style compressed                                                                                    
                                                                                                                                                                   
                                                                                                                                                                   
> visor-de-documentos@1.0.0 sass_compile D:\santiagoah\informacion\Desktop\Visor de documentos                                                                     
> npm run sass_env && node-sass styles -o styles --output-style compressed --source-map                                                                            
                                                                                                                                                                   
                                                                                                                                                                   
> visor-de-documentos@1.0.0 sass_env D:\santiagoah\informacion\Desktop\Visor de documentos                                                                         
> SET node-sass=node_modules\.bin\node-sass                                                                                                                        
[...]

Issue #793, opened 2015-03-24
Unable to get it working on Windows


PhiLhoSoft:
Either it is a true bug report against the library, or it is one against the documentation (the top readme)...
```
> node -v
v0.12.0
> npm -v
2.7.3
> node C:\Dev\Web\Playground\node_modules\node-sass\bin\node-sass -v
node-sass       3.0.0-alpha.0   (Wrapper)       [JavaScript]
libsass         3.1.0           (Sass Compiler) [C/C++]
```
As you can see, I develop on Windows 7...
I am making a simple "skeleton" web application at https://github.com/PhiLhoSoft/Web-Playground
The idea is to experiment with a development stack, including Sass, of course.
I chose to skip Gulp / Grunt in favor of pure npm task running, so I have to learn the command line versions of the tools.

After reading the docs, I tried to generate a CSS file from my Sass one with the following script:

    "css": "node node-sass-run.js"

where the JS ile is as follows:
```
var sass = require("node-sass");

sass.render(
{
	file: "src/main/sass/playground.scss",
	includePaths: [ "src/main/sass" ],
	outFile: "src/main/css",
	outputStyle: "nested",
	sourceComments: true,
	sourceMap: true,
	importer: function(url)
	{
		return { file: url };
	}
},
function(error, result)  // >= v3.0.0
{
	if (error)
	{
		console.log(error);
	}
	else
	{
		console.log("Result path: " + result.path);
		console.log(result.stats);
//~ 		console.log(result.css.toString());
//~ 		console.log(result.map.toString()); // or console.log(JSON.stringify(result.map));
	}
});
```
When I run it, it displays no errors:
```
> node node-sass-run.js

{ file: 'philhosoft-base' }
Result path: undefined
{ entry: 'C:\\PersonnelPhilippeLhoste\\PhiLhoSoft\\Web\\Playground\\src\\main\\sass\\playground.scss',
  start: 1427190066028,
  includedFiles:
   [ 'C:/PersonnelPhilippeLhoste/PhiLhoSoft/Web/Playground/src/main/sass/_philhosoft-base.scss',
     'C:/PersonnelPhilippeLhoste/PhiLhoSoft/Web/Playground/src/main/sass/playground.scss' ],
  end: 1427190066050,
  duration: 22 }
```
Cool. Except I find no CSS file in the `src/main/css` folder. I thought the `outFile` parameter was used as destination... I also tried with an absolute path.
[...]

Issue #425, opened 2014-09-21
Wrong path used for sourceMap


halfdan:
This commit seems to break the path used in sourceMap generation: https://github.com/sass/node-sass/commit/fd6a201dff36a75e6b0658ba4051e44edd3c38d6

We're using `node-sass` and `grunt-sass` for Ghost with the following setup:

```
// ### grunt-sass
// compile sass to css
sass: {
    compress: {
        options: {
            style: 'compressed',
            sourceMap: true
        },
        files: [
            { dest: 'core/client/assets/css/<%= pkg.name %>.min.css', src: 'core/client/assets/sass/screen.scss' },
            { dest: 'core/client/docs/dist/css/<%= pkg.name %>.min.css', src: 'core/client/assets/sass/screen.scss' }
        ]
    }
},
```

This however leads to the following error:

```
Running "sass:compress" (sass) task
Warning: ENOENT, open '/home/vagrant/code/Ghost/core/client/assets/css/core/client/assets/css/ghost.min.css.map' Use --force to continue.
```

---

The two lines used in `sass.js` generate that path:

```
dir = path.dirname(options.outFile); // Contains the full path to the outfile
sourceMapFile = path.resolve(dir, options.sourceMap); // sourceMap also contains the full path
```

STRML:
Is there a reason why this was closed? This appears to cause incorrect source file locations in grunt-sass.

spelufo:
Option parsing is still badly broken. You have examples for a thousand frameworks on the readme but your cli doesn't work.

```bash
$ ls static/
main.sass
$ node_modules/.bin/node-sass -r static/
{
  "status": 4,
  "message": "File to read not found or unreadable: static/",
  "code": 4
}
$ node_modules/.bin/node-sass -w static/main.sass -o static/main.css
Provide a Sass file to render

  Example
    node-sass --output-style compressed foobar.scss foobar.css
    cat foobar.scss | node-sass --output-style compressed > foobar.css
```

am11:
@spelufo it is because you are not reading the docs carefully. `-o` option is for output directory.

> You have examples for a thousand frameworks on the readme but your cli doesn't work.

Possibly.. This is an open source project and the philosophy that follow is: DIY

glassdimly:
```
cat sass/test.scss | node-sass > test.css
{
  "message": "file to import not found or unreadable: base\nCurrent dir: ",
  "column": 9,
  "line": 1,
  "file": "stdin",
  "status": 1
}
```
[...]

PR #1177, opened 2015-09-24
adding detailed example to cli --help

glassdimly:
See https://github.com/sass/node-sass/issues/425, https://github.com/sass/node-sass/issues/1026, and https://github.com/sass/node-sass/issues/793


saper:
Is that .css at the end or did you intend .scss?

glassdimly:
I've added a description to my example, I think it's helpful. Let me know if what I have written matches  your expectation of what will happen.

To answer your questions, yes, I intend to output the file to css, not scss, and I believe that the final argument is supposed to be the output file. Am I correct in that? I had it working this way.

bin/node-sass
```diff
@@ -27,6 +27,9 @@ var cli = meow({
     'Example',
     '  node-sass --output-style compressed foobar.scss foobar.css',
     '  cat foobar.scss | node-sass --output-style compressed > foobar.css',
+    'Example: Watch the sass directory for changes, compile with sourcemap to css/style.css',
```
xzyfer: @saper I think this description is useful in #1178

PR #1216, opened 2015-10-24
Disable all tests with --watch

saper:
Testing cli this way is completely
unreliable. A new method of testing
is needed.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -215,7 +215,9 @@ describe('cli', function() {
         done();
       });
     });
+  });
 
+  describe.skip('using --watch option', function() {
     it('should not exit with the --watch option', function(done) {
       var src = fixture('simple/index.scss');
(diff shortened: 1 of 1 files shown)


xzyfer:
:+1:

xzyfer:
Looks like you missed one?

```
 1) cli "before all" hook for "should return a correct version info":
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
      at null.<anonymous> (C:\projects\node-sass\node_modules\mocha\lib\runnable.js:189:19)
```

saper:
Looks like node 5.x has watching broken on my favourite OS https://github.com/nodejs/node/issues/3541

saper:
this before all hook checking libsass version must be rewritten in an async way. Will fix of course.

xzyfer:
I believe this has been done.

Issue #1219, opened 2015-10-26
Importer returning contents and file no longer retain absolute paths


gmac:
There is a significant breaking API change in the way imported file contents reference subsequent imports between versions 3.3.3 and 3.4.0.

In **v3.3.3**, absolute file references provided with imported contents would be passed along for resolving subsequent imports. Example behavior in v3.3.3:

```javascript
// @import "a"
done({
  file: '/Users/me/sass/lib/a.scss',
  contents: '@import "b"'
});
```

_Next importer call for `@import "b"`_:

```javascript
function(uri, previousUri, done) {
   // uri: "b",
   // previousUri: "/Users/me/sass/lib/a.scss"
}
```

In **v3.4.0**, absolute file references are apparently discarded, and only the previous slug is passed along to subsequent importer calls. Example behavior in v3.4.0:

```javascript
// @import "a"
done({
  file: '/Users/me/sass/lib/a.scss',
  contents: '@import "b"'
});
```

_Next importer call for `@import "b"`_:

```javascript
function(uri, previousUri, done) {
   // uri: "b",
   // previousUri: "a"
}
```

This is a significant detriment to content importers that rely on the previously resolved content URI for file mapping, and a major enough change that I would have expected it to be part of a major version bump if intentional. If this change _is_ intentional, I'd be curious about the rationale for removing this exceptionally valuable piece of data from the implementation. Without this data, importers now need to maintain a complex external dependency mapping table that can –at best– still only guess upon a previous slug's mapped file.

mgreter:
I guess https://github.com/sass/node-sass/blob/master/src/binding.cpp#L12 is the culprit, should probably be `sass_import_get_imp_path`. The names have changed in libsass API and it looks like this got mixed up on the node sass side. Looks like node-sass could use some more unit tests.

gmac:
`Looks like node-sass could use some more unit tests.` :+1:

mgreter:
@gmac I guess their willing to accept Pull Requests in that direction :wink:

gmac:
Noted, and considered.

xzyfer:
This was fixed in https://github.com/sass/node-sass/pull/1220. Will land in v3.4.2.

Rush:
I am experiencing this problem as well, waiting for release then!

crypticsymbols:
+1 eagerly awaiting this

xzyfer:
I believe this is fixed in v3.4.2. Please open if the issue persists.

PR #1220, opened 2015-10-26
The previous file should be the absolute path, not the string used to import it.

chriseppstein:
node-sass 3.4 had a breaking public API change. This patch fixes the path that is passed as the previous file path when resolving imports. In node-sass 3.4 this value changed so that the import uri was passed. The issue here is that any code needed to resolve one import against another quickly falls apart if the resolved path returned by the importer is not what is passed back to the importer as the current sass file.

Basically, I think the updates for node-sass 3.4 to use libsass 3.3 mistakenly used the wrong value.

### DIFF
--- a/src/binding.cpp
+++ b/src/binding.cpp
@@ -9,7 +9,7 @@ Sass_Import_List sass_importer(const char* cur_path, Sass_Importer_Entry cb, str
 {
   void* cookie = sass_importer_get_cookie(cb);
   struct Sass_Import* previous = sass_compiler_get_last_import(comp);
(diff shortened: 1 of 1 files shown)


mgreter:
Related to https://github.com/sass/node-sass/issues/1219 (see my comments in there, which basically comes to the same conclusion).

xzyfer:
Thanks @chriseppstein this will land in v3.4.2 in the next couple days.

vlad:
Thanks for fixing this, hope this lands soon!

PR #1269, opened 2015-11-16
Update mocha dependency to >= 2.3.4

ockham:
That latest version includes an update (mochajs/mocha@039a8a7) to a newer version (2.2.0) of `debug`, which in turn has updated (visionmedia/debug@adc2402) its dependency on `ms` to 0.7.1 in order to include a [security fix](https://nodesecurity.io/advisories/46).

Untested!

Ideally, please also do a point release including this fixed dependency.


xzyfer:
Thanks for this @ockham

PR #1292, opened 2015-12-03
Add .bin to file path cd node_modules\node-sass\vendor

benclarke:
I think you're missing a .bin in that file path...

apologies if I'm doing this wrong. newbie.


xzyfer:
Thanks for the report @benclarke but this is not a bug. The `.bin` folder is created by npm.

benclarke:
sorry my bad.

PR #1307, opened 2015-12-19
assert(a === b) => assert.strictEqual(a, b)

saper:
When the test fails, we are getting actual
values in the error message instead of

    + true
    - false


xzyfer:
Nice.

Issue #1296, opened 2015-12-05
Async custom importer sass.NULL behavior different than sync importer


lingram-rmn:
Returning `sass.NULL` asynchronously does not result in the default libsass import behavior. It's quite possible that this will turn out to be an upstream bug in libsass, but I thought I would start here.

# Repro
Run the following in a context where node-sass is npm-installed:

```javascript
var assert = require('assert');
var fs = require('fs');
var path = require('path');
var sass = require('node-sass');

fs.writeFileSync(
  path.resolve('entry.scss'),
  [
    '/* entry.scss */',
    '@import "./import.scss";'
  ].join('\n')
);

fs.writeFileSync(
  path.resolve('import.scss'),
  '/* import.scss */'
);

var syncResult;
sass.render({
  file: path.resolve('entry.scss'),
  importer: function (url, prev, done) {
    return sass.NULL;
  }
}, function (err, result) {
  if (err) {
    throw err;
  }

  syncResult = result.css.toString();
});

sass.render({
  file: path.resolve('entry.scss'),
  importer: function (url, prev, done) {
    done(sass.NULL);
  }
}, function (err, result) {
  if (err) {
    throw err;
  }

  setTimeout(function () {
    assert.strictEqual(result.css.toString(), syncResult);
  }, 0);
});
```

## Expected
Script exits without an error.

## Observed
The equality assertion fails:

```
assert.js:86
  throw new assert.AssertionError({
        ^
AssertionError: '/* entry.scss */\n' === '/* entry.scss */\n/* import.scss */\n'
    at null._onTimeout (/Users/lingram/Projects/node-sass-importer-bug/test.js:44:12)
    at Timer.listOnTimeout (timers.js:119:15)
```

JohnAlbin:
@lingram-rmn I'm not certain, but is this a duplicate of issue #1291? Or possibly a manifestation of the same problem that the importer's node-sass is not equal to the node-sass that calls the importer?

lingram-rmn:
I don't think it's a duplicate, no. There's only one node-sass instance
involved.
On Dec 5, 2015 7:48 PM, "John Albin Wilkins" <notifications@github.com>
wrote:

> @lingram-rmn <https://github.com/lingram-rmn> I'm not certain, but is
> this a duplicate of issue #1291
> <https://github.com/sass/node-sass/issues/1291>? Or possibly a
> manifestation of the same problem that the importer's node-sass is not
> equal to the node-sass that calls the importer?
>
> —
[...]

PR #1308, opened 2015-12-20
Handle sass.NULL returned from the importer via done()

saper:
Whenever done() is called we should check for the sass.NULL
value as well.

Fixes https://github.com/sass/node-sass/issues/1296


saper:
With little effort we could do this in the binding as well, it would probably be a better place.

xzyfer:
LGTM

xzyfer:
Superseded this with #1318 which just adds tests.

PR #1318, opened 2015-12-27
Handle sass.NULL returned from the importer via done()

xzyfer:
Whenever done() is called we should check for the sass.NULL
value as well.

While here, allow calling done(null) to anticipate
removal of the sass.NULL one day.

Closes https://github.com/sass/node-sass/pull/1308
Fixes https://github.com/sass/node-sass/issues/1296


xzyfer:
Since this doesn't introduce any BC breaks I'm :ship:

Issue #1212, opened 2015-10-22
CLI options to watch a directory


ysimonson:
Hello, I've been trying to figure out what the proper command line arguments are to watch a directory. I tried several combos - all of which seem to do keep the CLI running, but not doing anything (files don't compile, nothing in stdout/stderr.)

Based off #700 , I tried this: `node-sass -w -r --output static/stylesheets static/sass`

Based off #1075, I tried this: `node-sass -w web/static/sass -o web/static/stylesheets`

xzyfer:
Are you watching `.sass` or `.scss` files. There is a known issue with the watcher and `.sass` files currently.

xzyfer:
Possibly related https://github.com/sass/node-sass/issues/1179

ysimonson:
It's `.scss` files

xzyfer:
```
node-sass [options] <input> [output]
```
So the answer to your questions is
```
node-sass -w -r web/static/sass web/static/stylesheets
```
If this doesn't work please create a small, fully working, github repo that shows your problem.

ysimonson:
@xzyfer that gives me the error: `An output directory must be specified when compiling a directory` (it actually prints it out twice), even though web/static/stylesheets exists.

xzyfer:
This should be fixed by #1322.

xzyfer:
This is fixed on master and will be in the next release.

PR #1322, opened 2015-12-30
Allow symlinked output directories

nibblebot:
Should fix the concern the user has in #1212 where using a symlinked output directory incorrectly errors out saying `An output directory must be specified when compiling a directory`
/cc @ysimonson

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -588,6 +588,24 @@ describe('cli', function() {
       });
     });
 
+    it('should not error if output directory is a symlink', function(done) {
+      var outputDir = fixture('input-directory/css');
+      var src = fixture('input-directory/sass');
+      var symlink = fixture('symlinked-css');
+      fs.mkdirSync(outputDir);
+      fs.symlinkSync(outputDir, symlink);
+      var bin = spawn(cli, [src, '--output', symlink]);
+
+      bin.once('close', function() {
+        var files = fs.readdirSync(outputDir).sort();
+        assert.deepEqual(files, ['one.css', 'two.css', 'nested'].sort());
+        var nestedFiles = fs.readdirSync(path.join(outputDir, 'nested'));
+        assert.deepEqual(nestedFiles, ['three.css']);
+        rimraf.sync(outputDir);
+        fs.unlinkSync(symlink);
+        done();
+      });
+    });
   });
 
   describe('node-sass in.scss --output path/to/file/out.css', function() {
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -109,7 +109,7 @@ function isDirectory(filePath) {
   var isDir = false;
   try {
     var absolutePath = path.resolve(filePath);
-    isDir = fs.lstatSync(absolutePath).isDirectory();
+    isDir = fs.statSync(absolutePath).isDirectory();
   } catch (e) {
     isDir = e.code === 'ENOENT';
   }


test/cli.js
```diff
@@ -588,6 +588,23 @@ describe('cli', function() {
       });
     });
 
+    it('should not error if output directory is a symlink', function(done) {
+      if (process.platform === 'win32') {
+        return done(); // no symlink support on win32
```
xzyfer: @am11 is this true? I thought Windows had a symlink equivalent.

xzyfer:
Thanks @nibblebot! Are you able to squash your commits?

I'm not sure bailing out early for Windows is correct. I've asked @am11 (our resident Windows expert) to confirm.

test/cli.js
```diff
@@ -588,6 +588,23 @@ describe('cli', function() {
...
+      var src = fixture('input-directory/sass');
+      var dest = fixture('symlinked-output-directory');
+      var bin = spawn(cli, [src, '--output', dest]);
+
+      bin.once('close', function() {
+        var files = fs.readdirSync(dest).sort();
+        assert.deepEqual(files, ['.gitkeep', 'one.css', 'two.css', 'nested'].sort());
+        var nestedFiles = fs.readdirSync(path.join(dest, 'nested'));
+        assert.deepEqual(nestedFiles, ['three.css']);
+        rimraf.sync(dest+'/*');
```
nibblebot: removes everything except `.gitkeep`

test/cli.js
```diff
@@ -588,6 +588,23 @@ describe('cli', function() {
...
+      if (process.platform === 'win32') {
+        return done(); // no symlink support on win32
+      }
+      var src = fixture('input-directory/sass');
+      var dest = fixture('symlinked-output-directory');
+      var bin = spawn(cli, [src, '--output', dest]);
+
+      bin.once('close', function() {
+        var files = fs.readdirSync(dest).sort();
+        assert.deepEqual(files, ['.gitkeep', 'one.css', 'two.css', 'nested'].sort());
```
nibblebot: `.gitkeep` since git cannot checkin empty dirs

am11:
@nibblebot, thanks. Your fix in `bin/node-sass` looks good. AFAICT, there was a reason we used lstat, but I am not recalling it.

For the test, we do not store symlink within our git repos, because they are non-portable file system artifacts. Instead in the test case, create a symlink on the fly, work with it and then dispose it. You can delete all the fixtures you added in this PR. This would be a portable solution.

test/cli.js
am11: @xzyfer, Windows has two kinds of symlinks: soft, junction. Both are supported by node.js. The crude difference between Windows and Unix symlinks is instead of a flat file with a simple link, Windows has it as some encrypted hash in registry (a special folder like control panel etc.). Therefore they are not portable (can't zip them, cant git-package them and so on).

am11:
Thanks @nibblebot. We can reuse an existing fixture sub-directory instead of `input-directory2` and just create a symlink in test and dispose it on `close`.

am11:
LGTM. :tada: 

@xzyfer. PTAL.

saper:
Interesting: symbolic links might also to cause the issue also in #1323, where we call `fs.realpathSync()`  and `fs.lstatSync()` gives `ENOENT` back.

xzyfer:
Great work everyone!

Issue #1323, opened 2015-12-31
Installation failure on Linux when process.execPath is relative


bobbwal:
I've been through the troubleshooting but no node-sass/vendor folder is created so that's as far as I can go. Here's my output https://gist.github.com/bobbwal/f4ce61210ff88d9f9f85

Thanks

bobbwal:
Also the server is running Ash shell, not sure if this could be a problem. I noticed this comment in another issue:
"npm install executes a normal shell and tells it to run node scripts/install.js. If it can't find node (or this is some kind of funky restricted shell), things won't work."

saper:
`bash` is by no means mandatory; I am using FreeBSD's `/bin/sh` without any issues. The shell needs to understand `sh -c` convention and be able to find node interpreter using `PATH` environmental variable. What is your `SHELL` set to? Does `/bin/sh -c node` invoke your node interpreter?

bobbwal:
Hi, I'm guessing that command invokes the interpreter, no errors.
    
    $ echo $SHELL
    /bin/ash

    $ /bin/sh -c node
    >

abhijitghogre:
Facing this issue on OS X El Capitan!

saper:
@bobbwal is /bin/sh and /bin/ash the same binary? Does it help if you try bash? Is the current directory symlinked?

bobbwal:
/bin/sh and /bin/ash are identical but I'm not sure what this means. Bash is not installed but I can have a statically compiled bash added on request. The directory is symlinked `www -> /www`

saper:
This symlink is interesting - can you elaborate more on the directory structure used? Where is your  `node_modules` directory? Is this some well-known hosting environment (maybe there are public accounts to test and try to reproduce the issue)?

saper:
What is your `process.execPath` ?

saper:
Can you try running the following snippet under your `node`:
````js
var fs=require('fs'); console.log(process.execPath); console.log(fs.lstatSync(process.execPath)); fs.realpathSync(process.execPath);
````

bobbwal:
Here you go:

    > var fs=require('fs'); console.log(process.execPath); console.log(fs.lstatSync(process.execPath)); fs.realpathSync(process.execPath);
    node
    Error: ENOENT: no such file or directory, lstat 'node'
        at Error (native)
        at Object.fs.lstatSync (fs.js:839:18)
        at repl:1:69
[...]

PR #1325, opened 2016-01-03
Accept process.execPath as-is

saper:
Remove red tape surrounding process.execPath
and use it as-is. In particular, do not try
to canonicalize the path returned by
process.execPath as this might be relative
in the restricted Linux environments
(for example using overlayfs).

We should be prepared to accept relative
or otherwise broken process.execPath and
have a faith it works (otherwise our
build scripts wouldn't be invoked
in the first place).

https://github.com/sass/node-sass/issues/1323

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -26,27 +26,6 @@ function collectArguments(args) {
 }
 
 /**
- * Get Runtime Info
- *
- * @api private
- */
-
-function getRuntimeInfo() {
-  var execPath = fs.realpathSync(process.execPath); // resolve symbolic link
-
-  var runtime = execPath
-               .split(/[\\/]+/).pop()
-               .split('.').shift();
-
-  runtime = runtime === 'nodejs' ? 'node' : runtime;
-
-  return {
-    name: runtime,
-    execPath: execPath
-  };
-}
-
-/**
  * Get binary name.
  * If environment variable SASS_BINARY_NAME,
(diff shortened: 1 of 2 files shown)


saper:
After a little consideration this is not a bug in the node's fs module either; `process.execPath` should be considered existing relative to the OS-dependent `PATH` resolution mechanism and not relative to the current directory wherever our scripts may be executed. In other words, we try to `fs.lstatSync('node')` with a current director that does not have file named `node` and we correctly get `ENOENT` in return.

xzyfer:
Are you sure this isn't introducing a regression. I recall this being added for specific reason.

https://github.com/sass/node-sass/pull/668

saper:
I am never sure, I always check. Without https://github.com/libuv/libuv/pull/293 (or similar) merged we are just getting the working path to the executable - we do not here at all whether it is symlinked or pulled out of moon. Even relative path will work here as long as it is in some directory on `PATH`. (See even worse problem on OpenBSD https://github.com/libuv/libuv/pull/652). We just want to call "node", preferably the one we are called with. There is no superportable way to do this 100% correctly, but this approach will work. 

Next time I'll boot my Windows box I'll check of course.

Mischi:
:+1: 

I've verified that this PR would fix node-sass on OpenBSD if process.execPath is relative.

xzyfer:
On further thought this makes sense.

PR #1358, opened 2016-01-26
Replace "request" by "got"

zckrs:
* Less dependency
* Use only stream

I use [got@5](https://github.com/sindresorhus/got/tree/v5.x) because node-sass support NodeJS 0.10 and 0.12.

/cc @sindresorhus 

Graph deps compare [Request Graph](http://npm.anvaka.com/#/view/2d/request) vs [Got Graph](http://npm.anvaka.com/#/view/2d/got)

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,12 +55,12 @@
     "gaze": "^0.5.1",
     "get-stdin": "^4.0.1",
     "glob": "^6.0.2",
+    "got": "^5.4.1",
     "meow": "^3.3.0",
     "mkdirp": "^0.5.1",
     "nan": "^2.0.8",
     "node-gyp": "^3.0.1",
     "npmconf": "^2.1.2",
-    "request": "^2.61.0",
     "sass-graph": "^2.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


xzyfer:
I'm not thrilled about being pinned to an older version a library, although it does appear to still be receiving update. NodeJS < 4 support is not going away time soon. Our most used version is 0.12.

sindresorhus:
// @floatdrop

floatdrop:
@xzyfer `got` 5.x will receive updates until Node.JS 0.10 is deprecated.

lgtm.

xzyfer:
Thanks for the clarification @floatdrop

xzyfer:
We need to revert this because need to support proxies. I didn't realise proxies weren't supported at the time, my mistake.

zckrs:
Sorry about that but got doesn't support proxy @sindresorhus?

sindresorhus:
[Not yet built-in](https://github.com/sindresorhus/got/issues/79), but you can easily add it: https://github.com/sindresorhus/got#proxies

xzyfer:
tunnel current has [failing tests] for node 0.10 and 0.12 which are important to us. It may still work but this doesn't inspire confidence and users are currently experiencing issues.

PR #1360, opened 2016-01-29
Additional test for compressed output

wytesk133:


xzyfer:
Thanks @wytesk133, we appreciate your contribution but  node sass does not test the Sass language, output modes or sourcemaps. These are handled by LibSass' comprehensive test suite.

PR #1363, opened 2016-01-30
Fix CLI usage

qm3ster:
Updated CLI usage as per #669


am11:
LGTM. Thanks for catching this. :+1:

PR #1381, opened 2016-02-15
Custom importers cause out of order chained imports.

nowells:
I added a failing test case to demonstrate a bug in the current `importers`.

This error appears when you have custom `importers` defined, when a file has a chained import where the first import references a file not handled by the custom importer, and the second import is handled by the custom importer, the import order is broken.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -186,6 +186,29 @@ describe('api', function() {
   describe('.render(importer)', function() {
     var src = read(fixture('include-files/index.scss'), 'utf8');
 
+    it('should respect the order of chained imports when using custom importers and one file is custom imported and the other is not.', function(done) {
+      sass.render({
+        file: fixture('include-files/chained-imports-with-custom-importer.scss'),
+        importer: function(url, prev, done) {
+          if (url !== 'file-processed-by-loader') {
+            return sass.NULL;
+          }
+          done({
+            file: fixture('include-files/file-processed-by-loader.scss')
+          });
+        }
+      }, function(err, data) {
+        assert.equal(err, null);
+
+        assert.equal(
+          data.css.toString().trim(),
+          'body {\n  color: "red"; }'
+        );
+
+        done();
+      });
+    });
+
     it('should still call the next importer with the resolved prev path when the previous importer returned both a file and contents property - issue #1219', function(done) {
       sass.render({
         data: '@import "a";',
--- a/test/api.js
+++ b/test/api.js
@@ -190,11 +190,22 @@ describe('api', function() {
       sass.render({
         file: fixture('include-files/chained-imports-with-custom-importer.scss'),
         importer: function(url, prev, done) {
+          // NOTE: to see that this test failure is only due to the stated
+          // issue do each of the following and see that the tests pass.
+          //
+          //   a) add `return sass.NULL;` as the first line in this function to
+          //      cause non-custom importers to always be used.
+          //   b) comment out the conditional below to force our custom
+          //      importer to always be used.
+          //
+          //  You will notice that the tests pass when either all native, or
+          //  all custom importers are used, but not when a native + custom
+          //  import chain is used.
           if (url !== 'file-processed-by-loader') {
             return sass.NULL;
           }
           done({
-            file: fixture('include-files/file-processed-by-loader.scss')
+            file: fixture('include-files/' + url + '.scss')
           });
         }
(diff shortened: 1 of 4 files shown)


xzyfer:
Thanks for the test case @nowells. I've taken a quick look into it and it's certainly weird.

xzyfer:
After some digging around the issue appears to be related to peculiar `@import` syntax you're using.
Splitting it into multiple imports resolves the issue.

```
@import "file-not-processed-by-loader";
@import "file-processed-by-loader";
```

My guess is this is a bug in how LibSass handles this kind of multiple import syntax.

xzyfer:
With a little more digging it looks like LibSass is changing the order of imports.

Multiple imports
```
    .render(importer)
file-not-processed-by-loader
file-processed-by-loader
-------
/Users/michael/Projects/Sass/node-sass/test/fixtures/include-files/file-processed-by-loader.scss
####################################################################
Block 0x101501290 (0@[0:0]-[3:0]) [root] 0
 Import_Stub 0x101501f10 (0@[0:8]-[0:38]) [file-not-processed-by-loader.scss]  0
 Import_Stub 0x101008f20 (0@[1:8]-[1:34]) [/Users/michael/Projects/Sass/node-sass/test/fixtures/include-files/file-processed-by-loader.scss]  0
####################################################################
```

Single import
```
  api
    .render(importer)
file-not-processed-by-loader
file-processed-by-loader
-------
/Users/michael/Projects/Sass/node-sass/test/fixtures/include-files/file-processed-by-loader.scss
####################################################################
Block 0x10100c030 (0@[0:0]-[3:0]) [root] 0
 Import_Stub 0x101502900 (0@[2:40]-[2:66]) [/Users/michael/Projects/Sass/node-sass/test/fixtures/include-files/file-processed-by-loader.scss]  0
 Import_Stub 0x1015029e0 (0@[2:40]-[2:66]) [file-not-processed-by-loader.scss]  0
####################################################################
```

nowells:
@xzyfer the bug is certainly in libsass, when I apply this diff and build my tests pass (but 2 media query tests fail)

```diff
diff --git a/src/parser.cpp b/src/parser.cpp
index 8e18e3d..83f4230 100644
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -285,10 +285,12 @@ namespace Sass {
   Import* Parser::parse_import()
   {
     Import* imp = SASS_MEMORY_NEW(ctx.mem, Import, pstate);
-    std::vector<std::pair<std::string,Function_Call*>> to_import;
     bool first = true;
     do {
       while (lex< block_comment >());
+
+      std::vector<std::pair<std::string,Function_Call*>> to_import;
+
       if (lex< quoted_string >()) {
         if (!ctx.call_importers(unquote(std::string(lexed)), path, pstate, imp))
         {
@@ -322,6 +324,15 @@ namespace Sass {
         if (first) error("@import directive requires a url or quoted path", pstate);
         else error("expecting another url or quoted path in @import list", pstate);
       }
+
+      for(auto location : to_import) {
+        if (location.second) {
+          imp->urls().push_back(location.second);
+        } else {
+          ctx.import_url(imp, location.first, path);
+        }
+      }
+
       first = false;
     } while (lex_css< exactly<','> >());
 
@@ -330,14 +341,6 @@ namespace Sass {
       imp->media_queries(media_queries);
     }
 
-    for(auto location : to_import) {
-      if (location.second) {
-        imp->urls().push_back(location.second);
-      } else {
-        ctx.import_url(imp, location.first, path);
-      }
-    }
-
     return imp;
   }
 
```

I believe this is because we should really be having `if (!ctx.call_importers(unquote(std::string(lexed)), path, pstate, imp))` follow the same logic and add their imports to the `to_import` array which is processed at the bottom after full parsing and resolution of media_queries, but instead the `call_importers` directly adds to the `imp->urls()` as the imports are discovered.

xzyfer:
I came to the same conclusion last night. The correct patch is a little
more involved but not difficult. I'll push a PR up tonight.

nowells:
@xzyfer :heart: :fireworks: awesome! Thanks for spending the time to track this down! I look forward to seeing the proper fix so I can learn how to solve this properly myself next time around!

xzyfer:
LibSass PR https://github.com/sass/libsass/pull/1921

xzyfer:
This test should pass now that #1416 has been merged.

PR #1384, opened 2016-02-16
Remove vendor directory on prepublish and pack

### LINKED ISSUE #1183: how come vendor directory is ignored in package.json?
I'm using node-sass in a build project of sorts which gets packaged on a CI/CD server using `npm pack`. However, because the [`package.json` is ignoring it in the files array](https://github.com/sass/node-sass/blob/v3.3.3/package.json#L35-L41) `npm pack` isn't including it in the output tarball.
Our CI runs on the same architecture of the prod machine, so, it should just work if the binary were to be included.

Thanks!

gkatsev:
This is necessary because if a user is bundling node-sass in their
package, the vendor directory should stay so that the user isn't
required to run 'npm rebuild' afterwards, assuming that it was
originally installed on a like-architecture.
However, at the same time, when publishing, we don't want to include the
binaries since they may not work for a user plus there's the build step
that tries to download the correct binary for the current architecture.
Unfortunately, npm's prepublish hook runs on both prepublish *and* npm
install (see npm/npm#10074), so, using in-publish as a workaround to get
the prepublish script to only run prepublish and pack but not on
install.
This fixes #1183.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,14 +32,16 @@
     "postinstall": "node scripts/build.js",
     "pretest": "node_modules/.bin/jshint bin lib scripts test",
     "test": "node_modules/.bin/mocha test",
-    "build": "node scripts/build.js --force"
+    "build": "node scripts/build.js --force",
+    "prepublish": "not-in-install && node scripts/prepublish.js || in-install"
   },
   "files": [
     "bin",
     "binding.gyp",
     "lib",
     "scripts",
-    "src"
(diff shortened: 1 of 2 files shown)


gkatsev:
Any ideas why travis is red? The one red build seems to have exited with error code 0 (i.e., it passed).

gkatsev:
Maybe because `npm install` failed?

Issue #1198, opened 2015-10-06
Performance when using javascript custom functions


eoneill:
Javascript custom functions are awesome, and my current work wouldn't be possible without them, but we're noticing some performance impact when passing along large maps.

I threw together a minimal test case to demonstrate this: [eoneill/node-sass-perf-test](https://github.com/eoneill/node-sass-perf-test)

The test has two functions (`perftest-native` and `perftest-js`) that simply return the value passed to it:
```scss
@function perftest-native($value) {
  @return $value;
}
```

```js
"perftest-js($value)": function($value) {
  return $value;
}
```

Each method is passed a map with 5000 entries (a bit absurd, granted).

The finding is that the javascript method is exponentially slower than the native (Sass) method:

```
[native] elapsed time: 2ms
[js] elapsed time: 32ms
```

Lemme know if I can help provide any more information.

Thanks again for all the awesome work you guys do!

thiakil:
30ms difference doesn't seem that much? How does this affect in real-world applications?

A JavaScript function will likely never be as efficient thanks to the context switch from C++ -> JS -> C++ and the associated v8 conversions.

(I'm just a curious 3rd party)

xzyfer:
@thiakil true it's not a show stopper, but it's certainly worth being aware of. It's likely we can improve things here. 

I'm not entirely surprised there is degraded performance. For reference the conversion process is 

```
C++ -> C (libsass) -> C (node-sass) -> JS (v8) -|
                                                -> < your function >
                                                       |
                                                       |
C++ <- C (libsass) <- C (node-sass) <- JS (v8)       <-

```

eoneill:
Yeah, this isn't so much a complaint as an observation.

For a real world scenario, we use maps to store UI pattern definitions...

```scss
@include restyle-define(button, (
  color: blue,
  ...
));
```
We then retrieve the definitions via another mixin call:

```scss
.btn {
  @include restyle(button);
}
```
We started noticing that as we added more definitions, the compile times started getting exponentially slower, so we did some performance analysis and found the above observation.

We know there are some performance optimizations we can make on our end, but it's rather insignificant compared to the overhead we're seeing here.

For example, we measured the time it took to invoke our javascript call from sass in a realistic scenario and found:
```
Time in Javascript: 8.113895ms
Time in Sass: 21.282153ms
```
[...]

PR #1390, opened 2016-02-18
invoke custom functions with options.context

eoneill:
This allows custom functions to execute against the current render context. This is similar to the [current behavior of importers](https://github.com/sass/node-sass/blob/99dc18c448fcbb53b9e38d52eada3b39b3a57337/lib/index.js#L321).

This was discussed briefly in https://github.com/sass/node-sass/pull/783#discussion_r26910939

The benefit of this is that custom functions can persist data (be stateful), for the lifespan of the render.

This would allow you to do something like...

```js
{
  ...,
  functions: {
    "register-foo($foo)": function($foo) {
      this.myNamespace.$foo = $foo;
      ...
    },
    "get-foo()": function() {
      return this.myNamespace.$foo;
    }
  }
}
```
This would help alleviate some of the perf overhead we've seen with #1198. In our case, we currently store all stateful things in Sass variables and have to pass them around whenever we need them in JS.

NOTE: this is potentially a "breaking" change if anyone is currently using functions with an implicitly bound context.
```js
function Foo() {
  this.data;
}
Foo.prototype.doSomething = function() {
  return this.data;
};
var foo = new Foo();
{
  functions: {
    "doSomething()": foo.doSomething
  }
}
```
This would break as we are now binding `doSomething` to `options.context`, rather than `foo`.

The fix on the consumer side would be to explicitly bind to the context they need:
```js
{
  functions: {
    "doSomething": foo.doSomething.bind(foo)
  }
}
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -343,7 +343,7 @@ module.exports.render = function(options, cb) {
           bridge.success(data);
         }
 
-        var result = tryCallback(cb.callback, args.concat(done));
+        var result = tryCallback(cb.callback.bind(options.context), args.concat(done));
 
         if (result) {
           done(result);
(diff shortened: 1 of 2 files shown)


xzyfer:
Thanks @eoneill. Since this feature is marked experimental I'm happy to land this breaking change in the next minor (v3.5.0).

Issue #1333, opened 2016-01-07
npmconf is a deprecated and unmaintained dependency


yannickcr:
node-sass is using [npmconf](https://github.com/npm/npmconf) but this package is now deprecated and unmaintained.

Therefore some npmconf issue like npm/npmconf#61 makes node-sass impossible to install in some environments.

It would be great if the npmconf dependency could be removed/replaced.

xzyfer:
Querying what the alternative is now. npmconf is rather important to our internals and cannot simply be removed without an alternative.

xzyfer:
>run npm get proxy and npm get https-proxy as child processes. If that feels kind of clunky and heavyweight, I agree, but that's the most robust way to do it.
> https://github.com/npm/npmconf/pull/61#issuecomment-169523714

xzyfer:
As per https://github.com/npm/npmconf/pull/61#issuecomment-170880758

We'll like pull just the proxy code from npmconf into node-sass core.

adambuczynski:
Is this an issues that is going to be addressed still? I'm getting this warning for all my installs now:

```
npm WARN deprecated npmconf@2.1.2: this package has been reintegrated into npm and is now out of date with respect to npm
```

xzyfer:
We won't be changing this is the near future. Warnings can be ignored.

adambuczynski:
I know they can be ignored, but they ain't pretty. Is there no other way to do the thing that npmconf does? Or perhaps just take the lib and embed it within node-sass?

danez:
You could simply replace the npmconf call with Environment variables. All configurations from npm are added to `process.env` if run in the context of npm.

PhantomJS fixed it that way.
https://github.com/Medium/phantomjs/commit/2620fb283439181cab0a99b89fdf4532455bc35d#diff-f16acefe4b6553580c43edab685f50f3L168

xzyfer:
This patch was reverted in v3.5.2 because it was not functionally equivalent and broke installations for some proxy users.

jorrit:
In which cases did it not work?

xzyfer:
Not 100% sure but we got many reports of installations failing behind proxies.

https://github.com/sass/node-sass/issues/1458

xzyfer:
I noticed once case that wasn't counted for was the `.npmrc` config file. Could be related.

PR #1413, opened 2016-03-16
Bump almost all deps and remove npmconf

delitescere:
This updates almost all the dependencies and fixes issue #1333 

- Didn't bump `get-stdin` as v5.0.0 and v5.0.1 both changed the API and use Promises. Possibly look at the 'then/promise' module to shim.
- switched to `cross-spawn-async` as no spawning is done synchronously, so using more specific dependency.
- `graceful-fs` warning is due to mocha being tardy itself.

Tests pass on node v0.10, v0.12, v4.4.0. No additional tests created, however manual testing of `HTTP_PROXY=x npm ...` and `npm config set proxy x; npm ...` was done to verify the `npmconf` removal and change in `install.js` works.


xzyfer:
Thanks @delitescere. At first glance things look good. I'll take a closer look tomorrow.

PR #1427, opened 2016-03-25
Remove use of global process.sass

xzyfer:
This replaces the global `process.sass` with proper module exports.
Unfortunately `process.sass`, although undocumented, is available
to other modules. This patch keeps the existing public API available
until we bump the major.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -8,6 +8,7 @@ var Emitter = require('events').EventEmitter,
     util = require('util'),
     path = require('path'),
     glob = require('glob'),
+    sass = require('../lib'),
     render = require('../lib/render'),
     stdin = require('get-stdin'),
     fs = require('fs');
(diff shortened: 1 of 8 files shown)


xzyfer:
@am11 @andrew do either you have some context as to why `process.sass` was used in this way?

andrew:
No idea, I don't remember setting it up, any idea when it was first introduced?

xzyfer:
@andrew did some digging. Looks like @am11 added the first use of `process` in #660 and later namespaced using to `process.sass` in #743.

To the looks of those commits is was just a way to pass information around so it looks safe to remove. Please confirm @am11?

am11:
Sorry, couldn't sync up earlier as I was travelling the whole past week.
I think the rationale at the time was to create a sass namespace in persistent process object and use it elsewhere.

This new approach is not just pretty ok, it is excellent as the old method was analogous to the practice of frequently appending `document` object in browser apps, which is discouraged as altering the object layout at runtime is a bad idea (the same reason `delete`ing object property is discouraged in JavaScript).

:+1:

xzyfer:
Thanks for the background @am11. Hope you enjoyed your travels :)

PR #1428, opened 2016-03-25
Better error messages for missing binaries

xzyfer:
This is another iteration on improving the infamous
>The `libsass` binding was not found

Messages will now provide more useful information which will
- give users a chance to resolve the problem themselves
- give us more debug information from the error message alone

Error messages produced now be one of the following flavours:

>Node Sass does not yet support your current environment: OS X 64-bit with Node.js 4.x
>For more information on which environments are supported please see:
>http://....

****

>Error: Missing binding /tmp/node-sass/vendor/darwin-x64-47/binding.node
>Node Sass could not find a binding for your current environment: OS X 64-bit with Node.js 4.x
>
>Found bindings for the following environments:
>  - OS X 64-bit with io.js 3.x
>  - OS X 64-bit with Node.js 5.x
>
>This usually happens because your environment has changed since running `npm install`.
>Run `npm rebuild node-sass` to build the binding for your current environment.

****

>Error: Missing binding /tmp/node-sass/vendor/darwin-x64-47/binding.node
>Node Sass could not find a binding for your current environment: OS X 64-bit with Node.js 4.x
>
>This usually happens because your environment has changed since running `npm install`.
>Run `npm rebuild node-sass` to build the binding for your current environment.
*****

It's worth noting that we can't really do much better than the following if users are using custom binary names or paths. I'm ok with this since those users almost never file issues.
>Node Sass could not find a binding for your current environment: OS X 64-bit with Node.js 4.x

/cc @saper

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,0 +1,48 @@
+/*!
+ * node-sass: lib/errors.js
+ */
+
+var sass = require('./extensions');
+
+function humanEnvironment() {
+  return sass.getHumanEnvironment(sass.getBinaryName());
+}
+
+function foundBinaries() {
+  return [
+    'Found bindings for the following environments:',
+    foundBinariesList(),
+  ].join('\n');
+}
+
+function foundBinariesList() {
+  return sass.getInstalledBinaries().map(function(env) {
+    return '  - ' + sass.getHumanEnvironment(env);
+  }).join('\n');
+}
+
+function missingBinaryFooter() {
+  return [
+    'This usually happens because your environment has changed since running `npm install`.',
+    'Run `npm rebuild node-sass` to build the binding for your current environment.',
+  ].join('\n');
+}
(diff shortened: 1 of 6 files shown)


xzyfer:
We should keep the path we were looking at as well. I'll add that back in.

saper:
I like it a lot.
one thing that kept me wondering for a while: why are using a subdirectory for a platform stuff instead of xxx-yy-zzz-binding.node files?

xzyfer:
Good question. Maybe @am11 has some background?

xzyfer:
Updated the error messages to include the expected binary path when it's missing. I'll add some tests and :ship:

xzyfer:
@saper as suspected there is a reason for the having the directories - https://github.com/sass/node-sass/issues/712#issuecomment-77762212

PR #1437, opened 2016-04-06
CLI: Before watching files, render files

yuheiy:


xzyfer:
@yuheiy thanks for PR but this behaviour is on purpose and we won't be changing it. With change it would not be possible to integrate the node-sass smart watcher into task runners like gulp and grunt.

PR #1441, opened 2016-04-10
Bump LibSass 3.3.5

xzyfer:
~~This isn't the final LibSass 3.3.5 SHA1, but it's close.~~
~~I'll be using the branch in my CITGM.~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "node-sass",
   "version": "3.5.0-beta.1",
-  "libsass": "3.3.4",
+  "libsass": "3.3.5",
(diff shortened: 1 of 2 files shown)


xzyfer:
Looks like LibSass master has a regression between 3.3.4 and 3.3.5 Tracking it down now. /cc @mgreter

xzyfer:
The regression was introduced in https://github.com/sass/libsass/pull/1986 and is causing segfaults in sassdash.

xzyfer:
LibSass 3.3.5 has been tagged and the CITGM has passed.

PR #1478, opened 2016-04-24
Implement sass value operations and stringification

mgreter:
Brings recently added functions for [sass values from libsass C-API] [2] to node-sass. I guess it was partly due to missing libsass documentation, which was not updated in this regard (until some time earlier this day). Was a pitty that so far only perl-libsass implemented them and it was pretty straight forward to do so for node.js (very first time that I'm really doing something in node.js, beside being a day time js programmer).

Each and every sass value type get these methods via their protoype:
```js
// execute operator on this and rhs
// operator: string identifier for operator
// [eq|gt|lt|or|and|neq|gte|lte|add|sub|mul|div|mod]
// rhs: right hand side value for operation
// return a new sass value (bool, string, etc)
var result = val.operation(operator, rhs);

// overloaded operator
var string = val.toString();
```

I also hope that this may bring some value back to libsass in general, as these functions expose quite a low level API, as concatenation or equality tests are the same as libsass uses internally. One big open question is still how we gonna test the API on the libsass side. So far we rely on our implementors to add some unit test. So we pretty much need a full featured test suite for sass operation and stringification. Altough I would much better like to have this directly in libsass (so it is run with our CI), but node-sass seems a good place to get something started. Specially since @chriseppstein seems to use it himself. And I consider him the authority to know how these operations should behave. I am for the most part just reverse engineering the existing :smile: Anyway, that would be one long-term goal here.

On the other hand they are also usefull for end users. Unfortunately it looks that v8 has decided to not support any kind of custom operator overloading. The closest thing I could find in this direction is a [patch that looks to come from QT] [1]. Not a compelling sign, and all other sources point to "no, can't do". So the next best thing is probably to use methods (I mean the language even has java in its name, so ...).

I figured this could go two ways. Either have multiple methods for each operation: a bit more well defined but cluttering the prototype. Use only one method with switch argument: the string argument feels a bit quirky but works and also adds some versatility (can easly be stored/passed). The first idea could also be implemented around the second approach.

Also not 100% sure if a node.js library should overload `toString`. But I don't see why we shouldn't do that for our own custom objects, until you need it for some ubique equality test. Didn't bother to add any documentation up to this point as I first want to gather some feedback!

[1]: https://codereview.qt-project.org/#/c/6386/1/src/v8/0008-Add-custom-object-compare-callback.patch
[2]: https://github.com/sass/libsass/blob/master/docs/api-value.md

### DIFF
--- a/src/sass_types/sass_value_wrapper.h
+++ b/src/sass_types/sass_value_wrapper.h
@@ -25,6 +25,8 @@ namespace SassTypes
       static v8::Local<v8::Function> get_constructor();
       static v8::Local<v8::FunctionTemplate> get_constructor_template();
       static NAN_METHOD(New);
+      static NAN_METHOD(Stringify);
+      static NAN_METHOD(Operation);
       static Sass_Value *fail(const char *, Sass_Value **);
 
     protected:
@@ -73,6 +75,8 @@ namespace SassTypes
     v8::Local<v8::FunctionTemplate> tpl = Nan::New<v8::FunctionTemplate>(New);
     tpl->SetClassName(Nan::New<v8::String>(T::get_constructor_name()).ToLocalChecked());
     tpl->InstanceTemplate()->SetInternalFieldCount(1);
+    Nan::SetPrototypeMethod(tpl, "operation", Operation);
+    Nan::SetPrototypeMethod(tpl, "toString", Stringify);
     T::initPrototype(tpl);
 
     return scope.Escape(tpl);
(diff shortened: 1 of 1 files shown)


src/sass_types/sass_value_wrapper.h
```diff
@@ -88,6 +92,65 @@ namespace SassTypes
   }
 
   template <class T>
+  NAN_METHOD(SassValueWrapper<T>::Operation) {
+    if (info.Length() < 2) {
+      return Nan::ThrowTypeError("Not enough arguments for operation");
+    }
+    if (!info[0]->IsString()) {
+      return Nan::ThrowTypeError("Operation value should be a string");
```
mgreter: Tried to stay as close as possible to other error messages.
But I just couldn't help wondering, why `should` and not `must`?
To me it looks like a very strict constraint one `should` meet :wink:

src/sass_types/sass_value_wrapper.h
```diff
@@ -88,6 +92,65 @@ namespace SassTypes
...
+    }
+    if (SassTypes::Value *rhs = SassTypes::Factory::unwrap(info[1])) {
```
mgreter: unwrap doesn't seem to upgrade native type automatically?
I would have wished for it to handle i.e. native numbers transparently for me.
Or is there another utility function that does just that?

saper: `unwrap` generates whatever it can, currently only non-native artificial SassTypes. https://github.com/sass/node-sass/pull/1310 is there to introduce proper primitive JavaScript null, I have also a patch to introduce native `true` and `false`. 

Numbers are more difficult, since Sass number is actually a `(number, unit)` tuple and JavaScript does not support that.

mgreter:
@saper maybe I missed athe point, but it was about the input to `unwrap`. I want to put in `2` and get a sass number `2` with no unit. `unwrap(2)` => `SassNumber(2, ""`)

PR #1487, opened 2016-04-27
Add friendly name for the Node 6 release and fallback

nschonni:
Noticed that the message when the binding module version is unknown it currently says "if false"
Not sure how to to test this properly, I just eyeballed the patch in the IDE.


xzyfer:
Thanks @nschonni. I've made some improves to this in #1491.

PR #1488, opened 2016-04-27
Replace TODO URL with release tab for supported versions

nschonni:
This would assume that a quick version table gets added to the release page after attaching the bindings during the release. EX:

OS | Node
------------ | -------------
Windows x86 & x64 | 0.10, 0.12, 1, 2, 3, 4, 5
OSX x64 | 0.10, 0.12, 1, 2, 3, 4, 5
Linux x86 & x64 | 0.10, 0.12, 1, 2, 3, 4, 5

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -2,7 +2,8 @@
  * node-sass: lib/errors.js
  */
 
-var sass = require('./extensions');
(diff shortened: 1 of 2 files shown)


xzyfer:
This probably a decent MVP way to do this.

Issue #1458, opened 2016-04-21
Post Install fails inside a proxy


toninofox:
Dear all, 
I am trying to install node-sass inside a proxy properly configured in node npm via 
`npm config set proxy http://myproxy.com:8080`

all dependencies works but the post install of node-sass fails

 ```
npm info lifecycle node-sass@3.5.1~install: node-sass@3.5.1
 
  node-sass@3.5.1 install D:\...\node_modules\node-sass
  node scripts/install.js
 
 Cannot download "https://github.com/sass/node-sass/releases/download/v3.5.1/win32-ia32-47_binding.node": 
 
 getaddrinfo ENOTFOUND github.com github.com:443
 
 Hint: If github.com is not accessible in your location
       try setting a proxy via HTTP_PROXY, e.g. 
    export HTTP_PROXY=http://example.com:1234
 
 or configure npm proxy via
npm config set proxy http://example.com:8080


```

In the install.js script of the node-sass module, there is the proxy configuration reported below:

```
function applyProxy(options, cb) {
  var env = process.env;

  options.proxy = env.npm_config_https_proxy ||
                  env.npm_config_proxy ||
                  env.npm_config_http_proxy ||
                  env.HTTPS_PROXY ||
                  env.https_proxy ||
                  env.HTTP_PROXY ||
                  env.http_proxy;

  cb(options);
}
```

the older versions (that work inside the proxy) are configured with the function as below

```
function applyProxy(options, cb) {
  npmconf.load({}, function (er, conf) {
    var proxyUrl;

    if (!er) {
      proxyUrl = conf.get('https-proxy') ||
                 conf.get('proxy') ||
                 conf.get('http-proxy');
    }

    var env = process.env;

    options.proxy = proxyUrl ||
                    env.HTTPS_PROXY ||
                    env.https_proxy ||
                    env.HTTP_PROXY ||
                    env.http_proxy;

    cb(options);
  });
}
```

Thank You!

thedarkdestructor:
We are experiencig the same exact issue!

shouldn't the search for a proxy rely on npm config properties instead of env variables?

A clarification will be highly appreciated.

basvandenheuvel:
Same issue here.

catchamonkey:
It worked before, just by using the env vars (which is nicer IMO).

basvandenheuvel:
@xzyfer version 3.5.1 introduced proxy problems. Solution?

xzyfer:
[...]

PR #1492, opened 2016-04-27
Replace deprecated npmconf package.

### LINKED ISSUE #1333: npmconf is a deprecated and unmaintained dependency

xzyfer:
This was previously attempted in #1413. Shortly after it's
release proxy users started experiencing installation issues (#1458) 
so this was reverted. It was later determined that #1458 was likely
at fault for the proxy issues.

Full credit for this patch goes to @delitescere.

I've also taken the liberty of cleaning the request config generation.

Fixes #1333


nschonni:
LGTM

xzyfer:
I'll release this on a the `next` dist-tag and ask some proxy users to try
it out. It pays to be cautious with proxy changes.

xzyfer:
YOLO

PR #1486, opened 2016-04-27
Update CI builds for NodeJS 6

nschonni:
### Blocked

- [ ] appveyor/ci#770

PR #1499, opened 2016-04-28
Add Node 6 as a supported runtime

xzyfer:


xzyfer:
Cancelled this builds. This is a noop as far as our tests are concerned until #1486 lands. It's currently blocked by appveyor/ci#770.

Going to ship this since #1486 will fail without it.

PR #1500, opened 2016-04-28
Simplify Node versions in Travis Ci

xzyfer:
Previous we manually installed NVM and did the version switching
ourselves. This was required because Travis didn't support NVM
on OSX builds. This has since [been fixed][1] so lets see how it
goes.

[1]: https://github.com/travis-ci/travis-ci/issues/2311

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
-language: cpp
+language: node_js
+node_js:
+  - "5"
+  - "4"
+  - "0.12"
+  - "0.10"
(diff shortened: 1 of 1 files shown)


xzyfer:
I'm somewhat amazed this worked although builds appear to be slower. I'll rebuild later when the US is asleep and Travis' load is lower.

xzyfer:
It's still slower but not painfully slow. It's worth for the reduction in complexity.

PR #1506, opened 2016-04-29
README: Add Node.js 6 to supported versions.

coderbyheart:
The support was added in sass/node-sass#1499

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # node-sass
 
-#### Supported Node.js versions 0.10, 0.12, 1, 2, 3, 4, and 5.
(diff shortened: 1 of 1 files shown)


xzyfer:
Thank you. I'm putting this off until we ship 3.7.0

coderbyheart:
Of course, just wanted to make sure that the README is accurate.

Zorgatone:
Is Node 6 supported??

I still get the `Error: Node Sass does not yet support your current environment: Windows 64-bit with false` error

xzyfer:
It's not, which is why this hasn't been merged.

Zorgatone:
Oh, ok. :+1:

PR #1510, opened 2016-04-30
[trash]

xzyfer:
Debugging Windows Node 6 CI failures from #1486.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,7 @@
 language: node_js
 node_js:
+  - "6"
   - "5"
-  - "4"
-  - "0.12"
-  - "0.10"
-  - "iojs"
 
 compiler: gcc
 sudo: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -90,7 +90,9 @@
   clone_folder: c:\projects\node_modules\node-sass
 
   init:
-    - ps: set-location -path c:\projects\node_modules\node-sass
+    - cmd: >-
+        subst s: c:\projects
+    - ps: set-location -path s:\node_modules\node-sass
 
   cache:
     - '%userprofile%\.node-gyp'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -101,9 +101,7 @@
   environment:
     SKIP_SASS_BINARY_DOWNLOAD_FOR_CI: true
     matrix:
-      - nodejs_version: 0.10
-      - nodejs_version: 0.12
-      - nodejs_version: 4
+      - nodejs_version: 6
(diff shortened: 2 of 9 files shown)


Keeo:
You can do it @xzyfer!

saper:
if sass.NULL is the problem, can you print the actual value of the sass.NULL pointer? if we get multiple sass.NULL instances then we have a problem.

xzyfer:
`sass.NULL` does not appear to be the issue.

xzyfer:
@saper I have managed to create a reduced test case. I feel like it related to https://github.com/nodejs/node/issues/6160 or some other regressions around `require`s cache introduced in 6.0.0.

So far I have been able to assert the issue only affects Windows (both 32 and 64 bit). See identical builds on [Linux][1], [OSX][2], and [Windows][3].

I suspect the `require` cache because my reduced test case `require`s our native modules twice.
- on Windows w/ Node 6 it's loaded at two different paths
- on Windows w/ Node 5 it's loaded twice at the same path
- on Linux w/ Node 6 it's loaded twice at the same path
- on OSX w/ Node 6 it's loaded twice at the same path

The error only happens in the first scenario.

[1]: https://travis-ci.org/sass/node-sass/jobs/126892448
[2]: https://travis-ci.org/sass/node-sass/jobs/126892449
[3]: https://ci.appveyor.com/project/sass/node-sass/build/job/5mhcfgpe8txcdyif

xzyfer:
It looks like I was correct about the require cache. Removing the `subset` in AppVeyor resolved the issue. This is because both the native modules requires are now both

```
C:\projects\node_modules\node-sass\vendor\win32-ia32-48\binding.nod
```

Where as with the `subset` the native module was required first from

```
C:\projects\node_modules\node-sass\vendor\win32-ia32-48\binding.nod
```
then secondly from
```
s:\node_modules\node-sass\vendor\win32-ia32-48\binding.nod
```

xzyfer:
Ok I've narrowed this down a bug with `require` and `subst`.

```
mkdir \temp1
cd \temp1
npm install node-sass
subst s: \temp1
node
>require('\temp1\node-sass\vendor\win32-ia32-48\binding.node');
>require('\s\vendor\win32-ia32-48\binding.node');
```

@saper when you're online can you please confirm this?

xzyfer:
Confirmed removing the subst get the tests passing in https://github.com/sass/node-sass/pull/1512

xzyfer:
I'm tempted to release v3.7.0 with MSVC PDB files and retroactively compile them once this is fixed in Node.

PR #1512, opened 2016-04-30
[trash] Assert Node 6 tests pass without subst

xzyfer:
After lots of investigation in #1510 I've determine the Node 6
failure we're seeing are due to the native extension being required
on different paths.


xzyfer:
Great. This proves we can compile file on Windows with Node 6.

PR #1514, opened 2016-05-01
Node 6 for Windows: Don't use subst for testing

saper:
Node 6.x has trouble probably including
modules from the substed drivers.
Since we use use subst only to set
symbols paths in the PDB file, let's
revert to the C: drive when doing tests.

Workaround for: https://github.com/nodejs/node/issues/6500

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,6 +42,7 @@
       - nodejs_version: 3
       - nodejs_version: 4
       - nodejs_version: 5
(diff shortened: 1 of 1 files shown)


xzyfer:
Did you miss the non-release config?

saper:
Of course I have. Thanks!

saper:
:dancers:

xzyfer:
Thank you kindly :tada:

PR #1518, opened 2016-05-02
Chore: Don't run partial build for tags or release

nschonni:
From guidance in http://help.appveyor.com/discussions/questions/261-only-deploy-when-theres-a-new-tag#comment_39426826

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,8 @@
     except:
       - master
 
+  skip_non_tags: true
+
   os: Visual Studio 2013
 
   configuration: release
@@ -79,6 +81,12 @@
         appveyor_repo_tag: true       # deploy on tag push only
 
 -
+  branches:
+    except:
+      - release
+
+  skip_non_tags: false
+
   os: Visual Studio 2013
 
   configuration: testing


nschonni:
Hmm, this doesn't seem like it likes it since there is no AppVeyor build triggered :(

appveyor.yml
```diff
@@ -79,6 +79,12 @@
         appveyor_repo_tag: true       # deploy on tag push only
 
 -
+  branches:
+    except:
+      - release
+
+  skip_non_tags: true
```
xzyfer: You'd want to skip tags here, not non-tags.

xzyfer:
You're only building tags now which is the inverse desired behaviour.

appveyor.yml
```diff
@@ -79,6 +81,12 @@
         appveyor_repo_tag: true       # deploy on tag push only
 
 -
+  branches:
+    except:
+      - release
+
+  skip_non_tags: false
```
xzyfer: Is this redundant?

xzyfer: Is there way to explicitly tell AppVeyor not to build tags?

nschonni: I added the `true` section in the tag config above, but maybe that's still not right

xzyfer: I would think you only need here if the configs cascade which I don't believe they do. Try removing this and see if AppVeyor still builds the branch.

xzyfer:
FWIW you should work on a fork otherwise you trigger a lot of builds.

PR #1519, opened 2016-05-02
Attempt to fix AppVeyor tag race condition

xzyfer:
Currently it's a bit of an ordeal trying to get the atomic puush
required for AppVeyor to register a commit on the release branch
and also a tag. Looks like AppVeyor expects you to do release from
the master branch.

Inspired by @nschonni work in #1518 in an attempt to work around
this issue this PR instructs AppVeyor to ignore tags on non-release
branches.

Fingers crossed this makes like easier.

Closes #1518

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -79,6 +79,13 @@
         appveyor_repo_tag: true       # deploy on tag push only
 
 -
+  branches:
+    except:
+      - release
+      - /v\d\.\d\.\d/
+
+  skip_tags: true
+
   os: Visual Studio 2013
 
   configuration: testing


xzyfer:
It was relatively less painful, but still not straight forward.

saper:
It is actually not @appveyor's fault. Github invokes webhooks many times for one commit in a funny combination. I don't have the advice from the AppVeyor forum at hand, but there I think there was a way to adjust github webooks so that only the stuff we need is sent.

xzyfer:
That may be the case this release was significantly easier than in the past. This patch may not help but it doesn't hurt either.

PR #1520, opened 2016-05-02
Don't run tests when building a Windows release

xzyfer:
This slows down the release builds and in the worst case scenario
causes a single job failing meaning we need to rerun the entire build.

Problems with tests will have been caught we'll before we start
building release binaries.

Thoughts @saper?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -51,9 +51,7 @@
     - git submodule update --init --recursive
     - npm install --msvs_version=2013
 
-  test_script:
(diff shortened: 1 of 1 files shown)


saper:
Not sure it's a good idea. `master` is the only place where all PRs really meet together. Apart from the (very rare) direct commits to master there is always a possibility of a bad merge. Plus only then we test all platforms.

xzyfer:
> master is the only place where all PRs really meet together

We still run tests on master. The intent here is to skip tests on the release branch.

PR #1528, opened 2016-05-05
Docs: Add jsdocs for human readable public methods

nschonni:
Renamed parameters to be a little clearer, and marked them public since
they are being exported

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -8,8 +8,17 @@ var eol = require('os').EOL,
     path = require('path'),
     defaultBinaryPath = path.join(__dirname, '..', 'vendor');
 
-function getHumanPlatform(arg) {
-  switch (arg || process.platform) {
+/**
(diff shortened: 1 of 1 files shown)


lib/extensions.js
```diff
@@ -18,17 +27,37 @@ function getHumanPlatform(arg) {
...
 }
 
-function getHumanArchitecture(arg) {
-  switch (arg || process.arch) {
+/**
+ * Provides a more readable version of the architecture
+ *
+ * @param  {string} arch - An instruction architecture name to match, or null to
+ * lookup the current process architection
+ * proces.arch value
```
xzyfer: typo?

xzyfer:
lgtm aside from my comment

lib/extensions.js
nschonni: Yeah, looks like I proofed poorly. Updated one or too more and repushed

PR #1532, opened 2016-05-06
docs: split up into separate files like lib sass

nschonni:
# DO NOT MERGE

Took a quick crack at splitting up the docs in the same way as libsass. Some link aliases will be broken right not, and a README.md would be needed for the folder.


xzyfer:
This has been on my to-do list for ages. Thank you for starting this. One
thing, I would drop the integrations. It's not useful and probably wildly
out of date.

PR #1539, opened 2016-05-10
Update watched file list when files are added or removed

### LINKED ISSUE #1538: CLI watcher's import graph is not updated when files are added or deleted
Originally mentioned in https://github.com/sass/node-sass/issues/1095#issuecomment-176529490. This has been a known issue for a long time but has only recently come up.

### Background

Most watchers will niavely recompile all watched files when a watched file is modified. This is the case with tools like gulp and grunt.

### Smart watching

The Node Sass watcher is a bit smarted. When the CLI watcher is started up it builds graph of Sass imports for the files it's watching. With this graph it can determine which files to recompile when a partial is changed. This is sometimes referred to as incremental compilation.

### The bug

Currently this graph is only generated when the CLI watcher is started. This means the grpah isn't updated when files are added or deleted. This latter isn't much of problem, but the former results in new files not triggering recompiles.

The only way to register these files in the graph is to restarted the watcher.

xzyfer:
Currently we build the Sass import graph when the CLI watcher is
started. However the graph is not update updated when files are
added or deleted. The latest is a big deal but the former results
in new files not triggered rebuilds when they're changed. The only
way to currently resolve this is to restart the CLI watcher.

This patch rebuilds the Sass import graph when files are added or
deleted to the watch always works as expected.

Fixes #1538

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -230,15 +230,23 @@ function getOptions(args, options) {
  */
 
 function watch(options, emitter) {
-  var watch = [];
+  var buildGraph = function(options) {
+    var graphOptions = {
+      loadPaths: options.includePath,
+      extensions: ['scss', 'sass', 'css']
+    };
 
-  var graphOptions = { loadPaths: options.includePath, extensions: ['scss', 'sass', 'css'] };
-  var graph;
-  if (options.directory) {
-    graph = grapher.parseDir(options.directory, graphOptions);
-  } else {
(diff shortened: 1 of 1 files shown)


xzyfer:
AppVeyor failure is a false alarm.

xzyfer:
Unfortunately we don't have watcher tests atm because they were flakey in CI. Gonna test it in production 🙊

PR #1545, opened 2016-05-12
ESLint conversion from JSHint

nschonni:
Used Polyjuice to convert over the existing config, and added in some of the failures for the existing files.
There is a bunch of whitespace changes, but this makes it match the actual editorConfig settings (use the `&w=1` in the query string when reviewing).

Found the coverage has been broken since the build matrix was changed over, so I tacked that onto this. I'll pull it to a separate PR if this doesn't go ahead.


nschonni:
Coverage is (silently) falling down, but I'm still trying to figure out if that's me or something else

xzyfer:
I looked at this a while ago. My plan was to adopt JSCS and use their
airbnb preset because I require almost no changes. JSCS will soon be part
of ESLint so I guess this makes sense iif we use an existing preset. I very
much dislike managing a config file full of rules.

I'm curious to know what value you see in this change?

JSHint is super stable and never caused us any issues. ESLint on the other
hand has constant churn and frequently changes their rules which makes
upgrading non-trivial.

xzyfer:
I'm not saying we shouldn't do this, I just think we should have a good reason.

nschonni:
Since looking at ESLint again for other projects I've found that stuff that was passing with JSHint would actually still have errors that ESLint would catch with the same (converted) ruleset. The other part is like you said about the combining of the JSCS functionality.

This is just me coding at night to help turn off my brain, so I'm not too worried if this stuff gets closed.

.eslintrc
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "globals": {},
+  "rules": {
+    "no-bitwise": 2,
+    "camelcase": 0,
+    "curly": 2,
+    "eqeqeq": 2,
+    "no-unused-expressions": [
+      2,
+      "allowTernary": true
```
pdehaan: Is this kosher, or should it be an object? http://jsonlint.com/ fails pretty hard here.

```js
    "no-unused-expressions": [
      2,
      {
        "allowTernary": true
      }
    ],
```

package.json
```diff
@@ -30,7 +30,7 @@
     "coverage": "node scripts/coverage.js",
     "install": "node scripts/install.js",
     "postinstall": "node scripts/build.js",
-    "pretest": "node_modules/.bin/jshint bin lib scripts test",
+    "pretest": "node_modules/.bin/eslint bin lib scripts test",
```
pdehaan: FWIW, npm should add the "node_modules/.bin/" to the `$PATH` automatically, so you should only need `"eslint bin lib scripts test"` here.

.eslintrc
nschonni: The extensionless version isn't a JSON object, but looking at the docs it looks like this format is deprecated now http://eslint.org/docs/user-guide/configuring#configuration-file-formats

nschonni: Ignore me, it is still supposed to be a valid JSON object, so this does need to be wrapped

nschonni:
Rebased and adde in the `node-sass` bin properly now

xzyfer:
No objections from me

nschonni:
Landing.
Note the 3.x of ESLint dropped support for Node < 4, so till we do the same we probably shouldn't upgrade it.

xzyfer:
Runtime support of dev dependencies doesn't affect end users. I'm ok with expecting development to have different runtime requirements.

saper:
Question: suppose it no longer works with old node engines we support, would that negatively impact our ability to run `npm test` (for example with Travis CI)?

xzyfer:
Ah true

xzyfer:
We could separate linting from testing. No real need to lint multiple times.

saper:
Good idea!

nschonni:
I'll rename the npm run script to lint, then move it to the build matrix with the coveralls stuff so it only gets run once

PR #1563, opened 2016-05-24
Replace deprecated cross-spawn-async with cross-spawn

danez:
Both packages have been merged and cross-spawn is now
doing the same as cross-spawn-async before.


xzyfer:
Thanks @danez

mebyz:
+1 ;)

PR #1566, opened 2016-05-26
Add cache for downloaded binary files

### LINKED ISSUE #1554: Use node-sass offline with npm 3.9.0
Hi, I just updated npm to 3.9.0 and using `npm install --cache-min=Infinity` which now works like an offline mode, but I got a problem with node-sass.

### Our setup:
We have a huge java/maven based workspace with some frontend stuff. We use the [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin) to call node/npm and grunt. This worksapce is build on Linux, MacOSX and Windows. For the frontend we use grunt-sass, which based on node-sass. Every `mvn clean install` will delete the `node_modules` folder and redownload all dependencies needed. This is important for our CI, because its the same way our customer are using the workspace. This works nice as long as you are online.

### Problem:
Some of our consultants and customers want/need to work offline. We couldn't support it yet, because of the problems with the npm cache. Since release 3.9.0 fixed this problem, it works offline with all our modules (if you at least build it once online = cached), but not with node-sass.

### Possible Solution:
1. Put the binaries into the bundle: not an option, like you said in https://github.com/sass/node-sass/issues/654#issuecomment-73477480 to save bandwidth.
2. Configure SASS_BINARY_PATH: not an option for us, because we have to put the binaries into our source workspace and need to configure it. So we have to maintain the binaries for all systems and versions... Sounds like an antipattern.
3. **Download and keep the binaries in a temp folder. Check this location, before redownloading them. This is how [phantomjs ](https://github.com/Medium/phantomjs)is doing it with their binaries. *My favorite solution*. This saves bandwidth and time AND would work offline too!**

Here is an extract of the npm log to show how phantomjs/node-sass are doing it:

phantomJS
```
[INFO] > phantomjs-prebuilt@2.1.7 install C:\workspace\basic-theme\node_modules\phantomjs-prebuilt
[INFO] > node install.js
[INFO]
[INFO] Considering PhantomJS found at C:\Program Files (x86)\phantomjs\bin\phantomjs.EXE
[INFO] Found PhantomJS at C:\Program Files (x86)\phantomjs\bin\phantomjs.EXE ...verifying
[INFO] PhantomJS detected, but wrong version 2.0.0 @ C:\Program Files (x86)\phantomjs\bin\phantomjs.EXE.
[INFO] Download already available at C:\Users\mriehema\AppData\Local\Temp\phantomjs\phantomjs-2.1.1-windows.zip
[INFO] Verified checksum of previously downloaded file
[INFO] Extracting zip contents
[INFO] Removing C:\workspace\basic-theme\node_modules\phantomjs-prebuilt\lib\phantom
[INFO] Copying extracted folder C:\Users\mriehema\AppData\Local\Temp\phantomjs\phantomjs-2.1.1-windows.zip-extract-1463578294927\phantomjs-2.1.1-windows -> C:\workspace\basic-theme\node_modules\phantomjs-prebuilt\lib\phantom
[INFO] Writing location.js file
[INFO] Done. Phantomjs binary available at C:\workspace\basic-theme\node_modules\phantomjs-prebuilt\lib\phantom\bin\phantomjs.exe
```

node-sass:
```
[INFO] > node-sass@3.7.0 install C:\workspace\basic-theme\node_modules\node-sass

[INFO] > node scripts/install.js
[INFO]
[ERROR] Cannot download "https://github.com/sass/node-sass/releases/download/v3.7.0/win32-x64-46_binding.node":
[ERROR]
[ERROR] getaddrinfo ENOENT github.com:443
[ERROR]
[ERROR] Hint: If github.com is not accessible in your location
[ERROR]       try setting a proxy via HTTP_PROXY, e.g.
[ERROR]
[ERROR]       export HTTP_PROXY=http://example.com:1234
[ERROR]
[ERROR] or configure npm proxy via
[ERROR]
[ERROR]       npm config set proxy http://example.com:8080
```

So it would be a very cool feature, if node-sass could support cached/temp binaries to save bandwidth and to support an offline mode! Thx.

mriehema:
- The binary file is downloaded to a tmp folder and copied to vendor destination. If the binary already exist in the tmp folder, it will be copied only.
- With this change node-sass will work offline, cache the binary and save bandwidth.
- The binary configuration parameters will work the same as before.

Fixes #1554.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -142,7 +142,7 @@ function getTempPath(binaryName) {
   ];
 
   for (var i = 0; i < candidateTmpDirs.length; i++) {
-    var candidatePath = path.join(candidateTmpDirs[i], 'node-sass', "v" + pkg.version);
+    var candidatePath = path.join(candidateTmpDirs[i], 'node-sass', 'releases', 'download', 'v' + pkg.version);
 
     try {
       mkdir.sync(candidatePath);
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -142,7 +142,7 @@ function getTempPath(binaryName) {
   ];
 
   for (var i = 0; i < candidateTmpDirs.length; i++) {
-    var candidatePath = path.join(candidateTmpDirs[i], 'node-sass');
+    var candidatePath = path.join(candidateTmpDirs[i], 'node-sass', "v" + pkg.version);
 
     try {
       mkdir.sync(candidatePath);
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -3,6 +3,7 @@
  */
 
 var fs = require('fs'),
+    fsextra = require('fs-extra'),
     eol = require('os').EOL,
     mkdir = require('mkdirp'),
     path = require('path'),
@@ -34,7 +35,7 @@ function download(url, dest, cb) {
     return response.statusCode >= 200 && response.statusCode < 300;
   };
 
-  var options = { 
+  var options = {
     rejectUnauthorized: false,
     proxy: getProxy(),
     headers: {
@@ -109,15 +110,65 @@ function checkAndDownloadBinary() {
       return;
     }
 
-    download(sass.getBinaryUrl(), sass.getBinaryPath(), function(err) {
-      if (err) {
-        console.error(err);
-        return;
-      }
+    var tmpPath = getTempPath(sass.getBinaryName());
+    if (!(sass.hasBinary(tmpPath))) { // download and install
+      download(sass.getBinaryUrl(), tmpPath, function(err) {
+        if (err) {
+          console.error(err);
+          return;
+        }
+        console.log('Binary downloaded at', tmpPath);
+        copyBinary(tmpPath, sass.getBinaryPath());
+      });
 
-      console.log('Binary downloaded and installed at', sass.getBinaryPath());
-    });
+    } else { // install only
+      copyBinary(tmpPath, sass.getBinaryPath());
+    }
   });
+}
+
+/**
+ * Find a temp folder for file
+ *
+ * @param {String} binaryName
+ * @returns {string}
+ * @api private
+ */
+
+function getTempPath(binaryName) {
+  var candidateTmpDirs = [
+    process.env.TMPDIR || process.env.TEMP || process.env.npm_config_tmp,
+    '/tmp',
+    path.join(process.cwd(), 'tmp')
+  ];
+
+  for (var i = 0; i < candidateTmpDirs.length; i++) {
+    var candidatePath = path.join(candidateTmpDirs[i], 'node-sass');
+
+    try {
+      fsextra.mkdirsSync(candidatePath, '0777');
+      return path.join(candidatePath, binaryName);
+    } catch (err) {
+      console.error(candidatePath, 'is not writable:', err.message);
+    }
+  }
+}
+
+/**
+ * Copy file
+ *
+ * @param tmp
+ * @param dest
+ * @api private
+ */
+function copyBinary(tmp, dest) {
+  try {
+    fs.createReadStream(tmp).pipe(fs.createWriteStream(dest));
+
+    console.log('Binary installed at', sass.getBinaryPath());
+  } catch (err) {
+    console.err('Cannot install binary', err.message);
+  }
 }
 
 /**
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -5,6 +5,7 @@
 var fs = require('fs'),
     eol = require('os').EOL,
     mkdir = require('mkdirp'),
+    tmpdir = require('os-tmpdir'),
     path = require('path'),
     sass = require('../lib/extensions'),
     request = require('request'),
@@ -109,7 +110,7 @@ function checkAndDownloadBinary() {
       return;
     }
 
-    var tmpPath = getTempPath(sass.getBinaryName());
+    var tmpPath = getTempPath();
     if (!(sass.hasBinary(tmpPath))) { // download and install
       download(sass.getBinaryUrl(), tmpPath, function(err) {
         if (err) {
@@ -117,11 +118,14 @@ function checkAndDownloadBinary() {
           return;
         }
         console.log('Binary downloaded at', tmpPath);
-        copyBinary(tmpPath, sass.getBinaryPath());
+        //copy only, if not already in vendor
+        if (tmpPath !== sass.getBinaryPath()) {
+          copyBinary(tmpPath);
+        }
       });
 
     } else { // install only
-      copyBinary(tmpPath, sass.getBinaryPath());
+      copyBinary(tmpPath);
     }
   });
 }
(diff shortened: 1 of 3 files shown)


xzyfer:
Thanks for this @mriehema. Generally speak it looks good. I'll take a closer look over the weekend.

One thing, I don't like the addition of fs-extras just for `mkdir -p`. I'd prefer you changed this to use [mkdirp](https://www.npmjs.com/package/mkdirp) instead.

xzyfer:
Thanks for that update @mriehema.

@saper @nschonni I'd love for y'all to be over this as well.

nschonni:
The code looks OK, but can't this just be achieved as-is by passing the temp directory (or the expression to test for what temp directory) to the existing [SASS_BINARY_PATH](https://github.com/sass/node-sass/blob/ad4c44ce51500643cc951c2cf8891df97689d56e/lib/extensions.js#L243)

xzyfer:
@nschonni there was a bunch of prior discussion in https://github.com/sass/node-sass/issues/1554. This is a comprise of sorts.

nschonni:
I did read it, but apparently totally missed 2. 
This will cause even more problems for when people change node versions:
1. Install node-sass 3.7, binary gets copied to temp
1. Upgrade node to 3.8, and rerun install. Copies over 3.7 binary from temp again
1. Clear all node_modules like our usual troubleshooting, still grabs node 3.7 binary from temp

I think a general extension point for people wanting to change the download behaviour might be something to consider. 
If this one is going to proceed, it'll need some other folder layer to indicate node-sass version or at least version testing like PhantomJS uses to see if it can use the version in temp path.

xzyfer:
@nschonni I'm not sure I'm following. The intent of this change is just changing the download directory from `vendor` to something out side of `node_modules` i.e. `/tmp`. The copying the downloaded binary to vendor. The idea being if you `rm -rf node_modules` subsequent install will hit disk before the network.

We will not be changing the directory structure at all so the version switching issue is no different than it is today. The thing that does changes is that `rm -rf node_modules` is not guaranteed to fetch the latest binary, but since binaries are immutable this is a non-issue.

I haven't had time confirmed the implementation here respects those constraints but it wont ship with out it.

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
       return;
     }
 
-    download(sass.getBinaryUrl(), sass.getBinaryPath(), function(err) {
-      if (err) {
-        console.error(err);
-        return;
-      }
+    var tmpPath = getTempPath(sass.getBinaryName());
+    if (!(sass.hasBinary(tmpPath))) { // download and install
```
nschonni: @xzyfer this is the part I'm saying will always fall back to the first installed binary file in the temp folder

xzyfer: As I said I've haven't look over the implementation yet. It's entirely likely there are issues that will need to be addressed.

mriehema: The name of the binary file in the tmp folder based on `sass.getBinaryName()`. E.g. `win32-x64-47_binding.node`. If you use another version of node-sass, the name will change too.

mriehema:
> The idea being if you rm -rf node_modules subsequent install will hit disk before the network.

Exactly.

> We will not be changing the directory structure at all so the version switching issue is no different than it is today. 

Correct. I kept everything the same, but split download and install the binary into two different steps and added a "temp layer".

scripts/install.js
xzyfer: You're both correct. The binaries won't class on different node/arch/os
combinations. However because we're no longer within node_modules changing
versions of node-sass would be an issue. When changing versions of
node-sass the installer would see a binary for their environment on disk
but it would not the correct binary for their node-sass version.

You'll need to create a folder for the node-sass version for this to work
as intended.

saper:
I think it goes towards a pretty good think. Can we have the cache format the same as for node-sass-binaries repository? In short, checking it out from git somewhere should be the easiest way to fill the cache.

xzyfer:
With this approach the cached binary filenames will match the ones in
node-sass-binaries. I don't think adding an install time dependency on git
is a good idea. Git is notoriously difficult to integrate with and Windows
users are unlikely to have it installed.

I think this approach is solid as long as the binaries are downloaded into
a node-sass version specific temp directory.

mriehema:
I added the package version as subfolder to the temp path. It looks now like the file-structure of a release.

Example:
Web: `https://github.com/sass/node-sass/releases/download/v3.7.0/win32-x64-47_binding.node`
Temp Win7: `C:\Users\mriehema\AppData\Local\Temp\node-sass\v3.7.0\win32-x64-47_binding.node`

@saper, you should be able to easily fill the cache with scrips now.
@xzyfer, any more issues, questions, problems to be resolved?

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
+ *
+ * @param {String} binaryName
+ * @returns {string}
+ * @api private
+ */
+
+function getTempPath(binaryName) {
+  var candidateTmpDirs = [
+    process.env.TMPDIR || process.env.TEMP || process.env.npm_config_tmp,
+    '/tmp',
```
xzyfer: Use https://github.com/sindresorhus/os-tmpdir

mriehema: process.env.TMPDIR is the same as os-tmpdir. Line 140 is a fallback, which should never be reached. These lines are copied from the install script of phantomjs. They use it long ago without any problems.

xzyfer: That's ok, I'd still prefer you use a tested package that abstracts away this concern.

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
       return;
     }
 
-    download(sass.getBinaryUrl(), sass.getBinaryPath(), function(err) {
-      if (err) {
-        console.error(err);
-        return;
-      }
+    var tmpPath = getTempPath(sass.getBinaryName());
```
xzyfer: Needs to handle a failure to create the temp directory.

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
 
 /**
+ * Find a temp folder for file
+ *
+ * @param {String} binaryName
+ * @returns {string}
+ * @api private
+ */
+
+function getTempPath(binaryName) {
```
xzyfer: This function doesn't need to care about the binary name.

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
+  var candidateTmpDirs = [
+    process.env.TMPDIR || process.env.TEMP || process.env.npm_config_tmp,
+    '/tmp',
+    path.join(process.cwd(), 'tmp')
```
xzyfer: The last resort fallback should the current behaviour of installing in the local `vendor` directory.

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
 /**
+ * Find a temp folder for file
+ *
+ * @param {String} binaryName
+ * @returns {string}
+ * @api private
+ */
+
+function getTempPath(binaryName) {
+  var candidateTmpDirs = [
```
xzyfer:
```js
var tmpdir = require('os-tmpdir');
// ...
var candidateTmpDirs = [ tmpdir(), process.env.npm_config_tmp, sass.getBinaryPath() ]
```

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
-      }
+    var tmpPath = getTempPath(sass.getBinaryName());
+    if (!(sass.hasBinary(tmpPath))) { // download and install
+      download(sass.getBinaryUrl(), tmpPath, function(err) {
+        if (err) {
+          console.error(err);
+          return;
+        }
+        console.log('Binary downloaded at', tmpPath);
+        copyBinary(tmpPath, sass.getBinaryPath());
```
xzyfer:
```js
copyBinary(tmpPath + sass.getBinaryName(), sass.getBinaryPath());
```

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
+      });
 
-      console.log('Binary downloaded and installed at', sass.getBinaryPath());
-    });
+    } else { // install only
+      copyBinary(tmpPath, sass.getBinaryPath());
```
xzyfer:
```js
copyBinary(path.join(tmpPath, sass.getBinaryName()), sass.getBinaryPath());
```

scripts/install.js
```diff
@@ -109,18 +109,68 @@ function checkAndDownloadBinary() {
...
+}
+
+/**
+ * Copy file
+ *
+ * @param tmp
+ * @param dest
+ * @api private
+ */
+function copyBinary(tmp, dest) {
```
xzyfer: This function is named specifically for copying the binaries so you can make the binary name it's responsibility.

```js
var from = path.join(tmp, sass.getBinaryName());
```

mriehema:
@xzyfer, still something missing, or does this this PR look good to you now?

xzyfer:
Sorry @mriehema we don't get notified of new commits so I didn't see these changes.

scripts/install.js
```diff
@@ -109,18 +110,63 @@ function checkAndDownloadBinary() {
...
 
 /**
+ * Find a temp folder for file
+ *
+ * @returns {string} temp folder including binary file name
+ * @api private
+ */
+
+function getTempPath() {
+  var candidateTmpDir = tmpdir() || process.env.npm_config_tmp;
```
xzyfer: This isn't correct anymore. It's not either or, it's an array of candidates. What you used to have was correct.

```
var candidateTmpDirs = [
  tmpdir(), 
  process.env.npm_config_tmp, 
  sass.getBinaryPath(),
];
```

mriehema: Please take a look at the whole function. I removed the unnecessary foreach() loop of this list. It's now either `tmpdir()`, or `process.env.npm_config_tmp`. If they fail, `sass.getBinaryPath()` is used as fallback.

Maybe we should (temporary) disable the skip_ci part of this install script to test it on your CI.

xzyfer: I saw the change and IMHO it's incorrect since `tmpdir()` can never return a falsy value. Also, even if there is a tmpdir there is no guarantee it can be written to. The previous behaviour was correct because it asserted that either `tmpdir()` or `process.env.npm_config_tmp` existed and could be written to.

mriehema:
Closed. A) not mergable, B) see #1714

PR #1573, opened 2016-05-29
Added tests to verify #1564 'underscore files being compiled when input directory is inside a directory that starts with . (dot)'

### LINKED ISSUE #1564: underscore files being compiled when input directory is inside a directory that starts with . (dot)
Sample output:

```
node-sass ./ --output=./css
Rendering Complete, saving .css file...
Wrote CSS to /Users/lluna/project/.mocos/css/_base.css
{
  "message": "Undefined variable: \"$gray-light\".",
  "column": 35,
  "line": 3,
  "file": "/Users/lluna/project/.mocos/scss/helpers/_helpers.scss",
  "status": 1
}
```

When building from a regular directory (e.g. mocos instead of .mocos) it works (doesn't attempt to compile the _ files).

Using node-sass --version:
node-sass	3.2.0	(Wrapper)	[JavaScript]
libsass  	3.2.5	(Sass Compiler)	[C/C++]

lunalisk:
resolves #1564

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -520,6 +520,21 @@ describe('cli', function() {
       });
     });
 
+    it('should compile all files in a folder that starts with dot', function(done) {
+      var src = fixture('.input-directory/sass');
+      var dest = fixture('.input-directory/css');
+      var bin = spawn(cli, [src, '--output', dest]);
+
+      bin.once('close', function() {
+        var files = fs.readdirSync(dest).sort();
+        assert.deepEqual(files, ['one.css', 'two.css', 'nested'].sort());
+        var nestedFiles = fs.readdirSync(path.join(dest, 'nested'));
+        assert.deepEqual(nestedFiles, ['three.css']);
+        rimraf.sync(dest);
+        done();
+      });
+    });
+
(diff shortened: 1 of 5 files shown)


lunalisk:
It seems the cause was the version of glob used in v3.2.0 (5.0.3). In master it works fine, so I just added some tests to verify.

lunalisk:
Although on a second thought this may just not belong in sass since glob was causing it...

xzyfer:
Interesting, do you have a link to the glob issue handy?

Regards,
Michael

lunalisk:
I believe this is where they fixed it: https://github.com/isaacs/node-glob/issues/226

xzyfer:
@lunalisk so if I understand correct this bug only exists in older versions of node-sass because of a bug in a older version of node-gulp. Since we updated our node-gulp version the problem is gone in node-sass@3.7.0 (as the passing test would suggest).

If that's correct then I agree this test doesn't belong. Thank you for all your hard work tracking this down!

lunalisk:
Indeed, that's correct. I'll close #1564 and this one.

PR #1577, opened 2016-06-02
Copying bindings instead of downloading from github

ankurp:
This change is to help build in environments where machines do not have access to download from github/internet or to natively build sass binding by compiling.


xzyfer:
Sorry this is not something we want to do. We offer multiple configurations options loading binaries from the local filesystem or a networked machine within your firewall.

https://github.com/sass/node-sass#binary-configuration-parameters

All binaries are available on the [github release](https://github.com/sass/node-sass/releases) for your version.

PR #1578, opened 2016-06-06
In watch mode, play an audible cue when there is an error.

PeterHatch:
This adds a feature I like from the CoffeeScript compiler - when in watch mode, make an audible noise if there is an error processing a watched file.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -143,8 +143,11 @@ function getEmitter() {
   var emitter = new Emitter();
 
   emitter.on('error', function(err) {
-    console.error(err);
-    if (!options.watch) {
+    if (options.watch) {
+      // add the bell character, to make an audible noise
+      console.error(err + '\x07');
+    } else {
+      console.error(err);
       process.exit(1);
     }
   });


xzyfer:
Thanks @PeterHatch. I'm not opposed to this but it certainly should not be the default. If we were adopt something like this it would have be opt-in.

saper:
Given the amount of noise characters generated by various tools those days (ANSI colors for example) this seems harmless in comparison. I am relying a lot on screen + xterm + X windows system to process a bell beep as the X11 urgent notification. Can we check if we talk to the terminal before we generate a BEL? Is Windows working fine here?

PeterHatch:
I updated it to be an option (--error-bell) - definitely open to changing the name and/or description, if desired.

Windows is working fine, and is what I use (Windows 8.1, anyway).

For reference, CoffeeScript does this automatically when watching, and there was an issue about making it optional that was rejected - https://github.com/jashkenas/coffeescript/pull/3346.

Also an issue for adding it as an option to Webpack - https://github.com/webpack/webpack/issues/388 - which was rejected in favor of making a plugin for that behavior.

xzyfer:
My hesitation for this to be the default is because we get lots of issues over every trivial change. Some of the issues raise on the coffeescript issue are real, I know I'd hate to start receiving bells.

Could you also please update the usage documentation in the readme - https://github.com/sass/node-sass#usage-1

PeterHatch:
Updated the usage documentation.

And yeah, reading the coffeescript issue convinced me it should be an option.  I also like having any documentation of it, as it's most useful once you know to expect it.  And making it an option means you can turn it on when not using watch mode, which I think may be useful when doing things like using another program (like nodemon) to do the watching.

xzyfer:
Thanks for your contribution @PeterHatch!

PR #1586, opened 2016-06-10
update setup steps to comply with instructions

mohuk:
package.json will not be created without npm init and a --save is required to append the installed version of package onto the package.json

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -81,6 +81,7 @@ Firstly create a clean work space.
 ```sh
 mkdir \temp1
 cd \temp1
+npm init
 ```
 
 Gather some basic diagnostic information.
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -81,6 +81,7 @@ Firstly create a clean work space.
 ```sh
 mkdir \temp1
 cd \temp1
+npm init
```
xzyfer: This should be `npm init --yes` so you're prompted for unnecessary input.

xzyfer:
Alternatively, a better approach would be to move the section starting with

```
Note which version was installed by opening the `package.json` file with a text editor.
```

And replace it with `npm ls node-sass`

mohuk:
There is no need to doing an `npm init --yes` if we are doing `npm ls node-sass` because `npm ls` will list the package installed irrespective of the `package.json` file's existence. I think `npm ls node-sass` should suffice.

xzyfer:
Correct, this is what I was suggesting.

nschonni:
LGTM, but the output sample of the `npm ls node-sass` might be helpful to replace the old json example

PR #1593, opened 2016-06-13
Move the two CLI Usage into separate lines

howlowck:
Also, added the `or` word to clarify the two lines are not to be run sequentially.

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,7 +475,8 @@ The interface for command-line usage is fairly simplistic at this stage, as seen
 Output will be sent to stdout if the `--output` flag is omitted.
 
 ### Usage
- `node-sass [options] <input> [output]`
+ `node-sass [options] <input> [output]`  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -475,7 +475,8 @@ The interface for command-line usage is fairly simplistic at this stage, as seen
 Output will be sent to stdout if the `--output` flag is omitted.
 
 ### Usage
- `node-sass [options] <input> [output]`
+ `node-sass [options] <input> [output]`  
+ Or:  
  `cat <input> | node-sass > output`
```
xzyfer: I think the `or` makes the section too cluttered. What is missing is a new line, and some explanation.

The usage in the CLI `--help` is [much better](https://github.com/sass/node-sass/blob/master/bin/node-sass#L24-L34) and should be replicated here.

PR #1595, opened 2016-06-15
Prevention options from leaking

### LINKED ISSUE #1168: Error using array of importers
I get the following error when providing an array of importers (including when the array contains a single importer):
```
TypeError: bridge.success is not a function
    at done (<app path>/node_modules/node-sass/lib/index.js:306:20)
    at Object.options.importer.(anonymous function) (<app path>/node-sass/lib/index.js:312:13)
    at Object.options.importer.(anonymous function) (<app path>/node_modules/node-sass/lib/index.js:309:32)
    at options.importer.(anonymous function) (<app path>/node_modules/node-sass/lib/index.js:309:32)
```

xzyfer:
The options object passed into `render` and `renderSync` is only
shallow copied. This is an issue when nested objects and arrays like
custom importer and functions. In the case of custom function we
wrap the provided function. This wrapped version replaces the provided
one which can then leak back into the calling code. This is an issue
if `render*` is called in a loop like in gulp-sass.

Fixes #1168


test/api.js
```diff
@@ -535,19 +535,18 @@ describe('api', function() {
       });
     });
 
-    it('should copy all options properties', function(done) {
+    it('should wrap importer options', function(done) {
```
xzyfer: This test didn't make sense. It was actually asserting this issue existed.

Issue #1602, opened 2016-06-21
Test cases failures ( npm test ) on powepc


ghatwala:
Hi All,

 I am part of team officially porting node-sass on power-pc ( ppc64le - Ubuntu/Rhel ) . I am getting test case failures while trying out the test case on both ubuntu/rhel using below..
```
$ git clone https://github.com/sass/node-sass.git && cd node-sass
$ git submodule update --init --recursive
$ npm install
$ node scripts/build -f
$ npm test
```
My set up details are as below ( ubuntu/rhel )
```
[root@p006vm11 ~]# lscpu
Architecture:          ppc64le
Byte Order:            Little Endian
CPU(s):                16
On-line CPU(s) list:   0-15
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             16
NUMA node(s):          1
Model:                 IBM pSeries (emulated by qemu)
L1d cache:             64K
L1i cache:             32K
NUMA node0 CPU(s):     0-15
```
node version details
```
jenkins@fcbcd185e28f:~$ node --version
v6.2.1
jenkins@fcbcd185e28f:~$ npm --version
3.9.3
```
Attaching both error logs for ubuntu/rhel .

Any pointers on resolve the test case failures. ?
[node-sass-rhel.txt](https://github.com/sass/node-sass/files/325107/node-sass-rhel.txt)
[node-sass-ubuntu.txt](https://github.com/sass/node-sass/files/325108/node-sass-ubuntu.txt)

ghatwala:
Error log snippet , missed in above post 
```
$ npm test
1165 passing (16s)
  7 pending
  2 failing

  1) api binding missing error should be useful:
     Error: Node Sass does not yet support your current environment: Unsupported platform (Windows) Unsupported architecture (ppc64) with Node.js 6.x
For more information on which environments are supported please see:
https://github.com/sass/node-sass/releases/tag/v3.8.0
      at Object.<anonymous> (lib/index.js:13:11)
      at require (internal/module.js:20:19)
      at test/api.js:1807:24
      at _tryBlock (assert.js:318:5)
      at _throws (assert.js:337:12)
      at Function.assert.throws (assert.js:367:3)
      at Context.<anonymous> (test/api.js:1806:16)

  2) api binding missing error should list currently installed bindings:
     Error: Node Sass does not yet support your current environment: Unsupported platform (Windows) Unsupported architecture (ppc64) with Node.js 6.x
For more information on which environments are supported please see:
[...]

PR #1603, opened 2016-06-21
Added support for "ppc64le" arch to ensure test cases passing

ghatwala:
Had raised a git issue and was reported to be a bug , but after slack discussions got the solution ... Git issue url : https://github.com/sass/node-sass/issues/1602

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -41,6 +41,7 @@ function getHumanArchitecture(arch) {
     case 'ia32': return '32-bit';
     case 'x86': return '32-bit';
     case 'x64': return '64-bit';
+    case 'ppc64': return '64-bit';
     default: return false;
   }
 }


xzyfer:
Oh was this all? I thought maybe there'd be changes to the C++ code. Good to know.

xzyfer:
As we discussed in slack we cannot merge this as it constitute a level of
support for ppc64 we're not ready or equipped to give.

However since this check was all that failed we can improve the rest suite
to not count the current environment as invalid.

lib/extensions.js
```diff
@@ -41,6 +41,7 @@ function getHumanArchitecture(arch) {
     case 'ia32': return '32-bit';
     case 'x86': return '32-bit';
     case 'x64': return '64-bit';
+    case 'ppc64': return '64-bit';
```
nschonni: 2 cents, I think the human readable version might be "PowerPC 64-bit" rather than normalizing it like the rest to just "64-bit"

xzyfer:
As discussed we won't be merging this because it signifies a level of support we're not willing or able to provide. Instead we'll address the fragility of the test suite that causes test failures in unsupported environments.

lib/extensions.js
ghatwala: @nschonni - below is contents of lscpu on my ppc64le system
```
[root@p006vm11 ~]# lscpu
Architecture:          ppc64le
Byte Order:            Little Endian
CPU(s):                16
On-line CPU(s) list:   0-15
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             16
NUMA node(s):          1
Model:                 IBM pSeries (emulated by qemu)
L1d cache:             64K
L1i cache:             32K
NUMA node0 CPU(s):     0-15

```
However what node expect it as below
```
[jenkins@463995b0acbe Amex_Node_Sass_Rhel_7.2]$ node -p "process.arch"
ppc64
```

So that's the reason made above change

saper:
How can we distinguish ppc64le from ppc64be if anyone would provide binaries?

ghatwala:
@saper - That's a fair comment !! . i have seen that on "Little endian " machines, arch returns as "ppc64le" whereas on "Big Endian" , arch returns as "ppc64" .
@xzyfer -  any comments here to incorporate both endianess in node-sass code ?

xzyfer:
@ghatwala as previously mentioned this is not something we will do without an ambassador from the community.

ghatwala:
@xzyfer - acknowledged !!!!

PR #1615, opened 2016-06-29
watch: git checkouts are not atomic, so we must debounce to ensure co…

### LINKED ISSUE #1614: Add option to debounce watch
When switching between git branches, node-sass’ watch seems to trigger before the checkout has complete, causing the build to fail.

Is it possible to debounce the watch to prevent this from happening?

OliverJAsh:
…mpletion

Fixes #1614


OliverJAsh:
May have to increase the timeout, not sure yet.

package.json
```diff
@@ -59,6 +59,7 @@
     "get-stdin": "^4.0.1",
     "glob": "^7.0.3",
     "in-publish": "^2.0.0",
+    "lodash": "^4.13.1",
     "lodash.clonedeep": "^4.3.2",
```
xzyfer: Prefer a more specific lodash function dependency like this one

bin/node-sass
```diff
@@ -272,7 +274,7 @@ function watch(options, emitter) {
         renderFile(file, options, emitter);
       }
     });
-  });
+  }, 50));
 
   gaze.on('added', function(file) {
```
xzyfer: This will also need to be denounced, as well as the dates event below.

OliverJAsh:
Pushed fixes although still not sure this fixes my problem. Will report back.

xzyfer:
I think a denounce is a bad idea. It's optimises for the poor experience of
switch branches at the good experience of day to day watching. For this to
be acceptable the denounce would need to happen for both the leading and
trailing edge.

OliverJAsh:
> both the leading and
> trailing edge.

Sorry but could you elaborate on this? Not sure I follow.

xzyfer:
It's important to understand what debouncing is. Debounce groups all instances of an event until a period of time passes without that event being fired, in this case 300ms, then it fires a single event. The firing of the event can be done at two place.

### Trailing edge

This is the default and happens at the end of the timeout period i.e. 300ms after the last event. This means that event watch event will wait at least 300ms before it fires, which is terrible user experience.

screen shot 2016-06-29 at 9 01 20 pm

### Leading edge

This triggers the first time an event is fired, but wont fire again until there has been at least 300ms without an event.

screen shot 2016-06-29 at 9 00 50 pm

***

Unfortunately lodash doesn't allow you to fire on both the leading and trailing event. As a result I'm unconvinced this approach adds value, but it certainly does make the existing day-to-day experience worse.

xzyfer:
The [gulp watch implementation](https://github.com/gulpjs/glob-watcher/blob/master/index.js#L31-L63) looks like it could be better suited to the task.

OliverJAsh:
Want me to give that a go? If I get time

xzyfer:
If you have time that'd be great

seansfkelley:
[`_.debounce`](https://lodash.com/docs/4.17.11#debounce) does support both leading and trailing-edge firings, unless there's some requirement to use an older Lodash version that I'm not aware of?

saper:
Closing per https://github.com/sass/node-sass/pull/1615#issuecomment-229325086

PR #1617, opened 2016-07-01
add eyeglass cli option

xi:
It is still quite confusing how to use eyeglass, and I guess part of the problem is that it is not supported by the node-sass cli.

You might argue that this does not belong here. I had first tried to create an eyeglass cli that would mirror node-sass. But that would be a lot of code duplication. The best argument against including this would be if there were other libraries like eyeglass. Obviously we can not add options for all of them. But as far as I know, there is only this one.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -56,11 +56,13 @@ var cli = meow({
     '  --error-bell               Output a bell character on errors',
     '  --importer                 Path to .js file containing custom importer',
     '  --functions                Path to .js file containing custom functions',
+    '  --eyeglass                 Use eyeglass (requires that eyeglass is installed)',
     '  --help                     Print usage info'
   ].join('\n')
 }, {
   boolean: [
     'error-bell',
+    'eyeglass',
     'follow',
     'indented-syntax',
     'omit-source-map-url',
@@ -223,6 +225,11 @@ function getOptions(args, options) {
     }
   }
 
+  if (options.eyeglass) {
+    var eyeglass = require("eyeglass");
+    options = eyeglass(options);
+  }
+
   return options;
 }
 
@@ -309,7 +316,7 @@ function run(options, emitter) {
     emitter.emit('error', 'The --source-map option must be either a boolean or directory when compiling a directory');
   }
 
-  if (options.importer) {
+  if (options.importer && typeof options.importer === 'string') {
     if ((path.resolve(options.importer) === path.normalize(options.importer).replace(/(.+)([\/|\\])$/, '$1'))) {
       options.importer = require(options.importer);
     } else {
@@ -317,7 +324,7 @@ function run(options, emitter) {
     }
   }
 
-  if (options.functions) {
+  if (options.functions && typeof options.functions === 'string') {
     if ((path.resolve(options.functions) === path.normalize(options.functions).replace(/(.+)([\/|\\])$/, '$1'))) {
       options.functions = require(options.functions);
     } else {


bin/node-sass
```diff
@@ -309,15 +316,15 @@ function run(options, emitter) {
     emitter.emit('error', 'The --source-map option must be either a boolean or directory when compiling a directory');
   }
 
-  if (options.importer) {
+  if (options.importer && typeof options.importer === 'string') {
```
xi: I am not completely sure about this check. I had first used `util.isString()`, but that is deprecated. I also did not want to add additional dependencies like `lodash.isstring`.

chriseppstein:
I'd rather that node-sass have a public api that could allow projects like eyeglass to automatically load themselves if they conform to that api.

Tho, I guess that's basically the feature that eyeglass gives to node-sass so ¯\_(ツ)_/¯

xzyfer:
IMHO

> It is still quite confusing how to use eyeglass

This is a problem with the eyeglass documentation. It focuses heavily on assets, integrations, and custom function but lacks a good getting started guide. This was something I'd previously hoped to resolve but my plate is very full.

> it is not supported by the node-sass cli

I'm not sure these concerns overlap. It's not clear to me what to expect from an `--eyeglass` flag.

>You might argue that this does not belong here

I don't believe this belong in node-sass.

> if there were other libraries like eyeglass

I know there is at least [sassport](https://github.com/davidkpiano/sassport). I'm also aware of a handful of projects that integrate node-sass.

> I had first tried to create an eyeglass cli

This is what I would have expected.

>I'd rather that node-sass have a public api that could allow projects like eyeglass to automatically load themselves if they conform to that api.

I'm not sure what this would look like. We do lack something equivalent to Sass's `--require` option which might make sense.

***

We have previously discussed some kind of `.sassrc` file for configuring custom functions and importers. The idea has come up a few time but didn't get enough traction.

xi:
Thanks for the quick and extensive feedback!

After thinking about this again, I now believe that the focus is actually not that much on eyeglass, but on package managing: It feels as eyeglass could be to node-sass what npm is to node. In that regard it is also different from sassport.

>  We have previously discussed some kind of .sassrc file for configuring custom functions and importers.

I thought about this and it seemed like a good idea at first, but then I realised it would be like re-implementing grunt. I really have no intention of turning node-sass into a fully featured build tool.

I guess the problem is really that node-sass has no build-in support for package managing. Eyeglass solves that problem, but it does not work with standard node-sass cli.

I am not sure what to make of this yet. I will try to look into the autoloading idea mentioned by chriseppstein.

xi:
Just a quick idea to hear what you think of this: What if the option was not called `--eyeglass` but `--npm`?

xi:
> We do lack something equivalent to Sass's `--require` option which might make sense.

I tried this one now. `node-sass --require eyeglass` should do the same as `node-sass --eyeglass` did before.

xi:
I did not get any feedback for two month now. Is there anything I can do to get this merged?

saper:
Sorry for keeping you waiting @xi 

I remember I was also very confused trying to start with eyeglass. Can you give me short a short example how things get easier? Maybe we could add it the docs.

xi:
Running `node-sass --require eyeglass style.scss` will use eyeglass to resolve dependencies. But this could also be used as a hook for other tools.

xi:
I am not sure why appveyor fails. Looks unrelated to me.

PR #1631, opened 2016-07-13
`process.config.target_defaults` not available on some platforms (electron)

davej:
`process.config.target_defaults` is required for building node-sass but it is not available on Electron so fallback to using 'Release' when undefined.

re: https://github.com/electron/electron/issues/6462


xzyfer:
LGTM. Any objections @saper @am11?

am11:
LGTM2. Thanks!

saper:
Oops, my bad. Thanks!

PR #1647, opened 2016-07-29
add "flatten" to cli in order to output a flat directory

devinargenta:
- Add a flatten flag to the CLI `-f` or `-flatten` to output a flattened directory

So:
```
- scss/page.scss
- scss/page-types/page1.scss
```

will output to:

```
css/page.css
css/page1.css
```

I couldn't find a way to do this with the current CLI so I added this. Figured it may be useful for other people with nested directories of pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,7 @@ Example:
     -o, --output               Output directory
     -x, --omit-source-map-url  Omit source map URL comment from output
     -i, --indented-syntax      Treat data from stdin as sass code (versus scss)
+    -f, --flatten              Flattens the output directory
     -q, --quiet                Suppress log output except on error
     -v, --version              Prints version info
     --output-style             CSS output style (nested | expanded | compact | compressed)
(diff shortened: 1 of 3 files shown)


xzyfer:
Thanks for this @devinargenta. 

I have mixed feelings about adding more logic into the CLI, ideally it should be "dumb". 
- take a sass and output css
- take a directory of sass files and output a directory of css files

Thoughts @saper @nschonni

nschonni:
I think this is more something for a build system since I don't believe the Ruby Sass commandline offers this either. 
Might be good to look at the issue about separating out the CLI to a separate package so this could turn into `npm install node-sass-devargenta-awesome-cli`

xzyfer:
>I think this is more something for a build system

Agreed.

> Might be good to look at the issue about separating out the CLI to a separate package

I have concerns about this that I'll comment on that issue when I have some time.

xzyfer:
Thanks @devinargenta but this isn't something we think belongs in the CLI.

devinargenta:
Thanks for your responses, I found this useful for me because of my project set up and the CLI's watch tool being so powerful (utilizing sass-graph and not re-rendering every file)

Appreciate you for looking into this. Thanks for all your work on this project.

PR #1659, opened 2016-08-10
Build: Rename pretest to lint and only run once in CI

nschonni:
This allows us to bump beyond ESLint 2.x in the future since it dropped support for Node <4


xzyfer:
lgtm

PR #1661, opened 2016-08-10
Fix node vs nodejs discrepency

xzyfer:
Some Ubuntu OSes install `node` as `nodejs`. This breaks our npm lifecycle scripts that explicitly executes `node`. Npm's documentation suggests that a `*.js` lifecycle script should be executed by npm itself (presumably using the correct node binary).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,12 +27,12 @@
   },
   "gypfile": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


xzyfer:
Eh this didn't work unsurprisingly

PR #1664, opened 2016-08-12
Coveralls: Add back errors.js to sourcefiles

nschonni:
Looks like I accidentally reverted my fix while rebasing
d6cd013429e056461c8766091f0a1cad3259dd7a


nschonni:
Landing since the CI failure is just for a test timeout in 0.10

PR #1668, opened 2016-08-16
Test: Increase timeout for CLI tests

nschonni:
They tend to bomb out from time to time on CI, so using the method
described here http://stackoverflow.com/a/15982893/455535 and the docs
http://mochajs.org/#test-level to increase the timeout from 2 seconds
to unlimited

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -13,6 +13,8 @@ var assert = require('assert'),
 describe('cli', function() {
 
   before(function(done) {
+    // The CLI tests can be flakey on CI, so don't timeout for 10 minutes
+    this.timeout(600000);
     var bin = spawn(cli, ['-v']);
     bin.stdout.setEncoding('utf8');
     bin.stdout.once('data', function(data) {


test/cli.js
```diff
@@ -13,6 +13,8 @@ var assert = require('assert'),
 describe('cli', function() {
 
   before(function(done) {
+    // The CLI tests can be flakey on CI, so we give them more time
+    this.timeout(20000);
```
nschonni: I said 10 seconds in the commit message, but it's actually 20 😊

xzyfer:
We've played with altering the timeout value in the past. Ideally we'd figure out the root cause.

We should set a ceiling, rather than unlimited. Builds should fail in a timely manner, especially since we have a bound number of Travis jobs.

nschonni:
I didn't go digging for a link, but the CI systems (Travis at least) will fail if no output is received for 10 minutes. 
I agree that a root cause would be better, but at least this shouldn't confuse new contributors as much.

xzyfer:
:+1:

xzyfer:
Does the same apply for AppVeyor? I can't be sure which system it was but I've seen build run for hours.

nschonni:
Seems like Appveyor has a hard limit of 60 minutes for the whole build lifecycle, but it's late and I couldn't find the doc section that deals with any idle timeout so I'm going to ping @FeodorFitsner  (like a jerk :p) and go to bed

xzyfer:
OK I think we should just set a reasonable timeout. We want some visibility over how bad this issue is getting.

nschonni:
Alright, I've changed it to 10 minutes, so Travis should bomb out anyway, and Appveyor will behave the same.

nschonni:
Ha, now Travis is failing and not picking up the timeout value :(

xzyfer:
I wonder why tests have gotten so much more unstable in recent weeks :/

xzyfer:
Superseded by #1692

xzyfer:
> Ha, now Travis is failing and not picking up the timeout value :(

Looking at the build, the AppVeyor failure was still timing out. I did some investigation in #1689 and I'm convinced timeouts are a symptom of a larger issue.

PR #1680, opened 2016-08-25
Implement SASS_PATH

### LINKED ISSUE #1678: Is there any plan to support $SASS_PATH?
I know `node-sass` has the `--include-path` option. However, the original Sass documentation mentioned the ability to provide include paths with a `SASS_PATH` environment variable.

This would be very helpful for me, as I need `node-sass` to look in a specific set of include paths *by default* in my environment. Specifying `--include-path` every time is not an option for me in this case.

There was a discussion about [including this in libsass](https://github.com/sass/libsass/issues/90), but they concluded that it should be the wrappers' job, rather than the job of libsass itself.

nottrobin:
This is intended to [implement the `SASS_PATH` environment variable](https://github.com/sass/libsass/issues/90#issuecomment-57822790).

Fixes #1678.

## Tests

I've added two new API tests under `.render`:

- "should check SASS_PATH in the specified order"
- "should prefer include path over SASS_PATH"

If you run `mocha test/api.js` you should see these tests pass.

## Manual checking

You can test it manually using the test fixtures as follows:

### SASS_PATH is picked up

``` bash
$ fixdir=`pwd`/test/fixtures/sass-path
$ export SASS_PATH=$fixdir/red
$ bin/node-sass $fixdir/index.scss
body {
  background: red; }
```

### Earlier paths are preferred

``` bash
$ export SASS_PATH=$fixdir/orange:$fixdir/red
$ bin/node-sass $fixdir/index.scss
body {
  background: orange; }
```

### Specified include-paths still take precedence

``` bash
$ bin/node-sass $fixdir/index.scss --include-path $fixdir/red
body {
  background: red; }
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -16,7 +16,6 @@ if (!sass.hasBinary(sass.getBinaryPath())) {
   }
 }
 
-
 /**
  * Require binding
  */
@@ -165,6 +164,26 @@ function getLinefeed(options) {
 }
 
 /**
+ * Build an includePaths string
+ * from the options.includePaths array and the SASS_PATH environment variable
+ *
+ * @param {Object} options
+ * @api private
+ */
+
+function buildIncludePaths(options) {
+  options.includePaths = options.includePaths || [];
+
+  if (process.env.hasOwnProperty('SASS_PATH')) {
+    options.includePaths = options.includePaths.concat(
+      process.env.SASS_PATH.split(path.delimiter)
+    );
+  }
+
+  return options.includePaths.join(path.delimiter);
+}
+
+/**
  * Get options
  *
  * @param {Object} options
@@ -173,12 +192,13 @@ function getLinefeed(options) {
 
 function getOptions(opts, cb) {
   var options = clonedeep(opts || {});
+
   options.sourceComments = options.sourceComments || false;
   if (options.hasOwnProperty('file')) {
     options.file = getInputFile(options);
   }
   options.outFile = getOutputFile(options);
-  options.includePaths = (options.includePaths || []).join(path.delimiter);
+  options.includePaths = buildIncludePaths(options);
   options.precision = parseInt(options.precision) || 5;
   options.sourceMap = getSourceMap(options);
   options.style = getStyle(options);
--- a/test/api.js
+++ b/test/api.js
@@ -14,6 +14,11 @@ var assert = require('assert'),
 describe('api', function() {
 
   describe('.render(options, callback)', function() {
+
+    beforeEach(function() {
+      delete process.env.SASS_PATH;
+    });
+
     it('should compile sass to css with file', function(done) {
       var expected = read(fixture('simple/expected.css'), 'utf8').trim();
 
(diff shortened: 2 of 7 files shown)


lib/index.js
```diff
@@ -265,6 +264,18 @@ function normalizeFunctionSignature(signature, callback) {
  */
 
 module.exports.render = function(opts, cb) {
+  // Add include paths from SASS_PATH environment variable, if present
```
nschonni: This normalization probably belongs up in the getOptions `options.includePaths = (options.includePaths || []).join(path.delimiter);`, then there is no change anything else

test/api.js
```diff
@@ -137,6 +137,41 @@ describe('api', function() {
       });
     });
 
+    it('should check SASS_PATH in the specified order', function(done) {
+      var src = read(fixture('include-path/index.scss'), 'utf8');
+      var expected = read(fixture('include-path/expected.css'), 'utf8').trim();
+
+      var separator = process.platform == 'win32' ? ';' : ':';
+      process.env.SASS_PATH = fixture('include-path/functions');
```
nschonni: Could be wrong with the Mocha lifecycle for the `env`, but it's probably good to unset the `SASS_PATH` value in an after test function

test/fixtures/include-path/lib-alternate/vars.scss
```diff
@@ -0,0 +1 @@
+$color: orange;
```
nschonni: `lib-alternate` might be hard to follow, what about `just keeping it as `SASS_PATH`?

nottrobin: Sorry I didn't quite follow that. Do you mean put my fixtures in their own directory called "sass-path" or something?

test/api.js
nottrobin: Yep good idea

lib/index.js
nottrobin: I've tidied this up.

test/api.js
nottrobin: I started writing the code to save any initial value for SASS_PATH, then set it back to what it was afterwards. But then I thought it's actually quite unlikely that any changes to `process.env` would effect the environment outside of this particular `mocha` cycle, so I went to test it:

``` bash
$ export SASS_PATH=hello
$ mocha test/api.js
...
 129 passing (255ms)
$ echo $SASS_PATH
hello
```

So I think all I need to do it make sure SASS_PATH changes don't bleed through from one test to the other. So I've reset SASS_PATH to be an empty string before each test.

test/fixtures/include-path/lib-alternate/vars.scss
nottrobin: I've tried to make my fixtures simpler and clearer. I hope it's the sort of thing you had in mind.

nottrobin:
How can a travis build take quite so long? @nschonni is something up? Is there some way for me to start the travis build again?

saper:
Looks like OSX jobs got stuck and couldn't even be started.

here's an update from @travisci  : https://www.traviscistatus.com/incidents/4mvp857qx8bw

nottrobin:
Thanks @saper, I hadn't seen that. It's all passed now. @nschonni are you happy that I've address all your points?

saper:
Looks good to me. Would that be possible to squeeze this into one commit?

nottrobin:
@saper sure no problem

xzyfer:
Nice work everyone. Added this to the next.minor milestone. I believe we have one other PR to land in 3.9.0 also.

nottrobin:
Okay commits squashed into 3788c5d9570f2141ccbd6a574af21fcd57d63110.

nottrobin:
That's odd, it's failing on a test which is nothing to do with this commit. Any ideas?

saper:
Big thank you for the contribution!

nottrobin:
No problem. Thanks for all your help with this @saper et al., it was fun! Do you know when the next minor release will be?

PR #1692, opened 2016-08-31
Retry CLI specs that fail three times

### LINKED ISSUE #1311: Occasional AppVeyor node 0.10 test failures
````
1098 passing (5s)
1 failing

1) cli "before all" hook:
   Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
    at null.<anonymous> (s:\node_modules\node-sass\node_modules\mocha\lib\runnable.js:215:19)
````

xzyfer:
For some reason our test fail due to timeouts on seemingly random
cli tests. This instability has been growing worse. I've spent the
better part of the last two days looking into the issue without a
resolution. I'm starting to think it's an artefact cause by the
virtualisation used in CI, potentially related to resources
starvation.

The specs that failure do so trying to spawn a child process. The
process never spawns resulting in an eventual timeout. This is why
@nschonni attempt addressing this by increasing the timeout in #1668
failed.

Since the spawn failure is random, and catastrophic the best solution
in the short-medium term is to instruct mocha to retry failing tests
individually a couple times.

I've scoped this behaviour to the cli specs because at the moment
we have some specs that are not reentrant.

Fixes #1311

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -11,6 +11,9 @@ var assert = require('assert'),
   LIBSASS_VERSION = null;
 
 describe('cli', function() {
+  // For some reason we experience random timeout failures in CI
+  // due to spawn hanging/failing silently. See #1692.
+  this.retries(4);
 
   before(function(done) {
     var bin = spawn(cli, ['-v']);


xzyfer:
I think I may be able to confirm the spawn failure by inspecting the exit and error events. Will investigate tomorrow.

nschonni:
:shipit:

xzyfer:
Confirmed the `exit` event fires with a `0` code and no signal. This proves the spawn fails, but apparently also succeeds 😱

nschonni:
Might be good to take a look at our cross-spawn dependency a little to see if they've fixed anything that sounds familar in the newer versions. I think i did, but i don't recall if i found anything

xzyfer:
I dug into it. It's pretty thing. The interesting thing is that they use a separate code path for node 0.10 which appears to be the main offender. Node < 0.12 didn't have async spawn, so they add an async API to the sync spawn. It's possible this isn't playing nice with mocha's async model + filesystem operations.

Node < 4 are EOL'd by the end of the year so I'm not going to lose sleep over them for this.

PR #1694, opened 2016-09-01
Add a nice little progress bar while downloading binary

xzyfer:
Forked from #1649.

This PR adds a small ascii progress bar to the binary download.
Stalled installations make up a steady background noise of our
issues. It's my hope that by making the binary fetch more
transparent we can stem the tide of those issues.

In order to address @saper's [concerns][1] this progress bar will
respect npm's `progress` config flag.

sep-01-2016 21-33-35

Closes #1649

[1]: https://github.com/sass/node-sass/pull/1649#issuecomment-242720257

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -8,6 +8,7 @@ var fs = require('fs'),
   path = require('path'),
   sass = require('../lib/extensions'),
   request = require('request'),
+  ProgressBar = require('progress'),
   pkg = require('../package.json');
 
 /**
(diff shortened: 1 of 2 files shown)


xzyfer:
@nschonni taking your advice I've replaced the `progress` module with `npmlog`.

scripts/install.js
```diff
@@ -56,6 +57,8 @@ function download(url, dest, cb) {
     }
   };
 
+  console.log('Start downloading binary at', url);
```
nschonni: Was the `console.log` intentional? Not sure if the `log.newGroup` makes sense here

xzyfer: Yes it was. The user wont get any progress output until we get a response which won't happen for things like connection timeouts. 

Tracker groups not appear to produce any output. I could be wrong, the docs aren't useful.

xzyfer:
Happy to iterate on this. For now I want to get 3.9.0 out before tagging the next libsass.

PR #1695, opened 2016-09-01
Increase the binary fetch request timeout

xzyfer:
The timeout applies to both the initial connection as well as
individual packets.

I found the 1s timeout to be really fragile when testing locally.
The intent of the timeout is to prevent the install process hanging
indefinitely. It makes sense to be very generous so this increases
the timeout to 60s.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -50,7 +50,7 @@ function download(url, dest, cb) {
   var options = {
     rejectUnauthorized: false,
     proxy: getProxy(),
-    timeout: 1000,
+    timeout: 60000,
     headers: {
       'User-Agent': getUserAgent(),
     }


nschonni:
The test failures from Travis are odd, but don't seem related

xzyfer:
Yeah looks to be related to tests not cleaning after themselves + retries. I mentioned this could happen in #1692. I'll look at making the tests that touch the FS reentrant. IIRC @saper had some thoughts about this.

nschonni:
Maybe just move them to the allowed failures section? https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

xzyfer:
They shouldn't fail though. Specifically it happens because the test creates a file in a temp dir, then spawn  errors for an unrelated reason which causes a retry. On the second attempt the file is still there cause the create file step to error. Now that we know we [can catch the spawn failures](https://github.com/sass/node-sass/pull/1692#issuecomment-244030602) we can make sure to clean up after our selves before retrying.

PR #1700, opened 2016-09-06
ESLint 3.x updates

nschonni:
- Bump to the 3.x branch since we only run it in Node 4
- Add the recommended options, but left our existing overrides
- Rename the eslintrc file to JSON to match the newer format that allows it to skip guessing
- Delete the old jshintignores since it is no longer applicable
- Enforce and fix a few variable declaration issues
- Ensure the lint runs during the regular Travis script section so it fails the build

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -182,6 +182,7 @@ function getEmitter() {
  */
 
 function getOptions(args, options) {
+  var cssDir, sassDir, file, mapDir;
   options.src = args[0];
 
   if (args[1]) {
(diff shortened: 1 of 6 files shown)


bin/node-sass
```diff
@@ -182,6 +182,7 @@ function getEmitter() {
  */
 
 function getOptions(args, options) {
+  var cssDir, sassDir, file, mapDir;
```
xzyfer: Maybe it's personal preference I find this gross.

xzyfer: That's probably an argument for cleaning up the cli code in general.

xzyfer:
lgtm

bin/node-sass
nschonni: Yeah, if we get to just versions that support ES2015, then we can use stuff like `let` and `const`, but I don't think it's work adding in something like Babel just to get that right now

PR #1698, opened 2016-09-05
Update spec runner for full sass-spec coverage

nschonni:
- Update the spec runner from work in sass-spec
- Bump sass-spec tag to 3.3.6

PR #1705, opened 2016-09-06
Sass spec without Ruby and git submodules

### LINKED ISSUE #1686: Remove Ruby dependency from test

xzyfer:
Remove sass-spec git submodule by making the Sass Spec specs an npm package.
This compliments #1698 and together they accomplish #1686.

Fixes #1686

/cc @nschonni

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "libsass"]
 	path = src/libsass
 	url = git://github.com/sass/libsass.git
-[submodule "test/sass-spec"]
-	path = test/fixtures/spec
-	url = https://github.com/sass/sass-spec.git
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "istanbul": "^0.4.2",
     "mocha": "^2.4.5",
     "mocha-lcov-reporter": "^1.2.0",
-    "rimraf": "^2.5.2"
+    "rimraf": "^2.5.2",
+    "sass-spec": "^3.3.6-3"
   }
 }
--- a/test/spec.js
+++ b/test/spec.js
@@ -6,28 +6,12 @@ var assert = require('assert'),
   sass = process.env.NODESASS_COV
       ? require('../lib-cov')
       : require('../lib'),
-  util = require('./util');
+  util = require('./util')
+  spec = require('sass-spec').dirname;
 
 describe('spec', function() {
   this.timeout(0);
   var suites = util.getSuites();
-
-  describe('test/sass-spec directory', function() {
-    it('should be a cloned into place', function(done) {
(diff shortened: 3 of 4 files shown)


xzyfer:
Builds will fail until #1698 is merged

saper:
\o/

nschonni:
Closing manually since it didn't auto-close with #1698

PR #1706, opened 2016-09-06
Re-enable watcher tests

xzyfer:
These have been disabled for a while because they were suspected
of causing CI instability. We addressed a bunch of CI stability
issues in v3.9.0 that may have been incorrect attributed to these
specs.

I'm curious to see whether these are still causing issues. I'm
curious to see how CI holds up after a couple builds.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -226,7 +226,7 @@ describe('cli', function() {
       }, 100);
     });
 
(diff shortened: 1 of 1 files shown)


xzyfer:
Was worth a shot. Not going to sink time into this now. We're not touching the watcher/cli atm.

Issue #1709, opened 2016-09-07
Drop Node EOL versions support


nschonni:
Since node is dropping it from even security patching https://nodejs.org/en/blog/community/v5-to-v7/
- [ ] Remove from Appveyor
- [ ] Remove from Travis
- [ ] Update install script to print the version isn't supported. Something about the fact that it must be manually built and no bugs will be investigated.
- [ ] Update README supported version

0.10 and 0.12 may need their own issues depending on the timing since they EOL October and December respectively.

xzyfer:
We can make reasonable determinations about support from the download count of the release binaries. We get +2M installs a month so we should have a good data. Download counts are exposed via the HG release API.

https://developer.github.com/v3/repos/releases/#get-a-single-release

nschonni:
OK, I looked at the 3.8.0 release metrics since it was around for 2 months.
iojs seems to have only ~50 downloads per platform/arch, so I think it is safe to drop now.
5/47 still seems to have quite a few downloads, so I'm thinking that a warning message might be better for now

xzyfer:
Dropping iojs support has come up a few times. My argument has always been that it really costs us nothing to support old version of Node at this point aside from CI time. Is there a particular reason you want to drop older versions?

It's also worth noting we're planning to release LibSass 3.4 soon. This will essentially be a major version bump for LibSass. I was thinking this might be an opportune time to bump the major for node-sass if we had something to land.

nschonni:
I see it as one of those "why can't I run Windows XP forever things". Yes, it still works, but it's better to encourage people off platforms that don't get security updates anymore (especially devs)

xzyfer:
I don't necessarily agree. It's not hurting anyone, doesn't cause any work for us. Either way I don't feel comfortable dropping any support without a major bump given the most people will lock to `^3`.

xzyfer:
Looks like something went wrong here :) Mildly freaked out

nschonni:
Ah, that's what happened #1705 was the one that was supposed to autoclose with that issue

igor-savin-ht:
@nschonni So which versions are currently in-scope for removing within this task?

xzyfer:
See https://github.com/sass/node-sass/issues/2111

xzyfer:
This has been addressed in #2312

PR #1712, opened 2016-09-07
[DNM] Remove iojs building and warn on runtimes soon to be dropped by NPM

nschonni:
For discussion, didn't really test to see if the warnings would cause issues
Related to #1709


xzyfer:
This belongs in the same bucket as #1709. More discussion is required.

xzyfer:
We're to re-asses support for 5.0.

PR #1717, opened 2016-09-13
Update ISSUE_TEMPLATE with Bootstrap format

nschonni:
Took the basic structure from Bootstrap since they have a large project that gets a lot of issues

Not sure if the last line is to harsh. I also dropped their part about pointing people to SO for basic questions.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,11 +1,23 @@
-## Reporting an installation issue
-1. I'm running on a [supported platform](https://github.com/sass/node-sass-binaries/blob/master/README.md#compatibility)
-1. Have you recently installed a new version of Node? Native bindings are downloading for the specific version of Node when you install `node-sass`. Delete your `node_modules` and then re-install.
-1. Create [Gist](https://gist.github.com/) with the npm.log file and add it to this issue
-1. I've [searched for existing issues](https://github.com/sass/node-sass/search?type=Issues)
-1. I've followed the [Troubleshooting Guide](TROUBLESHOOTING.md)
+Before opening an issue:
+- Read the common workarounds in the [TROUBLESHOOTING.md](https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md)
+- [Search for duplicate or closed issues](https://github.com/sass/node-sass/issues?utf8=%E2%9C%93&q=is%3Aissue)
+- [Validate](http://sassmeister.com/) that it runs with both Ruby Sass and LibSass
+- Prepare a [reduced test case](https://css-tricks.com/reduced-test-cases/) for any bugs
+- Read the [contributing guidelines](https://github.com/sass/node-sass/blob/master/CONTRIBUTING.md)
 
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -1,11 +1,22 @@
...
-1. I've followed the [Troubleshooting Guide](TROUBLESHOOTING.md)
-
-## My Sass isn't compiling
-1. My code compiles on [SassMeister](http://www.sassmeister.com/) when using the Ruby Sass complier
-1. Try the Beta version of `node-sass` to see if the latest `libsass` fixes this
-1. Create a reproducible sample, and [open an issue on `libsass`](https://github.com/sass/libsass/issues/new). Compiler issues need to be fixed there, not here.
+Before opening an issue:
+- Read the common workarounds in the [TROUBLESHOOTING.md](https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md)
+- [Search for duplicate or closed issues](https://github.com/sass/node-sass/issues?utf8=%E2%9C%93&q=is%3Aissue)
+- [Validate](http://sassmeister.com/) that it runs with both Ruby Sass and libsass
```
xzyfer: LibSass

ISSUE_TEMPLATE.md
```diff
@@ -1,11 +1,22 @@
...
+
+When reporting an bug, **you must provide this information**:
+
+- NPM version (`npm -v`):
+- Node version (`node -v`):
+- Node Process (`node -p process.versions`):
+- Node Platform (`node -p process.platform`):
+- Node architecture (`node -p process.arch`):
+- node-sass version (`node -p "require('node-sass').info"`):
+
```
xzyfer: We also want `npm ls node-sass`

ISSUE_TEMPLATE.md
```diff
@@ -1,11 +1,22 @@
...
+When suggesting a feature, syntax, or compilation issue:
```
xzyfer: Suggesting in a feature is complicated. I would skip this because depending on the circumstance the feature belongs in either Ruby Sass, LibSass, or Node Sass.

ISSUE_TEMPLATE.md
```diff
@@ -1,11 +1,22 @@
...
+
+- [Open an issue on `libsass`](https://github.com/sass/libsass/issues/new). You
```
xzyfer: LibSass

ISSUE_TEMPLATE.md
nschonni: Should that replace the `require` version?

xzyfer: No having both is good. `npm ls node-sass` is so we can assert they don't have two, potentially, incompatible version of node-sass installed.

ISSUE_TEMPLATE.md
nschonni: Tweaked it to drop the feature part, but leave the LibSass reporting for syntax stuff

PR #1718, opened 2016-09-13
Note that libsass should be installed in readme

alexgleason:
Was having trouble installing an npm package on a fresh install of Ubuntu. While obvious that libsass needs to be installed, including it in the instructions should help.


nschonni:
Thanks, but LibSass doesn't need to be previously installed. You'll either install a pre-built binding file, or compile LibSass with the included source files

PR #1725, opened 2016-09-22
Respect NO_PROXY when downloading the binary

xzyfer:
Turns our handling proxies is complicated. The sanest thing to do
is try matching [npm's behaviour](https://github.com/npm/npm/commit/40afd6aaf34), and liable [to change](https://github.com/npm/npm/commit/40afd6aaf34) to
our benefit. The TLDR; of which is to let `request` do whatever
it wants unless npm has been explicitly configured otherwise.

Fixes #1724

/cc @nschonni @mmrath

PR #1733, opened 2016-09-26
POC for adding unit test coverage to install

xzyfer:
The installer is getting more complicated. The core of it is a http
request which makes testing it difficult/messy. Pulling out core
logic into small util functions that we can unit test would give me
much more confidence moving forward.

I would prefer we don't export functions from scripts so people don't
start relying on undocumented APIs. This is why I opted for helper
utils but I'm not committed to them.

@nschonni this is along the lines of what I was thinking in our Slack
chat today. I've applied it the same approach in #1725 as an example.
Would love to hear your thoughts.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -9,7 +9,7 @@ var fs = require('fs'),
   sass = require('../lib/extensions'),
   request = require('request'),
   log = require('npmlog'),
-  pkg = require('../package.json');
+  userAgent = require('./util/useragent');
 
 /**
  * Download file, if succeeds save, if not delete
@@ -53,7 +53,7 @@ function download(url, dest, cb) {
     proxy: getProxy(),
     timeout: 60000,
     headers: {
-      'User-Agent': getUserAgent(),
+      'User-Agent': userAgent(),
     }
   };
 
@@ -92,18 +92,6 @@ function download(url, dest, cb) {
   } catch (err) {
     cb(err);
   }
-}
-
-/**
- * A custom user agent use for binary downloads.
- *
- * @api private
- */
-function getUserAgent() {
-  return [
-    'node/', process.version, ' ',
-    'node-sass-installer/', pkg.version
-  ].join('');
 }
 
 /**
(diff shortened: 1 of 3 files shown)


test/scripts/util/useragent.js
```diff
@@ -0,0 +1,13 @@
+var assert = require('assert'),
+  pkg = require('../../../package.json'),
+  ua = require('../../../scripts/util/useragent');
+
+describe('useragent', function() {
```
nschonni: I'd probably keep this file at the root of `test` and just use other levels of `describe` if you're trying to show depth (for now)

xzyfer: You're completely right. I don't know what I was thinking. I was hungry.

xzyfer: Fixed.

xzyfer:
@nschonni bed time for me. If you think this is sane feel free to merge and follow along these lines for testable cache dir resolution in #1714.

You should also be able to commit to directly to this branch if you want.

test/useragent.js
```diff
@@ -0,0 +1,15 @@
+var assert = require('assert'),
+  pkg = require('../package.json'),
+  ua = require('../scripts/util/useragent');
+
+describe('util', function() {
+  describe('useragent', function() {
+    it('should look as we expect', function() {
+      var reNode = 'node/' + process.version;
+      var reSass = 'node-sass-installer/' + pkg.version;
+      var reUA = new RegExp('^'+reNode+' '+reSass+'$');
```
nschonni: Spacing around operators. I think ESLint should have caught that stuff, but maybe something is off in the config

scripts/install.js
```diff
@@ -9,7 +9,7 @@ var fs = require('fs'),
   sass = require('../lib/extensions'),
   request = require('request'),
   log = require('npmlog'),
-  pkg = require('../package.json');
+  ua = require('./util/useragent');
```
nschonni: think it is better to spell out `userAgent`

xzyfer: Fixed.

PR #1734, opened 2016-09-26
Remove skip download for CI flag

### LINKED ISSUE #1453: Reconsider the process.env.SKIP_SASS_BINARY_DOWNLOAD_FOR_CI
I've always had some doubts about this variable but during the v3.5.0 release process (#1451) it became obvious to me: why have a separate variable only to stop the download, if another (`SASS_FORCE_BUILD` or `-f` flag) is used to force the build anyway?

So, we have those effects:

SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=1 and SASS_FORCE_BUILD=0 : no download, the build will be started due to empty `vendor` directory if the binary hasn't been planted there otherwise (manual install)

SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=0 and SASS_FORCE_BUILD=1: download occurs, but the build proceeds anyway, overwriting the downloaded file (if successful).

Maybe we should use only SASS_FORCE_BUILD and drop the unofficial SKIP_SASS_BINARY_DOWNLOAD_FOR_CI ?

xzyfer:
This flag means we skip a bunch of code paths in CI which is not
ideal. Since we tell people to use `npm rebuild` we should eat our
own dog food.

With this patch we do an install, then do a rebuild hitting all the
major install code paths.

Fixes #1453

/cc @nschonni @saper

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ language: node_js
 
 compiler: gcc
 
-env:
-  global:
-    - SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
-
 jobs:
   include:
     - stage: test
@@ -114,6 +110,7 @@ install:
   - npm install
 
 script:
+  - npm run build --force
   - npm test
 
 cache:
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -96,11 +96,6 @@ function download(url, dest, cb) {
  */
 
 function checkAndDownloadBinary() {
-  if (process.env.SKIP_SASS_BINARY_DOWNLOAD_FOR_CI) {
-    console.log('Skipping downloading binaries on CI builds');
-    return;
-  }
-
   var cachedBinary = sass.getCachedBinary(),
     cachePath = sass.getBinaryCachePath(),
     binaryPath = sass.getBinaryPath();
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,7 +32,6 @@
     - '%AppData%\npm-cache'
 
   environment:
-    SKIP_SASS_BINARY_DOWNLOAD_FOR_CI: true
     matrix:
       - nodejs_version: 0.10
         GYP_MSVS_VERSION: 2013
@@ -86,6 +85,7 @@
     - node --version
     - npm --version
     - npm install
+    - npm run build --force
 
   test: off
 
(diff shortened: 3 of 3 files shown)


xzyfer:
Personally I would like to also kill off `SASS_FORCE_BUILD` so there's one-true-way which is `npm rebuild`. @saper may have some thoughts on that.

scripts/install.js
```diff
@@ -153,15 +153,6 @@ function checkAndDownloadBinary() {
 }
 
 /**
- * Skip if CI
- */
-
-if (process.env.SKIP_SASS_BINARY_DOWNLOAD_FOR_CI) {
```
xzyfer: I originally scheduled this for 4.0 since removing this is technically a breaking change. To get this out in 3.11.0 we should keep this block. As long as we're not using it in CI I'm happy.

nschonni:
I kind of like this https://github.com/mapbox/node-sqlite3#source-install, but I can see the downside of a generic flag too

xzyfer:
I'm surprised there's no offical way to do this. I couldn't even find signs on of a strong convention. I'll keep asking around.

saper:
`node-pre-gyp` has:

`npm install --build-from-source`

`node-pre-gyp --fallback-to-build` (I think most packages have this as the default).

My real concern is here is that we shouldn't be building twice.

.travis.yml
```diff
@@ -46,6 +42,7 @@ before_install:
 
 script:
   - npm install
+  - npm run build
```
xzyfer: Make sense, updated.

xzyfer:
Rebased this on the v5 branch

saper:
I am using this for my FreeBSD build machinery, with the following npm invocation:

   env CC.host=${CC} CXX.host=${CXX} LINK.host=${CXX} LINK.target=${CXX} V=1 
     SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=yes LIBSASS_EXT=yes 
     CXXFLAGS="${CXXFLAGS}"  LIBSASS_LIBRARY=${LOCALBASE}/lib/libsass_pic.a
     LIBSASS_CFLAGS=-I${LOCALBASE}/include NPM_CONFIG_CACHE=${DISTDIR}/npm 
     NPM_CONFIG_UNSAFE_PERM=1 NPM_CONFIG_NODEDIR=${LOCALBASE}/include/node
     NPM_CONFIG_CACHE=${BUILD_NPM_CACHE} npm install --nodedir=${LOCALBASE} --production

Haven't analyzed `rebuild` in detail; worst case if it cannot be coerced to do what I want I can always patch it in again.

PR #1744, opened 2016-10-11
Fix process argument binary configuration example

pselden:
Should be --sass-binary-site instead of --SASS_BINARY_SITE


xzyfer:
Thanks!

PR #1748, opened 2016-10-12
Add watch mode to tests

marvinhagemeister:
This also fixes process variables like `process.platform` not being properly restored after a test run because getters have a higher precedence.

Tested on both `v0.12.0` and `v6.7.0`.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -1839,4 +1839,139 @@ describe('api', function() {
       done();
     });
   });
+
+  describe('binding', function() {
+    beforeEach(function() {
+      delete require.cache[sassPath];
+    });
+
+    afterEach(function() {
+      delete require.cache[sassPath];
+    });
+
+    describe('missing error', function() {
+      it('should be useful', function() {
+        process.env.SASS_BINARY_NAME = 'Linux-x64-48';
+        assert.throws(
+          function() { require(sassPath); },
+          new RegExp('Missing binding.*?\\' + path.sep + 'vendor\\' + path.sep)
+        );
+      });
+
+      it('should list currently installed bindings', function() {
+        assert.throws(
+          function() { require(sassPath); },
+          function(err) {
+            var etx = require('../lib/extensions');
+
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
     "postinstall": "node scripts/build.js",
     "lint": "node_modules/.bin/eslint bin/node-sass lib scripts test",
     "test": "node_modules/.bin/mocha test",
+    "test-watch": "node_modules/.bin/mocha test -w",
```
xzyfer: Please remove this. You can already do this with npm.

```
npm test -- -w
```

marvinhagemeister: you're right, I've removed that

marvinhagemeister:
@xzyfer Sorry for not getting back any sooner. Upon revising this PR it turns out that getters are actually not needed at all and you can set the value of a property directly with `value` via `Object.defineProperty`. That is because the `process` variables are only marked as non writable which only [disallows assignments via an assignment operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty).

test/api.js
```diff
@@ -1839,4 +1839,139 @@ describe('api', function() {
       done();
     });
   });
+
+  describe('binding', function() {
+    beforeEach(function() {
+      delete require.cache[sassPath];
```
nschonni: I think @xzyfer removed this stuff in a recent PR, so I'll let him go through this again

xzyfer:
Thanks @marvinhagemeister I've update the `get()` to `value`.

marvinhagemeister:
@xzyfer The committed fix is incomplete. The leakage happens because `process.platform = prevValue;` does nothing. It doesn't assign the previous value because `process.platform` is a non writable property.

Non writable properties can only be overwritten with `Object.defineProperty`:

```js
Object.defineProperty(process, 'platform', {
  value: "whatever",
});
```

I've rebased the PR against master, but lack the permissions to reopen the PR.

Issue #1144, opened 2015-09-14
Default options via js file


chriseppstein:
This came up from an eyeglass issue: https://github.com/sass-eyeglass/eyeglass/issues/44#issuecomment-140175505

I'd like to be able to do `node-sass --defaults my-node-sass-defaults.js`

where my-node-sass-defaults.js

can be either a javascript or json file that would be required and whatever is exported will be assumed to be a js object that is the defaults for compiling. 

This would allow for trivial eyeglass integration or to pass other options that cannot be represented as CLI args (E.g. importers/functions)

xzyfer:
We've previously discussed configuration at length. Our primary concern was how to prevent custom importers and custom functions negatively affecting code portability like we see in the PostCSS ecosystem.

We agree that being able to autoload a set of custom importers and functions would ease the cognitive burden of using the CLI. We however see this a subset of a larger configuration problem. 

What happens if you depend on a 3rd party module that expect a custom function to have been loaded? It's either on your, or the 3rd party module's author to explicitly load the dependency. We think the responsibility lies with the 3rd party module's author.

Internally we load certain configuration from the `nodeSassConfig` object in our `package.json`. Our plan was to either extend this to support an `importers` and `functions` array, or move to an rc file format like `.eslintrc`. This would allow module to declare the importers and/or functions they expect (which also need to be an npm dependencies of that module) and we'll autoload them.

chriseppstein:
FWIW, ruby sass has this same philosophy as node-sass and ruby-sass has the -r option (and the --compass option) that effectively supports this.

> What happens if you depend on a 3rd party module that expect a custom function to have been loaded?

The important thing from an ecosystem perspective is to allow modules and their dependencies to be stated explicitly. npm gives most of that for free but it has to get loaded into node-sass. 

This is basically the problem that eyeglass solves, but until the community settles on a "winner" in the sass modules for node-sass space, it makes sense to support them all via some generic interface like the one I requested here.

> negatively affecting code portability like we see in the PostCSS ecosystem

The postcss issue arises because of the extreme nature of this problem. Loading functions and importers into options passed to node-sass is not going to kill an ecosystem -- it might just make it bumpy to get started.

PostCSS actually has a good solution to this problem now. The `@use` directive loads a post-css plugin for just the current file. If the plugin is not found, the user gets an error.

kristerkari:
I would like to bring back the discussion of a config file for `node-sass`. 

[...]

PR #1756, opened 2016-10-21
Add support for the --config CLI option

shawnbot:
This PR adds support for a `--config` CLI option, which allows you to configure `node-sass` via a standalone JSON or `.js` file:

``` sh
echo '{"outputStyle":"compact"}' > config.json
node-sass --config path/to/config.json foo.scss > foo.css
echo 'module.exports = {outputStyle: "compact"};' > config.js
node-sass --config config.js bar.scss > bar.css
```

This makes it much simpler configure multiple include paths for, say, both one-off builds and watch tasks, simplifying npm scripts:

``` json
{
  "scripts": {
    "build-css": "node-sass --config sass.config.js",
    "watch-css": "node-sass --config sass.config.js --watch"
  }
}
```

Some things to be aware of:
- There's currently no support for overriding options from the config via an explicit CLI argument because of the way that `meow()` is called. I think the way to do this would be to nix the `default` option from `meow()` and do something like:
  
  ``` js
  options = Object.assign({}, defaults, config, options);
  ```
- I reassigned the `-c` alias to `--config` from `--source-comments` (which was undocumented anyway), and added an `-s` alias for the latter.
- I originally had some tests for configuring `includePath` and `importer` via JSON, but these were difficult to get right, and brittle because the cwd from which they're resolved changes based on how you call `mocha`. It might be worth suggesting in the docs that users configure the "resolved" options (`includePath`, `importer`, and `functions`) via JS so that they can more easily specify absolute paths.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -34,6 +34,7 @@ var cli = meow({
     '    --source-map true --source-map-contents sass',
     '',
     'Options',
+    '  -c, --config               Specify configuration options via JSON or .js file',
     '  -w, --watch                Watch a directory or file',
     '  -r, --recursive            Recursively watch directories or files',
     '  -o, --output               Output directory',
@@ -72,6 +73,7 @@ var cli = meow({
     'watch'
   ],
   string: [
+    'config',
     'functions',
     'importer',
     'include-path',
@@ -83,7 +85,8 @@ var cli = meow({
     'source-map-root'
   ],
   alias: {
-    c: 'source-comments',
+    c: 'config',
+    s: 'source-comments',
     i: 'indented-syntax',
     q: 'quiet',
     o: 'output',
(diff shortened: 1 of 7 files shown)


bin/node-sass
```diff
@@ -227,6 +231,36 @@ function getOptions(args, options) {
...
+
+/**
+ * Resolve
+ *
+ * @param {String} file
+ * @param {String} name
+ * @api private
+ */
+
+function resolve(file, name) {
```
shawnbot: This function makes DRY the old logic in `run()` below, and emits better errors in the event that `require()` fails.

nschonni:
Thanks, but I'm :-1: on this because I'd like to keep the CLI the same as Ruby Sass.
There is disucssion about this type of feature in #1144

kaelig:
Actually, I found syntax disparities between the Ruby Sass cli and the node-sass cli somewhat upsetting. But I'm 👍  on this PR's feature.

xzyfer:
Yes this is something we're actively discussing

Issue #1776, opened 2016-10-29
Hosting binding.node built using node-chakracore


kunalspathak:
- NPM version (`npm -v`): 3.10.3
- Node version (`node -v`): v7.0.0-pre9 
- Node Process (`node -p process.versions`):
```cmd
{ http_parser: '2.7.0',
  node: '7.0.0-pre9',
  chakracore: '1.3.0.0',
  uv: '1.9.1',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  modules: '48',
  openssl: '1.0.2h' }
```
- Node Platform (`node -p process.platform`):win32
- Node architecture (`node -p process.arch`):
- node-sass version (`node -p "require('node-sass').info"`):
- npm node-sass versions (`npm ls node-sass`):

When encountering a syntax, or compilation issue:

- [Open an issue on `LibSass`](https://github.com/sass/LibSass/issues/new). You
may link it back here, but any change will be required there, not here

Currently `binding.node` [released](https://github.com/sass/node-sass/releases) are built using `nodejs v8`. Developers who want to use [node-chakracore](https://github.com/nodejs/node-chakracore), can't rely on this binary and need to build it locally using `node-chakracore`. I would like to start a discussion on what will it take to start hosting `binding.node` that is built using node-chakracore. Needless to say, this will need some changes in download script inside `scripts\install.js` to take care of downloading appropriate `binding.node` based on node engine that is running the install script.

*If you delete this text without following it, your issue will be closed.*

nschonni:
I could be wrong, but I think the best place to get this fixed/compatible would be with https://github.com/nodejs/nan

kunalspathak:
@nschonni , `node-chakracore` is already compatible with `nan` APIs, however `chakrashim`, the shim layer of node-chakracore has its own implementation of v8 data-structures. With that module compiled with `v8` can't work with `chakracore` because of difference in native objects layout, etc.

To give you an analogy, it is like someapp.exe can't work on Ubuntu even if the source code of someapp is same.

matthargett:
Very much looking forward to get getting fixed for the next release so we can use several popular react and react-native projects with node-chakracore. Let me know if any additional testing of the PR is needed before merging.

xzyfer:
I want Node Sass to support chakracore but we won't do a half-job of it. It will take a little while, but we will ship aim to ship officially support, not a hack.

matthargett:
[...]

PR #1777, opened 2016-10-29
Enable install and build using node-chakracore

kunalspathak:
This PR enables windows user to install and locally build this module using `node-chakracore`. See https://github.com/nodejs/node-chakracore/issues/136.

Individual commits have description of changes. If this is enabled, `node-chakracore` user can do the following in order to install `node-sass` module.
- Open 'Node.js chakracore command prompt`
- `npm install`

`npm install` would build the module locally using `node-chakracore` instead of downloading pre-built binary that is compiled with `nodev8`. See #1776.

Test result: All unit test pass.

``` cmd
  5908 passing (22s)
  322 pending
```


xzyfer:
For clarification, the crux of the issue is that `node-chakracore` isn't compatible with `node-gyp`?

scripts/build.js
```diff
@@ -132,14 +132,31 @@ function build(options) {
...
-    var args = [require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')), 'rebuild', '--verbose'].concat(
-      ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].map(function(subject) {
+    var args = ['rebuild', '--verbose'].concat(
+      ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].map(function (subject) {
         return ['--', subject, '=', process.env[subject.toUpperCase()] || ''].join('');
       })).concat(options.args);
 
-    console.log(['Building:', process.execPath].concat(args).join(' '));
+    // To make node-chakracore work, check if node-gyp is in the path.
+    // If yes, use it instead of using node-gyp directly from node_modules because it might not be
```
xzyfer: I find this comment to be rather misleading since it is only true if we're running under chakracore.

scripts/build.js
```diff
@@ -132,14 +132,31 @@ function build(options) {
...
+      process.env.Path.split(';').find(function (path) {
+        return path.startsWith(nodePath) &&
+        path.endsWith('node-gyp-bin');
+      }) !== 'undefined';
+
+    var exeName = process.execPath;
+    if (useInstalledNodeGyp) {
+      exeName = 'node-gyp';
+    } else {
+      args.unshift(require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')));
```
xzyfer: Am I correct in assuming this will fail if `process.jsEngine === 'chakracore'`?

xzyfer:
Generally speaking I'm 💯  for support chakracore. However we will either fully support it, or not support it all. IMHO half way solution is poor a experience for user, and a potential burden on us the maintainers.

To get this over the line we'll need to do the following at a minimum
- [ ] support chakracore in the binary resolution process i.e.`getBinaryName`
- [ ] update the `release` target in `.appveyor.yml` to build chakracore binaries in CI
- [ ] update the `default` target in `.appveyor.yml` to test against chakracore in CI

scripts/build.js
```diff
@@ -132,14 +132,31 @@ function build(options) {
       process.exit(1);
     }
 
-    var args = [require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')), 'rebuild', '--verbose'].concat(
-      ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].map(function(subject) {
+    var args = ['rebuild', '--verbose'].concat(
```
xzyfer: Please move the node-gyp resolution logic into it's own function.

kunalspathak: Sure.

kunalspathak:
@xzyfer , I have updated the PR to address the review comments. I have also updated to address `getBinaryName()` for different engines. By default (for nodev8) it will retain the existing format so released binaries will continue to work and downloadable. 

It will be appreciated if you can take care of `appveyor.yml` changes.

scripts/build.js
```diff
@@ -156,6 +151,35 @@ function build(options) {
...
+    }) !== 'undefined';
+
+  var exeName = process.execPath;
+  if (useInstalledNodeGyp) {
+    exeName = 'node-gyp';
+  } else {
+    args.unshift(require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')));
+  }
+
+  console.log(['Building:', exeName].concat(args).join(' '));
```
xzyfer: Please move this back to the `spawn` call site.

scripts/install.js
```diff
@@ -129,6 +129,11 @@ function checkAndDownloadBinary() {
   if (sass.hasBinary(binaryPath)) {
     return;
   }
+  
+  if (process.jsEngine && process.jsEngine === 'chakracore') {
```
xzyfer: This isn't required if we're supplying prebuilt binaries.

xzyfer:
@kamilogorek I'm happy to take a look. Is there any documentation to can recommend for getting node-chakracore update and running?

kamilogorek:
@xzyfer are you sure that you wanted to mention me? :)

xzyfer:
@kamilogorek I did not, apologies.

@kunalspathak 

> I'm happy to take a look. Is there any documentation to can recommend for getting node-chakracore update and running?

kamilogorek:
@xzyfer no worries! glad I could help :)

xzyfer:
I've opened a feature request with AppVeyor for Chakracore support.

xzyfer:
@kunalspathak can you please follow the steps in https://github.com/blog/2247-improving-collaboration-with-forks so I can collaborate on this PR.

scripts/install.js
kunalspathak: true.

scripts/build.js
kunalspathak: Sure.

kunalspathak:
@xzyfer , I have enabled "Allow edits from maintainers" for this PR. Let me know if you need anything else.

lib/extensions.js
```diff
@@ -181,9 +181,11 @@ function getBinaryName() {
   } else if (pkg.nodeSassConfig && pkg.nodeSassConfig.binaryName) {
     binaryName = pkg.nodeSassConfig.binaryName;
   } else {
+    var jsEngine = process.jsEngine ? '-' + process.jsEngine : '';
     binaryName = [process.platform, '-',
                   process.arch, '-',
-                  process.versions.modules].join('');
+                  process.versions.modules,
+                  jsEngine].join('');
```
nschonni: Seeing this now, might be better to `join('-')` to simplfy this, then just add the `jsEngine`

scripts/build.js
```diff
@@ -156,6 +153,34 @@ function build(options) {
...
 
+function resolveNodeGyp(options) {
+  var args = ['rebuild', '--verbose'].concat(
+    ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].map(function (subject) {
+      return ['--', subject, '=', process.env[subject.toUpperCase()] || ''].join('');
+    })).concat(options.args);
+
+  // For node-chakracore, check if node-gyp is in the path.
+  // If yes, use it instead of using node-gyp directly from
+  // node_modules because the one in node_modules is not
```
nschonni: Is it just a matter of bumping to a newer version in [package.json](https://github.com/sass/node-sass/blob/master/package.json#L68) or are users expected to have a custom version that only ships with node-chakracore?

kunalspathak: Users will have custom version of `node-gyp` that ships with node-chakracore.

lib/extensions.js
nschonni: Not a manditory change, I was just thinking this
```js
     binaryName = [process.platform,
                   process.arch,
                  process.versions.modules].join('-');
     binaryName += process.jsEngine ? '-' + process.jsEngine : '';
```
But maybe that isn't any more readable :)

kunalspathak: yeah. I was inclined towards something like this
```js
binaryName = [process.platform,
                  process.arch,
                  process.versions.modules,
                  process.jsEngine]
                  .filter(function(val) {return val;}).join('-');

```

scripts/build.js
nschonni: Since it looks like the only difference (by just browsing the History) for your copy of gyp looks like https://github.com/nodejs/node-chakracore/commit/023b96c446c90f75b2e98a6e6240f935ff7b8375 and https://github.com/nodejs/node-chakracore/commit/3a09525dd257357a3a763d4a0451daa1b3678c2c. Have you tried to just upstream those changes to gyp? Then it would just be a matter of bumping the minimum node-gyp version

nschonni: Ah, just noticed https://github.com/nodejs/node-gyp/pull/873

xzyfer:
@kunalspathak specifically I'm trying to install node-chakracore from the release binaries in our CI.

See my attempts in https://github.com/sass/node-sass/blob/trash/chakra/appveyor.yml#L122-L126
and the resulting build failure https://ci.appveyor.com/project/sass/node-sass/build/2103/job/23igf1j45b47bp0c

I'm not entire sure how to deal with the msi file in CI to get a functioning node.exe.

nschonni:
@xzyfer you might try the same as what AppVeyor does for the supported versions https://github.com/appveyor/ci/blob/master/scripts/nodejs-utils.psm1#L116

kunalspathak:
@xzyfer , regarding build failure, can you make sure that installed node is in the path? Also if you want to use same command prompt, you might want to call `<node_chakracore_installation_path>\nodevars.bat` which will set right environment variables.

xzyfer:
> can you make sure that installed node is in the path
@kunalspathak sorry for not being clear, my issue was that I don't know how to make an MSI file do those things :)

@nschonni aha this looks like what I was after. I took a look around that repo but got caught of the disparity between those function names and the ones in their docs.

kunalspathak:
@nschonni , there is extra customization that we did for `node-chakracore` installer in our private fork. You can diff below files for node vs. node-chakracore. It forces to use installed sdks/libs/headers rather than downloading from nodejs.org.

* `<node_installation_path>\node_modules\npm\node_modules\node-gyp\lib\configure.js`

image

* `<node_installation_path>\node_modules\npm\node_modules\node-gyp\addon.gypi`

image

nschonni:
I'll see if I can figure out what would be needed in the module to just natively support the install

lib/extensions.js
xzyfer: Prefer `_` if `val` is just a pass through.

```js
.filter(function(_) { return _; });
```

Personally I prefer code to be explicit.

```js 
var binaryName = [
  process.platform,
  process.arch,
  process.versions.modules
];

if (process.jsEngine) {
  binaryName.push(process.jsEngine);
}

return binaryName.join('-');
```

xzyfer:
I've updated this PR with the required AppVeyor config, as well tidying up the node-gyp resolution. 

All that's missing is some jsdoc comments for the new functions we I think this is done.

appveyor.yml
```diff
@@ -26,6 +26,15 @@
     - cmd: >-
         subst s: c:\projects
     - ps: set-location -path s:\node_modules\node-sass
+    - ps: Uninstall-Product node 4 $env:platform
+    - ps: Write-Host "Installing node-chakracore..." -ForegroundColor Cyan
+    - ps: Write-Host "Downloading..."
+    - ps: $msiPath = "$env:USERPROFILE\node-chakracore-$env:platform.msi"
+    - ps: $file = "https://github.com/nodejs/node-chakracore/releases/download/node-chakracore-7.0.0-pre9/node-chakracore-$($env:platform).msi"
```
kunalspathak: @xzyfer , what is the expectation here when we release new binary? Will we have to update this version? How does it work for `nodev8` when new versions are released?

scripts/build.js
```diff
@@ -154,32 +163,29 @@ function build(options) {
...
-      return path.startsWith(nodePath) &&
-      path.endsWith('node-gyp-bin');
-    }) !== 'undefined';
-
-  var exeName = process.execPath;
-  if (useInstalledNodeGyp) {
-    exeName = 'node-gyp';
-  } else {
-    args.unshift(require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')));
+  if (process.jsEngine && process.jsEngine === 'chakracore') {
```
kunalspathak: Here, it is ok to check `process.jsEngine === 'chakracore'`

lib/extensions.js
```diff
@@ -181,11 +181,17 @@ function getBinaryName() {
...
-                  process.versions.modules,
-                  jsEngine].join('');
+    binaryName = [
+      process.platform,
+      process.arch,
+      process.versions.modules,
+    ];
+
+    if (process.jsEngine) {
+      binaryName = binaryName.concat(binaryName);
```
kunalspathak: should be binaryName.concat(process.jsEngine);

kunalspathak:
@xzyfer , I noticed some issues and fixed in recent commits. If they look good, feel free to merge it.

xzyfer:
@kunalspathak thanks for the clean up.

appveyor.yml
xzyfer: This is barrier to offering official support. We also will not release official release binaries until node-chakracore can a offer stable ABI. This PR will enable node-chakracore compatibility, but no support.

kunalspathak: > We will not release official release binaries

Could you please elaborate on this? Do you mean that `binding.node` built using `node-chakracore` will not get hosted on https://github.com/sass/node-sass/releases ?

> until node-chakracore can a offer stable ABI

You mean until node-chakracore binaries are not available on CI system of nodejs.org, right?

xzyfer: > Do you mean that binding.node built using node-chakracore will not get hosted on 

Correct. Installing node-sass with node-chakracore will fallback to local compilation as were you originally doing in this PR.

> You mean until node-chakracore binaries are not available on CI system of nodejs.org

Exactly.

xzyfer: >I am confused about what was the intent of making .appveyor.yml changes as part of this PR

It was the best way IMHO to verify this PR is working as expected. As I originally stated we weren't willing to have hacks for node-chakracore when it doing the right way was just as easy.

> Was it a prep work so when node-chakracore binaries become available

Correct. Also having CI working with chakracore gives us confidence in support it.

kunalspathak: @xzyfer Thanks for the clarification. Let me know if anything else is needed to get this merged.

xzyfer: @kunalspathak no. IMO this is ready to go it's just a matter of finding the appropriate time to release it. I suspect it will after LibSass 3.4 stable see #1800.

xzyfer:
ChakraCore support has landed in node-gyp https://github.com/nodejs/node-gyp/pull/873

kunalspathak:
Ping. Any update on when this will get merged?

xzyfer:
Apologies I missed that nodejs/node-gyp#873 had landed. This will land soon, thanks.

kunalspathak:
Ping. Any ETA on when this will get merged?

xzyfer:
I'll take a look asap. I'm currently on the job hunt so a lot of things are
on hold. As an aside, do you know what wasn't support is like for chakra
core?

kunalspathak:
> I'm currently on the job hunt

All the best.

> do you know what wasn't support is like for chakra
core?

Sorry, I didn't understand your question. Could you please elaborate?

xzyfer:
Sorry I meant WASM support

kunalspathak:
I see. WASM is supported in Microsoft Edge, but not yet through chakracore. This is in our backlog.

xzyfer:
@kunalspathak I've squashed a bunch of commits and rebased off the latest master. I'm happy to ship this when CI is happy.

xzyfer:
Looks like the ChakraCore builds aren't playing nice with Nan? https://ci.appveyor.com/project/sass/node-sass/build/2548/job/8odim0l79rgg9i6g

kunalspathak:
Yes, this is a recent break in `node-chakracore` because of adding JSON specific APIs in `NaN`. 
cc - @boingoing who will fix it soon.

kunalspathak:
@xzyfer  - The `JSON` issue is address recently https://github.com/nodejs/node-chakracore/commit/85ccdd165603d9aade32c6e9dc204fcbe331bf15 . Do you have a rough ETA on when this can land?

xzyfer:
Aiming for roughly 48hrs.

appveyor.yml
xzyfer: @kunalspathak are you able to update this to use nvs? Pinning this to an out dated version of node-chakracore isn't great

xzyfer:
Kicking off CI again

saper:
CI says NAN JSON support is breaking the build

saper:
@kunalspathak can you bump the version of the chakracore in your patch? probably 7.0.0-pre9 does not include https://github.com/nodejs/node-chakracore/pull/313 yet

kunalspathak:
@saper - Yes, I will have a release that contains JSON fix and update this PR to use nvs.

kunalspathak:
@xzyfer , @saper  - I have updated the PR with NVS support for travis and appveyor. However seems travis doesn't support way to specify environment variables inside `jobs/stage` hierarchy (or at least I am not aware of how to do it). Can one of you take a look and fix it to set `NODEJS_VERSION` environment variable for each stage?

.travis.yml
```diff
@@ -1,45 +1,53 @@
...
     - stage: platform-test
-      node_js: "0.12"
+      NODEJS_VERSION: node/0.12
+      os: linux
+    - stage: platform-test
+      NODEJS_VERSION: node/0.10
       os: linux
     - stage: platform-test
-      node_js: "0.10"
+      NODEJS_VERSION: chakracore/latest
```
nschonni: Rather than changing over the entire matrix to NVS, what about just providing an alternate `before_install` just for these targets.
```yaml
    - stage: platform-test
      node_js: "node"
      os: linux
      before_install: ....
```

appveyor.yml
```diff
@@ -32,44 +32,53 @@
     - '%AppData%\npm-cache'
 
   environment:
+    # https://github.com/jasongin/nvs/blob/master/doc/CI.md
```
nschonni: Might be worth looking at https://github.com/appveyor/ci/blob/master/scripts/nodejs-utils.psm1 to get "native" support for Chakra on Appveyor

.travis.yml
kunalspathak: I reviewed that option, but the way travis works is if you specify `nvs_use`, it will internally use `nvm` to install node. There is no easy way to have both travis's `nvm` and `nvs` at same time. Is there any concern of using `nvs` that I can address in this PR?

appveyor.yml
kunalspathak: That is definitely an option (and likewise in `nvm` and hence `node-chakracore`), but at this time we just preferred using `nvs` since it added [support for CI](https://github.com/jasongin/nvs/blob/master/doc/CI.md) recently.

xzyfer:
Looks like the latest appveyor build hasn't run node-chakracore at all.

kunalspathak:
> Looks like the latest appveyor build hasn't run node-chakracore at all.

Yes, looks like some settings of this repo for `appveyor`. This repo is picking up `appveyor` of upstream branch instead of the copy present in PR. `travis` is working as expected i.e. it is picking my copy of `yml` file. Can you double check? I added similar PR for [expressjs/express](https://ci.appveyor.com/project/dougwilson/express/build/641) and it did pick my version of `appveyor.yml`.

scripts/build.js
```diff
@@ -84,6 +90,40 @@ function build(options) {
...
+ */
+function resolveNodeGyp() {
+  if (process.jsEngine === 'chakracore') {
+    // For node-chakracore, check if node-gyp is in the path.
+    // If yes, use it instead of using node-gyp directly from
+    // node_modules because the one in node_modules is not
+    // compatible with node-chakracore.
+    var nodePath = path.dirname(process.execPath);
+    var nodeGypName = process.platform === 'win32' ? 'node-gyp.cmd' : 'node-gyp';
+    var globalNodeGypBin = process.env.Path.split(';').filter(function(envPath) {
```
saper: is this correct for non-Windows?

what is the difference between chakraized node-gyp and the normal one?

scripts/build.js
```diff
@@ -100,9 +100,10 @@ function resolveNodeGyp() {
     // If yes, use it instead of using node-gyp directly from
     // node_modules because the one in node_modules is not
     // compatible with node-chakracore.
-    var nodePath = path.dirname(process.execPath);
-    var nodeGypName = process.platform === 'win32' ? 'node-gyp.cmd' : 'node-gyp';
-    var globalNodeGypBin = process.env.Path.split(';').filter(function(envPath) {
+    const nodePath = path.dirname(process.execPath);
```
xzyfer: Please don't use const. We still support node 0.10.

xzyfer:
@kunalspathak our appveyor is split into two sections, one for [releases][1], and one for normal [branches][2]. You've only update the release build which is why you're changes aren't being picked up for this branches builds.

[1]: https://github.com/sass/node-sass/blob/master/appveyor.yml#L2
[2]: https://github.com/sass/node-sass/blob/master/appveyor.yml#L103

scripts/build.js
kunalspathak: Sure.

scripts/build.js
kunalspathak: > is this correct for non-Windows?

good catch @saper . I have updated it with appropriate delim.

> what is the difference between chakraized node-gyp and the normal one?

See https://github.com/nodejs/node-gyp/pull/873 for more details and the [description](https://github.com/sass/node-sass/pull/1777/commits/eec8afa3c240f26261fc548698b2fe8ae77a131d) of my first commit.

In short, the changes are to add engine specific libs and headers while compiling native modules.

Since `node-gyp` PR has landed, we might not need this. The reason I still want to have this around is we could have a case of having a stale non-chakracore supported `node-gyp` in module path which will break building `node-sass` with `node-chakracore` as it breaks today.

xzyfer: We have node-gyp as a dependency in our package.json to get around these exact issues. You could bump that dependency to the minimum safe version.

xzyfer:
This is getting close. To fix the current CI error you'll need to update the [getHumanEnvironment][1] to understand there may be a fourth segment, the js engine, the binary name.

[1]: https://github.com/sass/node-sass/blob/aec7049bef2e5262083784b3fc5b6f7d97fb005c/lib/extensions.js#L83

kunalspathak:
Yes, figured it out. I will update the PR tomorrow. Although not sure how to fix `travis`. Any idea?

xzyfer:
My best guess is that these strings need to be quoted i.e. "node/8"

kunalspathak:
Doesn't look like that is the case. See [here](https://github.com/jasongin/nvs/blob/master/doc/CI.md#example-travisyml) for example. The problem is in a given matrix of `jobs/stages` I am not able to set those environment variables. I will do trial-n-error with various options tomorrow so bare with me if you see too many commits 😄 .

.travis.yml
```diff
@@ -1,45 +1,77 @@
...
+    - NVS_VERSION=1.3.0
     - SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
 
 jobs:
   include:
     - stage: test
-      node_js: "node"
+      env:
+        global:
+          - NODEJS_VERSION: node/6
```
xzyfer: This should be the latest node current release i.e. v8.1.2

xzyfer:
Looking that example I think the syntax is wrong

```diff
-NODEJS_VERSION: node/8
+NODEJS_VERSION=node/8
```

.travis.yml
kunalspathak: I was not sure what is picked by default when you specify `node_js` value to `node`. I will update it to `8` and nvs should pick the right version.

xzyfer: By default the `node` is the latest stable, currently 8.1.2. We'd want the equivalent to `node/latest`

kunalspathak:
I had that initially but had to revert it in [this commit](https://github.com/sass/node-sass/pull/1777/commits/c7124bbfdc190903822187925f9b018a85d11267) because `travis` was complaining about it.

xzyfer:
Hmm according to https://lint.travis-ci.org/ `=` should be fine. I'll give the doc a read tonight. Stages are still relatively new.

matthargett:
Ugh, this is bothersome. The reason for this last fail will generally make builds flaky:
```
Failed to query GitHub releases: https://github.com/nodejs/node-chakracore/releases/

HTTP response status: 403

{"message":"API rate limit exceeded for 208.78.110.204. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

kunalspathak:
@matthargett - The `chakracore/latest` jobs are failing because the repo need to add github API token to download chakracore releases as documented [here](https://github.com/jasongin/nvs/blob/master/doc/CI.md#github-api-tokens). @xzyfer  might help here?

xzyfer:
Ugh we've run into this before with LibSass. We don't have a good solution but it's been on my radar to fix. This makes it a priority. 

Although both Travis and AppVeyor support encrypted environment variables (ideal for storing github tokens) they're only available of repo builds, not forks/PRs for security reasons. The issue is made worse by GitHub not having read-only tokens.

Our proposed solution is to host a tiny reverse proxy that has a GitHub token baked it, and a whitelist of repos and queries it can perform. If we can point nvs at such a reverse proxy that would make our lives much easier.

xzyfer:
It looks like nvs can be configured to use an alternate remote which can be our reverse proxy to https://github.com/nodejs/node-chakracore/releases/

kunalspathak:
Yes, we are thinking of a way to host chakracore stable releases on `nodejs.org` or some alternate location that won't need API tokens.

xzyfer:
Until then I will aim to setup said reverse proxy tonight (~12hrs from now). It's been on my backlog for a long time.

xzyfer:
Reverse proxy to the GitHub API is available at https://github-api-reverse-proxy.herokuapp.com i.e. https://github-api-reverse-proxy.herokuapp.com/repos/sass/libsass/releases

There is a very restrictive whitelist of repos at the moment which includes only
- sass/libsass
- nodejs/node-chakracore

nschonni:
To avoid all the proxy stuff we can just stick with NVM. From their ChakraCore support thread I used https://gist.github.com/chicoxyzzy/77dcbf21485d80b3ad870ed0d3286579

https://github.com/nschonni/node-sass/commit/8869853c8995f4b7005364afa486321975c0331b#diff-354f30a63fb0907d4ad57269548329e3R47

EX: https://travis-ci.org/nschonni/node-sass/jobs/248655584

It's currently pulling down 9 with the `node` tag, but that could probably be changed to 8

xzyfer:
Well this is what I'm confused about

```
Failed to query GitHub releases: https://github.com/nodejs/node-chakracore/releases/
HTTP response status: 403
{"message":"API rate limit exceeded for 208.78.110.204. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

Suggests that it's hitting the ChakraCore "mirror" at https://github.com/nodejs/node-chakracore/releases. It's not clear which part of NVS is hitting the github API.

kunalspathak:
@nschonni , The url you have is for chakracore-nightly and what we want to support is chakracore-release. `chakracore-night` is already present in nvs and used under `chakracore-nightly\latest`. We are working on having some folder in `nodejs.org` for `chakracore-release` after which we can use `nvs` seemlessly.

nschonni:
I'm kinda 👎 on the NVS, i'm not seeing the benifit of the switch. If ChakraCore ends up in the `/dist/` index.tab/json that NVM will support it too

xzyfer:
I'd prefer we limit nvs to chakracore builds. Travis and appveyor use nvm
under the covers. If we can use configure nvm to install chakracore that'd
be ideal

saper:
I agree with @xzyfer and @nschonni - nvs is virtually unknown

xzyfer:
I've been looking into this, and there some realities we need to accept.
- NVM doesn't appear to want to support node-chakracore until it's releases are hosted until nodejs.org
- NVS is a good alternative to NVM for node-chakracore stable releases
- LibSass needs the proxy solution anyway so it's not a big deal

## Options:

### Hard coded versions
The GitHub API request is required to determine the `latest` node-chakracore release, and download the appropriate release binary. The versioned release binaries are currently just nightly builds hosted on GitHub releases. So we could hardcode the nightly version and periodically bump it i.e. `NODEJS_VERSION: chakracore-nightly/201706274be96a0e53`

We can do this with both NVM or NVS, however the I am not encouraged by the comments at NVM (https://github.com/creationix/nvm/issues/1067#issuecomment-308524339).

### Use the API proxy
This would be temporary under NVS. Once the stable binaries are hosted under nodejs.org we can ditch the proxy. We could also ditch NVS for NVM at this point but my previously mentioned concerns still apply.

I've opened an PR with NVS to add support for GitHub API proxies - https://github.com/jasongin/nvs/pull/61

### Official support

Ideally the Travis and AppVeyor support Node ChakraCore officially. There are upstream issues at https://github.com/appveyor/ci/issues/1148 and https://github.com/travis-ci/travis-ci/issues/8017

jasongin:
> The versioned release binaries are currently just nightly builds hosted on GitHub releases.

I had thought that also, since running `node --version` prints a "nightly" tag even on a "released" Node-ChakraCore build. However, @kunalspathak told me the releases are actually built with some different parameters. (I don't recall the details.) I suggested they should fix the `--version` output in that case, and he said he'd look into it.

xzyfer:
Thanks for the heads up @jasongin. I wonder if that matters for our CI. 

It possibly has some affect on the node-sass official release binaries, but that's still an open question.

nschonni:
There is now a `releases` index.tab https://nodejs.org/download/chakracore-release/ so NVS shouldn't be needed on the Travis side

kunalspathak:
@nschonni  - I am not that familiar with `nvm`. Are you saying that `nvm` should just work installing chakracore-release binaries? If yes, can you update the PR for travis changes?

nschonni:
@kunalspathak i'll see about a PR, but it would be the same as https://github.com/nschonni/node-sass/commit/8869853c8995f4b7005364afa486321975c0331b#diff-354f30a63fb0907d4ad57269548329e3R47 except 

`NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/chakracore-nightly/` 
becomes 
`NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/chakracore-release/`

kunalspathak:
Alright @nschonni . I will take a look at it.

appveyor.yml
```diff
@@ -32,44 +32,53 @@
...
+    NVS_VERSION: 1.3.0
```
nschonni: If you bump this to `1.4.0` it looks like the rate limit issues should go away https://github.com/jasongin/nvs/releases/tag/v1.4.0

kunalspathak: Yes, forgot to update it.

kunalspathak:
@nschonni  - Is there a way to trigger the build again? Seems like `appveyor.yml` picked the old version of nvs.

nschonni:
I will when i'm home. Quick thought would be if you updated it in both halves of the AppVeyor config since there is a duplicate Debug and Release/Tag sections                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: Kunal PathakSent: Tuesday, July 11, 2017 7:16 PMTo: sass/node-sassReply To: sass/node-sassCc: Nick Schonning; MentionSubject: Re: [sass/node-sass] Enable install and build using node-chakracore (#1777)@nschonni  - Is there a way to trigger the build again? Seems like appveyor.yml picked the old version of nvs.

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/sass/node-sass","title":"sass/node-sass","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/sass/node-sass"}},"updates":{"snippets":[{"icon":"PERSON","message":"@kunalspathak in #1777: @nschonni  - Is there a way to trigger the build again? Seems like `appveyor.yml` picked the old version of nvs."}],"action":{"name":"View Pull Request","url":"https://github.com/sass/node-sass/pull/1777#issuecomment-314598245"}}}

kunalspathak:
@nschonni  - That was it. Fixed it.

xzyfer:
Thanks again for all your work on this @kunalspathak and @nschonni

scripts/build.js
```diff
@@ -84,6 +90,19 @@ function build(options) {
 }
 
 /**
+ * Resolve node-gyp command to invoke
+ *
+ * @api private
+ */
+function resolveNodeGyp() {
```
nschonni: Do we still need this now that the version is bumped to 3.6?

appveyor.yml
```diff
@@ -32,44 +32,53 @@
...
         APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-      - nodejs_version: 7
+      - NODEJS_VERSION: node/7
         GYP_MSVS_VERSION: 2015
         APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-      - nodejs_version: 8
+      - NODEJS_VERSION: node/8
         GYP_MSVS_VERSION: 2015
         APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      - NODEJS_VERSION: chakracore/8.1.2
```
nschonni: Will a plain `chakracore/8` work here so we don't need to bump it regularly?

xzyfer: This raises the question of what kind of ABI guarantees we get with CharkraCore. 

Can we continue to rely on process.modules.version as the ABI version?
Is CharkraCore supported by N-API?
What's the estimated time line on WASM support in CharkraCore?

matthargett: * WASM has been in ChakraCore since 1.4
* node-chakracore 8.x is included in the N-API repo ( @digitalinfinity)

xzyfer: Thanks @matthargett. I was under the impression from https://github.com/sass/node-sass/pull/1777#issuecomment-308307135 that WASM wasn't supported in node-chakracore. Can you please confirm?

digitalinfinity: WASM is supported in Node-ChakraCore if you're using the WebAssembly methods from node. Using basic.wasm from [here](https://github.com/Microsoft/ChakraCore/blob/master/test/wasm/basic.wasm), the following code worked with Node-ChakraCore:
```
const fs = require('fs');
const buf = fs.readFileSync('basic.wasm')

async function test() {
    try {
        const module = await WebAssembly.compile(buf);
        const inst = new WebAssembly.Instance(module, {test: {foo: function(a){console.log(`foo called: ${a}`); return 2;}}});
        console.log(inst.exports.a(1));
    } catch (reason) { 
        console.log(`Failed: ${reason}`)
    }
}

test();
```

Note that ChakraCore doesn't expose JSRT APIs for embedders to directly invoke WASM from their applications- the workaround is they'll have to do it through JavaScript.

Also, minor technical point but while node-chakracore was included in the N-API repo during N-API prototyping, N-API has merged into the master branch of nodejs/node, so active N-API development for Node-ChakraCore is now happening in the master branch of nodejs/node-chakracore. But the meta point still stands that node-chakracore supports N-API.

scripts/build.js
kunalspathak: No and that's why i removed my changes in https://github.com/sass/node-sass/pull/1777/commits/5021dff1df351c5a3c4f3c81cbfb65c09ec23aa2 . I just kept the resolution of `node-gyp` in separate function for easy readability.

kunalspathak:
@nschonni  - I fixed `chakracore` version to `chakracore/8`.

scripts/build.js
nschonni: :+1:  that makes sense

kunalspathak:
Thanks @nschonni  and @xzyfer  so far for shaping the PR. Let me know if you need anything else to merge this PR.

saper:
What are the `process.versions.module` values for chakracore engines? How often do they change? How many shall we support?

kunalspathak:
@saper - did you mean to say `process.versions.modules`. If yes, it is `57` currently and has same value as node-v8. We don't control it and it will be set from upstream node that we inherit.

saper:
Thanks. This is mainly used to denote changes in the V8 API, I thought Chakra might have come up with something different?

saper:
Thanks. This is mainly used to denote changes in the V8 API, I thought Chakra might have come up with something different?

saper:
In other words: suppose I have downloaded a chakracore node-sass binary module 57 today. How would I know I need a new one, say, 6 months from now? Is API stability guaranteed? We have lots of problems with users forgetting to update their modules.

kunalspathak:
It should work similar to way it works with node-v8. So lets say in future node.js 10.X , the modules version was updated to say `79`, when we release node-chakracore from 10.X branch, it will have same module versions i.e. `79` because to release that we would have shimmed required v8 APIs . Does that make sense ?

saper:
Do I understand this correctly @kunalspathak that chakracore is V8-API compatible? If so, what is breaking ABI compatibility, so that the existing binaries cannot be used?

kunalspathak:
> chakracore is V8-API compatible

Yes. In order for `node-chakracore` to work we shim all the V8 APIs that node or a native module calls into.

saper:
ok, that solves the API problem, i.e. module compiled by hand should generally work.

How do we ensure ABI compatibility between modules?

Examples on situations where ABI may work:
* You are bumping ChakraCore engine version (for example from 1.5.2 to 1.5.3). It may or may not break the ABI - both ChakraCore engine versions are attached to `NODE_MODULE_VERSION` 57. 
* You decide to do some changes in the chakrashims: for example you decide to allow more than 8 local variables on the stack and therefore bump `kOnStackLocals` to something else, which will almost probably break ABI.

Even today, will node-sass binding.module compiled with ChakraCore 1.5.2 work with 1.5.3? Is the reverse true?

(I cannot get ChakraCore to build, so I cannot test at the moment).

This is a major concern from my side - if we are to offer binary module downloads, ABI stability must be guaranteed.

kunalspathak:
> This is a major concern from my side - if we are to offer binary module downloads, ABI stability must be guaranteed.

That is a valid concerns. However ABI compatibility is guaranteed by `chakrashim` and not `chakracore`. So even if we switch chakracore versions, as long as chakrashim is v8 API compatible, it should just work.

> for example you decide to allow more than 8 local variables on the stack and therefore bump kOnStackLocals to something else

I am not sure how this will break the compatibility. This is purely chakrashim's optimization that is nothing to do with breaking compatibility.

saper:
If my binding is compiled with kOnStackLocals = 8 and v8handlescope.cc coming from the updated ChakraCore uses, say, 16, couldn't that lead to a mismatch in the size of `HandleScope` ?

In general: can we have something that will be bumped whenever one of the `deps/` exposed to the modules changes?

kunalspathak:
> In general: can we have something that will be bumped whenever one of the deps/ exposed to the modules changes?

We could have that, but the plan is not to diverge chakrashim from v8 headers unless v8 changes come from upstream. So in that case, version will only get bumped when upstream bump the version.

xzyfer:
I touched on some of these concerns earlier. IIRC we greatly reduce risk from ABI changes by adopting N-API which is very close on our road map, and may land in the same release as this.

kunalspathak:
@xzyfer , do you have an ETA on when this can land?

saper:
This is a volunteer-driven project, so no ETAs here. I am concerned with releasing something quick now and worry about future compatibility "later".

kunalspathak:
Sorry @saper , I didn't mean to pressurize to land this PR prematurely. Its been a while the PR is open and I thought I addressed all the concerns. Let me know if there are any other issues that you want me to address and I will be happy to fix them.

davej:
@kunalspathak Would you consider maintaining a fork and publishing to npm? Would love to use this as many of my frontend projects are constantly requiring re-builds when I upgrade node. Usually node-sass is the only native dep so would love to use this.

kunalspathak:
Thanks for asking @davej , but I am not sure if I understand your problem. This PR simply adds a support to build and work with `node-chakracore`. When you upgrade node, you might still need to rebuild.

davej:
@kunalspathak sorry I skimmed the issue and saw a mention of N-API and assumed that's what the PR was about.

nschonni:
Sorry, closing since ChakraCore node is archived right now

PR #1782, opened 2016-11-03
Handle permission errors when attempting to cache binaries

### LINKED ISSUE #1781: Recieve error when install node-sass by yarn .
I have installed node-sass by yarn, but I get error: 
```
error /home/zendkofy/public_html/react-redux/visualsliderdesigner/node_modules/node-sass: Command failed.
Exit code: 1
Command: sh
Arguments: -c node scripts/install.js
Directory: /home/zendkofy/public_html/react-redux/visualsliderdesigner/node_modules/node-sass
Output:
path.js:7
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:1211:7)
    at /home/zendkofy/public_html/react-redux/visualsliderdesigner/node_modules/node-sass/scripts/install.js:122:26
    at /home/zendkofy/public_html/react-redux/visualsliderdesigner/node_modules/mkdirp/index.js:30:20
    at FSReqWrap.oncomplete (fs.js:123:15)
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```
Could you please let me know to fix this error ?

xzyfer:
This uses a candidate model for determining an suitable cache
directory. This is different from the current approach with eagerly
chooses a cache directory. If that directory isn't accessible for
some reason (i.e. permissions) we have to bail out. With this
approach we can try a prioritised list of candidate directories.

Additionally this address the case where npm cache config is
blank.

Fixes #1781

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -262,6 +262,17 @@ function getBinaryPath() {
 }
 
 /**
+ * Looks for the configured cache path. If none is found, fall back to the NPM
+ * cache folder
+ *
+ * @api public
+ */
+function getCachePath() {
+  return process.env.npm_config_sass_binary_cache ||
+         process.env.npm_config_cache;
+}
+
+/**
  * Does the supplied binary path exist
  *
  * @param {String} binaryPath
@@ -289,6 +300,7 @@ module.exports.hasBinary = hasBinary;
 module.exports.getBinaryUrl = getBinaryUrl;
 module.exports.getBinaryName = getBinaryName;
 module.exports.getBinaryPath = getBinaryPath;
+module.exports.getCachePath = getCachePath;
 module.exports.getVersionInfo = getVersionInfo;
 module.exports.getHumanEnvironment = getHumanEnvironment;
 module.exports.getInstalledBinaries = getInstalledBinaries;
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -5,6 +5,7 @@
 var eol = require('os').EOL,
   fs = require('fs'),
   pkg = require('../package.json'),
+  mkdir = require('mkdirp'),
   path = require('path'),
   defaultBinaryPath = path.join(__dirname, '..', 'vendor');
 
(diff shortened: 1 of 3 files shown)


xzyfer:
This is super messy but gets the job done.

mriehema:
Node-sass 3.11.0 included caching the binaries (#1554). Yarn caused a problem and you removed it in 3.11.1 and this is a new fixed PR with caching again for a future version. Did I get it right?

xzyfer:
>Node-sass 3.11.0 included caching the binaries

Correct. 

> Yarn caused a problem and you removed it in 3.11.1 

Yes and no. We removed it due to two issues.

1) we were querying npm for it's cache directory, and using it. However npm scripts run under a restricted user which may not have write permissions to that directory.

2) yarn isn't correctly handling npm config the same as npm (see https://github.com/yarnpkg/yarn/issues/1632). The result of this issue is that we would receive `''` as the cache directory. We were not handling this gracefully.

> this is a new fixed PR with caching again for a future version

Correct. We are re-enabling the caching. This time have a set of candidate directories (this was in your original PR but was removed) which could be used a cache. We try each one until we find out that is writeable and use it. Otherwise we fallback to the `vendor` directory which is the same as not caching.

xzyfer:
@nschonni this is pretty close now, care to take a look?

mriehema:
@xzyfer Thank you!

lib/extensions.js
```diff
@@ -260,6 +261,80 @@ function getBinaryPath() {
...
 
 /**
+ * An array of paths suitable for use as a local disk cache of the binding.
+ *
+ * @return {[]String} an array of paths
+ * @api public
+ */
+function getCachePathCandidates() {
+  return [
+    process.env.npm_config_sass_binary_cache,
```
nschonni: Maybe, just use `||` for falsey checking instead of the filter

nschonni: Nevermind, saw that you're looping through later, so you want all truethy values

lib/extensions.js
```diff
@@ -260,6 +261,80 @@ function getBinaryPath() {
...
+ *
+ * @return {String} directory to cache binding
+ * @api public
+ */
+function getBinaryCachePath() {
+  var i,
+    cachePath,
+    cachePathCandidates = getCachePathCandidates();
+
+  for (i = 0; i < cachePathCandidates.length; i++) {
```
nschonni: Since the `i` isn't used outside of the element lookup, `foreach in` might be better

lib/extensions.js
```diff
@@ -260,6 +261,80 @@ function getBinaryPath() {
...
+ * @api public
+ */
+function getCachedBinary() {
+  var i,
+    cachePath,
+    cacheBinary,
+    cachePathCandidates = getCachePathCandidates(),
+    binaryName = getBinaryName();
+
+  for (i = 0; i < cachePathCandidates.length; i++) {
```
nschonni: `foreach`

scripts/install.js
```diff
@@ -92,37 +92,47 @@ function download(url, dest, cb) {
  */
 
 function checkAndDownloadBinary() {
-  if (sass.hasBinary(sass.getBinaryPath())) {
+  if (process.env.SKIP_SASS_BINARY_DOWNLOAD_FOR_CI) {
+    console.log('Skipping downloading binaries on CI builds');
```
nschonni: Might want to use the npmlog stuff from the other PR

lib/extensions.js
```diff
@@ -260,6 +261,80 @@ function getBinaryPath() {
...
+ * An array of paths suitable for use as a local disk cache of the binding.
+ *
+ * @return {[]String} an array of paths
+ * @api public
+ */
+function getCachePathCandidates() {
+  return [
+    process.env.npm_config_sass_binary_cache,
+    process.env.npm_config_cache,
+    getBinaryPath(),
```
nschonni: I think this should be removed since it's not really a cache path, but the final fallback

scripts/install.js
```diff
@@ -92,37 +92,47 @@ function download(url, dest, cb) {
...
     return;
   }
+  var binaryPath = sass.getBinaryPath();
 
-  mkdir(path.dirname(sass.getBinaryPath()), function(err) {
+  if (sass.hasBinary(binaryPath)) {
+    return;
+  }
+
+  mkdir(path.dirname(binaryPath), function(err) {
```
nschonni: Since it looks like fs.stat https://nodejs.org/api/fs.html#fs_fs_stat_path_callback isn't supposed to be called before doing the write, maybe try to create the cachePath here, then fallback to the vendor if it can't

lib/extensions.js
xzyfer: `forEach` makes exiting early very messy.

lib/extensions.js
xzyfer: `forEach` makes exiting early very messy.

scripts/install.js
xzyfer: This was already here, it has just been moved. The npmlog issue is a separate concern.

scripts/install.js
xzyfer: I'm not sure what `fs.stat` has to do with `mkdirp`? 

Having the vendor directory in the cache candidates list means that if we can't write to the npm cache path then we are writing to vendor. The fallback has already happened.

scripts/install.js
```diff
@@ -92,37 +92,47 @@ function download(url, dest, cb) {
...
+
+    download(sass.getBinaryUrl(), cacheBinary, function(err) {
       if (err) {
         console.error(err);
         return;
       }
 
-      console.log('Binary downloaded and installed at', sass.getBinaryPath());
+      console.log('Binary downloaded to ' + cacheBinary);
+      fs.createReadStream(cacheBinary).pipe(fs.createWriteStream(binaryPath));
```
xzyfer: The worse case here is that we risk copying from the vendor dir to the vendor dir. Could be problematic.

lib/extensions.js
xzyfer: Expanding on this, `forEach` is intended to creating side effects. This is closer to a `reduce` but that also really messy to exit early from.

scripts/install.js
xzyfer: Are you suggesting we move the `mkdir` into the `download` callback?

lib/extensions.js
xzyfer: I'm not sure, I need to think about this. In the install script we only ever care about the final destination. The edge case is that it's possible for the cache directory and vendor directory to be one and the same.

xzyfer: Nope, you're completely right, this is a bad abstraction.

xzyfer:
I made some revisions. I'm not in love with it but it's an improvement. IMHO the install script needs a lot of love.

PR #1786, opened 2016-11-03
Switch from console.log to npmlog

mcdado:
npmlog has the advantage of using the same log level configuration of npm.

Running `npm install gulp-sass --no-progress --silent` I found out that this package is particularly noisy.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -48,7 +48,7 @@ function download(url, dest, cb) {
     return response.statusCode >= 200 && response.statusCode < 300;
   };
 
-  log.http('node-sass install', 'Start downloading binary at', url);
+  log.http('node-sass install', 'Start downloading binary at' + url);
 
   try {
     request(url, downloadOptions(), function(err, response) {
@@ -108,7 +108,7 @@ function checkAndDownloadBinary() {
         return;
       }
 
-      log.info('node-sass install', 'Binary downloaded and installed at', sass.getBinaryPath());
+      log.info('node-sass install', 'Binary downloaded and installed at' + sass.getBinaryPath());
     });
   });
 }
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -48,7 +48,7 @@ function download(url, dest, cb) {
     return response.statusCode >= 200 && response.statusCode < 300;
   };
 
-  console.log('Start downloading binary at', url);
+  log.http('Start downloading binary at', url);
 
   try {
     request(url, downloadOptions(), function(err, response) {
@@ -98,17 +98,17 @@ function checkAndDownloadBinary() {
 
   mkdir(path.dirname(sass.getBinaryPath()), function(err) {
     if (err) {
-      console.error(err);
+      log.error(err);
       return;
     }
 
     download(sass.getBinaryUrl(), sass.getBinaryPath(), function(err) {
       if (err) {
-        console.error(err);
+        log.error(err);
         return;
       }
 
-      console.log('Binary downloaded and installed at', sass.getBinaryPath());
+      log.info('Binary downloaded and installed at', sass.getBinaryPath());
     });
   });
 }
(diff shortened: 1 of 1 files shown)


xzyfer:
Thanks @mcdado this seems sensible. I'll take a closer look this weekend.

mcdado:
FWIW: this only bugs me because I try to do `npm install --quiet --no-progress` in my deployment scripts, but with `node-sass` in the dependencies, the silencing options don't work 😅
I get a lot of repeated lines from the download progress of the binary. I don't know if I changed all of the console statements or not.

xzyfer:
That's interesting since the progress bar is already using npmlog
specifically so we can respect the no progress config. Have you got some
example output?

mcdado:
Sure, here you go: [link to video](https://cl.ly/0I0a3c0j273j)

mcdado:
In deployment I have this command: `npm install --quiet --no-progress --depth 0 1>/dev/null`

Note that I even redirect `stdout` to `/dev/null`!

But I get these lines:
````
[?25l[..................] - :
output [..................] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .................] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .................] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ................] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ...............] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ..............] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ..............] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .............] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ............] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ...........] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ...........] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ..........] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .........] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .........] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ........] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .......] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ......] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .....] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .....] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ....] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ...] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ..] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ..] - https://github.com/sass/node-sass/releases/download/v3.1
output [ .] - https://github.com/sass/node-sass/releases/download/v3.1
output [ ] - https://github.com/sass/node-sass/releases/download/v3.1
output [?25h

**Comment by xzyfer** [comment]
Interesting, looks like a bug with how either we or npm handles `--no-progress`

File: scripts/install.js [hunk 1]
```diff
@@ -48,7 +48,7 @@ function download(url, dest, cb) {
     return response.statusCode >= 200 && response.statusCode < 300;
   };
 
-  console.log('Start downloading binary at', url);
+  log.http('Start downloading binary at', url);
```
nschonni: I think this makes sense, but the messages need to be updated since the first parameter to the function in npmlog is the [`prefix`](https://www.npmjs.com/package/npmlog#loglevelprefix-message-) rather than just contating the series of objects
EX:
```js
log.http('install.js', 'Start downloading binary at ' + url)
```

mcdado:
I hope it's okay that I used `node-sass install`. In case I can change it.

scripts/install.js
```diff
@@ -48,7 +48,7 @@ function download(url, dest, cb) {
     return response.statusCode >= 200 && response.statusCode < 300;
   };
 
-  console.log('Start downloading binary at', url);
+  log.http('node-sass install', 'Start downloading binary at', url);
```
nschonni: The `, url` doesn't work the same way as it did with `console`, so it either needs a temaplate in the message string or just switch it to `+ url`

mcdado: Sorry, that slipped when I replaced the parameters! I'll change it now.

mcdado: Done in 182ff9d

PR #1815, opened 2016-11-23
Pass NODEDIR to node-gyp

mgedmin:
Sometimes (e.g. for offline builds) it's convenient to be able to pass a custom `--nodedir` to `node-gyp`.


mgedmin:
Actually I've just figured out I can set `npm_config_nodedir` in the environment and `node-gyp` interprets that, so this is unnecessary.

PR #1827, opened 2016-12-09
Removed top-level return statement from cli

### LINKED ISSUE #1826: Question: bin file is returning eventEmitter
Here: https://github.com/sass/node-sass/blob/dc92c18e1dffd4acbab69e76c4bcda238f52da27/bin/node-sass#L432

From my understanding, you shouldn't `return` outside of functions. What is the purpose of returning an event emitter? Does it have to do with piping? I can't find any docs regarding that.

bfred-it:
Looking at [the commit that introduced it,](https://github.com/sass/node-sass/commit/c7539f19d2e97cd52cf7a52c00c8e8fff3f4409a) it seems that it's leftover of the `cli.js` which was then required in a separate file.

Closes #1826

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -428,5 +428,3 @@ if (options.src) {
     run(options, emitter);
   });
 }
-
-return emitter;


xzyfer:
@bfred-it I believe you are correct. Thank you.

PR #1830, opened 2016-12-10
Drop dependency lodash.isarray

### LINKED ISSUE #1829: Dependency on lodash.isarray
I saw that 79e86f32ce3661569748164d26e2f2667a79699a introduced a dependency on [`lodash.isarray`](https://www.npmjs.com/package/lodash.isarray) among others.

Is that really necessary? Any particular reason to not use the standard `Array.isArray()`?

`Array.isArray()` already appears in other parts of the code and we're having this message on install:

```
warning node-sass > lodash.isarray@4.0.0: This package is deprecated. Use Array.isArray.
```

daltones:
Fixes #1829.

The code of `lodash.isarray` is actually this:
```js
var isArray = Array.isArray;

module.exports = isArray;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "in-publish": "^2.0.0",
     "lodash.assign": "^4.2.0",
     "lodash.clonedeep": "^4.3.2",
-    "lodash.isarray": "^4.0.0",
     "lodash.mergewith": "^4.6.0",
     "meow": "^3.7.0",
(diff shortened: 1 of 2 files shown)


xzyfer:
`Array.isArray` is not available until Node 4.6.2. We support back to Node 0.10.

xzyfer:
My apologies you are correct.

Issue #1823, opened 2016-12-07
Only partial data sent to process.stdout when output reaches a certain size


baseten:
- NPM version: 3.10.8
- Node version: v6.9.1
- Node Process: 
```
{ http_parser: '2.7.0',
  node: '6.9.1',
  v8: '5.1.281.84',
  uv: '1.9.1',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  icu: '57.1',
  modules: '48',
  openssl: '1.0.2j' }
```
- Node Platform: darwin:
- Node architecture: x64
- node-sass version:
```
node-sass	3.13.0	(Wrapper)	[JavaScript]
libsass  	3.3.6	(Sass Compiler)	[C/C++]
```
- npm node-sass versions: node-sass@3.13.0

When piping output to another script, if the compiled code is over a certain length then only the first chunk appears to be sent. From the command line all seems fine:

```
./node_modules/node-sass/bin/node-sass sass/app.scss
```

But when piping to another node script, the output is cut off:

```
./node_modules/node-sass/bin/node-sass sass/app.scss | scripts/test.js
```

Where scripts/test.js is the following:

```
#!/usr/bin/env node

process.stdin.on( 'readable', () => {
    var chunk = process.stdin.read();

    if ( chunk !== null ) {
        console.log( chunk.toString() ); // logs incomplete css
    }
} );

process.stdin.on( 'end', () => {
    console.log( 'end' );
} );
```

baseten:
I'm not sure if this is any use, but if I output the partial string to file rather than logging to console the filesize is 65,536 bytes

xzyfer:
My gut feeling is that this is probably either the max size of the node
buffer, and the C sass string data type.

We'll need to do some investigation.

marcusdarmstrong:
I ran into this issue a few days ago. Downgraded to 3.8.0 to eliminate the blockage, looked like it was related to a node.js change to no longer flush output buffers on exit unless explicitly told to.

I was also able to workaround this by inserting this snippet into the package as a local mod:
`[process.stdout, process.stderr].forEach(function(stream){
    if (stream._handle && stream._handle.setBlocking)
        stream._handle.setBlocking(true);
});`

xzyfer:
There were two separate issues in play. Firstly `console.log` has max buffer length that could be exceeded by some large CSS output, and we were exiting before fully flushing the output buffer.

This has been fixed #1834.

arunstan:
[...]

PR #1834, opened 2016-12-13
Fix output buffer is imcomplete working in child_process

ysugimoto:
When I send a large buffer, over 50KB's CSS string, working in `child_process.spawn`, `node-sass` outputs only partial strings.

Certainly, it is caused by nodejs's `console.log` max size.
https://github.com/nodejs/node/issues/9546

So I modified to send output by continuous partial strings, its asynchronous.

This PR will fix this issue: https://github.com/sass/node-sass/issues/1823.


bin/node-sass
```diff
@@ -161,11 +163,17 @@ function getEmitter() {
...
+    asyncWaiting = true;
+
+    // In nodejs <= 7.1.0, console.log cannot write output larger than 8275 bytes data.
+    // So that, we output partial strings manually with some asynchronous way.
+    stdout(data);
   });
 
   emitter.on('done', function() {
-    if (!options.watch && !options.directory) {
+    if (!options.watch && !options.directory && !asyncWaiting) {
```
xzyfer: isn't `asyncWaiting` always true now?

xzyfer:
https://www.npmjs.com/package/stdout-stream seems like a better alternative than writing out own stdout stream lib.

bin/node-sass
ysugimoto: `asyncWaiting` is initialized on false when called `getEmitter()`.
https://github.com/sass/node-sass/pull/1834/files#diff-c19ab2481e2703fcf0d2808f4c762f63R147

I read project source code, emit `log` event only stdout case ( is it right? ).
So that, `asyncWaiting` set to be true only emitted `log` event.

Please advise if there are other concerns that you need to worry about. thanks!

ysugimoto:
I replaced out stdout lib to suggested package, it also works fine!

ysugimoto:
It works fine, but Travis CI is down for some testings...

bin/node-sass
xzyfer: I don't think this is right. I believe we always want to exit when emitter.log is called. This is what happens when you call the node-sass cli. If I understand correctly all we need to do is

```js
emitter.on('log', stdout.write);
```

Am I missing something?

xzyfer:
I'm sorry the issue here is not clear. 

Either the problem is related to `console.log`s max buffer size, in which we can fix by

```
emitter.on('log', stdout.write)
```

It's because `process.stdout` is no longer blocking, in which case we can use https://github.com/isaacs/block-stream. Either way i don't think this is the correct solution.

ysugimoto:
Yes, I think so.

However, `process.stdout` is no longer blocking, then we stop `process.exit` until stdout stream output completely. It's called by `emitter.emit('done')` immediately, So I add the flag of `asyncWaiting`.

xzyfer:
OK I did some digging. You're correct that the issue is due how [process.exit](https://nodejs.org/api/process.html#process_process_exit_code) works.

> It is important to note that calling `process.exit()` will force the process to exit as quickly as possible _even if there are still asynchronous operations pending_ that have not yet completed fully, _including_ I/O operations to `process.stdout` and `process.stderr`.

I see ways to solve this:
- remove the `process.exit` (I don't think we need it)
- use [exit](https://www.npmjs.com/package/exit) instead

ysugimoto:
Thank you for your advice. I replaced to `process.exit` to `exit` package.

And I removed exiting in `emitter.on('done', ...)` callback because of process exited even though using `stdout-stream` package, I wondering...
After removed this point, It works fine: CLI, spawned process, and watch mode.

bin/node-sass
```diff
@@ -161,13 +163,12 @@ function getEmitter() {
   });
 
   emitter.on('log', function(data) {
-    console.log(data);
+    // In nodejs <= 7.1.0, console.log cannot write output larger than 8275 bytes data.
+    stdout.write(data);
   });
 
   emitter.on('done', function() {
```
xzyfer: Please delete the `emitter.on('done'` block all together.

ysugimoto:
https://github.com/sass/node-sass/pull/1834/files#diff-c19ab2481e2703fcf0d2808f4c762f63R165

Use `bind()` because of avoid error on `stdout-stream`.

package.json
```diff
@@ -56,6 +56,7 @@
     "async-foreach": "^0.1.3",
     "chalk": "^1.1.1",
     "cross-spawn": "^3.0.0",
+    "exit": "^0.1.2",
```
xzyfer: Please remove.

lib/render.js
```diff
@@ -47,21 +47,14 @@ module.exports = function(options, emitter) {
   var stdin = options.stdin;
 
   var success = function(result) {
-    var todo = 1;
```
xzyfer: The `done` still needs to be fired. It's used in other places. Please undo these changes and just remove 

```
emitter.on('done', function() {		
    // noop		
});
```

xzyfer: The `done` still needs to be fired. It's used in other places. Please undo these changes and just remove 

```
emitter.on('done', function() {		
    // noop		
});
```

xzyfer:
Thank you for your work and patience. I'll merge this when CI is happy.

PR #1836, opened 2016-12-16
Add a platform variant for installing a musl binary

lox:
This adds a new element `platformvariant` to the naming scheme for precompiled binding binaries:

```
{platform}_{platformvariant}-{arch}-{abiversion}_binding
```

This allows for statically compiled musl binaries to be made available for Alpine Linux.

The check for whether this is needed is done by detecting symbols in the node binary, as suggested in https://github.com/sass/node-sass/issues/1589#issuecomment-266470069. I don't love this approach, but it's a whole lot simpler than `exec`'ing out to `ldd`. 

See https://github.com/sass/node-sass/issues/1589 for more details on background.

Tests:

```bash 
$ docker run -v "$PWD:/code" --rm -w /code node:7-alpine node scripts/install.js 
Downloading binary from https://github.com/sass/node-sass/releases/download/v4.0.0/linux_musl-x64-51_binding.node 
```

```bash 
$ docker run -v "$PWD:/code" --rm -w /code node:7 node scripts/install.js 
Downloading binary from https://github.com/sass/node-sass/releases/download/v4.0.0/linux-x64-51_binding.node
```

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -182,8 +182,14 @@ function getBinaryName() {
   } else if (pkg.nodeSassConfig && pkg.nodeSassConfig.binaryName) {
     binaryName = pkg.nodeSassConfig.binaryName;
   } else {
+    var platform = process.platform;
+    var variant = getPlatformVariant();
+    if (variant !== "") {
+      platform += "_" + variant;
+    }
+
     binaryName = [
-      process.platform, '-',
+      platform, '-',
       process.arch, '-',
       process.versions.modules
     ].join('');
@@ -357,6 +363,23 @@ function getVersionInfo(binding) {
     ['node-sass', pkg.version, '(Wrapper)', '[JavaScript]'].join('\t'),
     ['libsass  ', binding.libsassVersion(), '(Sass Compiler)', '[C/C++]'].join('\t'),
   ].join(eol);
+}
+
+/**
+ * Gets the platform variant, either an empty string for non-Linux systems or the
+ * libc implementation (glibc|musl) on Linux.
+ *
+ * @api public
+ */
+
+function getPlatformVariant() {
+  if (process.platform != "linux") {
+    return "";
+  }
+  if (fs.readFileSync(process.execPath).indexOf('libc.musl-x86_64.so.1')) {
+    return "musl";
+  }
+  return "glibc";
 }
 
 module.exports.hasBinary = hasBinary;
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -381,7 +381,14 @@ function getPlatformVariant() {
   }
 
   try {
-    contents = fs.readFileSync(process.execPath).toString();
+    contents = fs.readFileSync(process.execPath);
+
+    // Buffer.indexOf was added in v1.5.0 so cast to string for old node
+    // Delay contents.toStrings because it's expensive
+    if (!contents.indexOf) {
+      contents = contents.toString();
+    }
+
     if (contents.indexOf('libc.musl-x86_64.so.1') !== -1) {
       return 'musl';
     }
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -172,7 +172,9 @@ function getArgument(name, args) {
  */
 
 function getBinaryName() {
-  var binaryName;
+  var binaryName,
+    variant,
+    platform = process.platform;
 
   if (getArgument('--sass-binary-name')) {
     binaryName = getArgument('--sass-binary-name');
@@ -183,9 +185,8 @@ function getBinaryName() {
   } else if (pkg.nodeSassConfig && pkg.nodeSassConfig.binaryName) {
     binaryName = pkg.nodeSassConfig.binaryName;
   } else {
-    var platform = process.platform;
-    var variant = getPlatformVariant();
-    if (variant !== '') {
+    variant = getPlatformVariant();
+    if (variant) {
       platform += '_' + variant;
     }
 
@@ -373,10 +374,12 @@ function getVersionInfo(binding) {
  */
 
 function getPlatformVariant() {
+  var contents = '';
+
   if (process.platform !== 'linux') {
     return '';
   }
-  var contents = '';
+
   try {
     contents = fs.readFileSync(process.execPath);
     if (!contents.indexOf) {
@@ -386,6 +389,7 @@ function getPlatformVariant() {
       return 'musl';
     }
   } catch (err) { } // eslint-disable-line no-empty
+
   return '';
 }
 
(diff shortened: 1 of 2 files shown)


lib/extensions.js
```diff
@@ -359,6 +370,23 @@ function getVersionInfo(binding) {
...
+ */
+
+function getPlatformVariant() {
+  if (process.platform != 'linux') {
+    return '';
+  }
+  if (fs.readFileSync(process.execPath).indexOf('libc.musl-x86_64.so.1') != -1) {
+    return 'musl';
+  }
+  return 'glibc';
```
nschonni: Probably better to just return empty unless it's musl. Would simplify the code above

lib/extensions.js
```diff
@@ -359,6 +365,22 @@ function getVersionInfo(binding) {
...
+ * Gets the platform variant, currently either an empty string or 'musl' for Linux/musl platforms.
+ *
+ * @api public
+ */
+
+function getPlatformVariant() {
+  if (process.platform !== 'linux') {
+    return '';
+  }
+  if (fs.readFileSync(process.execPath).indexOf('libc.musl-x86_64.so.1') !== -1) {
```
xzyfer: `fs.readFileSync` can throw

```js
var contents = '';
try {
  contents = fs.readFileSync('foo.bar');
  if (contents.indexOf('libc.musl-x86_64.so.1') !== -1) {
    return 'musl';
  }
} catch (err) { }
return ''
```

lox: Thanks, both excellent points. Addressing.

lib/extensions.js
```diff
@@ -375,10 +375,14 @@ function getPlatformVariant() {
...
     return '';
   }
-  if (fs.readFileSync(process.execPath).indexOf('libc.musl-x86_64.so.1') !== -1) {
-    return 'musl';
-  }
-  return '';
+  var contents = '';
+  try {
+    contents = fs.readFileSync('foo.bar').toString();
+    if (contents.indexOf('libc.musl-x86_64.so.1') !== -1) {
```
S-YOU: sorry I sent comment while you are updating this.
toString is slow on where Buffer.indexOf is available. so this would be longer version.

```
try {
  var muslName = 'libc.musl-x86_64.so.1';
  var contents = fs.readFileSync(process.execPath);
  var hasMusl = contents.indexOf ? contents.indexOf(muslName) > 0 : contents.toString().indexOf(muslName) > 0;
  if (hasMusl) {
    return 'musl';
  }
} catch (err) { console.log (err); }
  return '';
}
```

lox: Starts to get a bit micro-optimization-y for my tastes, but sure.

lox:
This last failure has me baffled.

S-YOU:
I think this test has timeout.

https://github.com/sass/sass-spec/blob/master/spec/libsass-closed-issues/issue_657/default/input.scss

lib/extensions.js
```diff
@@ -359,6 +365,29 @@ function getVersionInfo(binding) {
...
+  if (process.platform !== 'linux') {
+    return '';
+  }
+  var contents = '';
+  try {
+    contents = fs.readFileSync('foo.bar');
+    if (!contents.indexOf) {
+      contents = contents.toString();
+    }
+    if (contents.indexOf('libc.musl-x86_64.so.1') !== -1) {
```
shouze: ping @ncopa can you confirm it's ok to cover all Alpine releases or should we look for a wider pattern?

shouze:
LGTM, thx @lox, asked @ncopa to validate the musl check ;)

lib/extensions.js
lox: Are you asking whether this test will work for all Alpine releases? If so, yes.

lib/extensions.js
```diff
@@ -359,6 +365,29 @@ function getVersionInfo(binding) {
...
+ * @api public
+ */
+
+function getPlatformVariant() {
+  if (process.platform !== 'linux') {
+    return '';
+  }
+  var contents = '';
+  try {
+    contents = fs.readFileSync('foo.bar');
```
S-YOU: Ah, its not foo.bar, its `process.execPath`

gmaliar:
Hey guys, I've PR'ed on top of @lox because I've hit a few more issues down the road with the `linux_musl` variant.

@lox can merge it in https://github.com/lox/node-sass/pull/3 and it's passing CI.

https://travis-ci.org/gmaliar/node-sass/builds/184758295

lox:
Thanks @gmaliar!

lib/extensions.js
lox: So much fail.

lib/extensions.js
```diff
@@ -359,6 +366,29 @@ function getVersionInfo(binding) {
...
+ */
+
+function getPlatformVariant() {
+  if (process.platform !== 'linux') {
+    return '';
+  }
+  var contents = '';
+  try {
+    contents = fs.readFileSync(process.execPath);
+    if (!contents.indexOf) {
```
xzyfer: Were you able to figure out how contents wasn't a string or buffer? This intent of line isn't clear.

xzyfer:
Pushed some minor cleanups. Happy :ship: this when CI is green.

lib/extensions.js
S-YOU: just tried on 0.10. 0.12 and 7, they all return Buffer, indexOf is not available in 0.10 and 0.12

```
%docker run --rm -it mhart/alpine-node:0.12 sh                                                                    
# node
> fs.readFileSync(process.execPath)
<Buffer 7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 03 00 3e 00 01 00 00 00 43 b3 2a 00 00 00 00 00 40 00 00 00 00 00 00 00 78 64 fa 00 00 00 00 00 00 00 ... >
> fs.readFileSync(process.execPath).indexOf
undefined
```

```
%docker run --rm -it mhart/alpine-node:7 sh   
# node
> fs.readFileSync(process.execPath)
<Buffer 7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 03 00 3e 00 01 00 00 00 11 03 57 00 00 00 00 00 40 00 00 00 00 00 00 00 f8 1d 26 02 00 00 00 00 00 00 ... >
> fs.readFileSync(process.execPath).indexOf
[Function: indexOf]
```

S-YOU: and performance difference on node 7.0
```
> console.time('find');fs.readFileSync(process.execPath).indexOf('libc.musl-x86_64.so.1') !== -1;console.timeEnd('find') 
find: 15.800ms

> console.time('find');fs.readFileSync(process.execPath).toString().indexOf('libc.musl-x86_64.so.1') !== -1;console.timeEnd('find')
find: 325.538ms
```

Also I think toString will use additional memory like 20M binary x UTF16 = 40M, If I am not wrong.

xzyfer: Thanks for the clarification @S-YOU. I've gone back to lazily tostring'ing.

PR #1896, opened 2017-02-13
Watcher should track newly created files

### LINKED ISSUE #1891: watching newly created files
Greetings!
When I start node-sass in recursive watching mode it works just fine with the exception that newly created files are not being watched.
For example, if I wach a directory and I add new _styles.scss_ file, node-sass ignores this file, providing the process has already been started before. 
Is it possible to change this behavior? Or do I just miss something?

xzyfer:
Currently newly created are added to the graph but not added the
watcher.

Fixes #1891


xzyfer:
I've digging into the watcher tests. I don't think I've addressed all the flakiness but I'm confident I can now address the flaky specs as we notice them.

I've reenabled them as part of this PR.

xzyfer:
Some watcher tests still appear to be flakey. I've done a little more hardening. I'll re-disable any particularly troublesome specs until I have a better solution. Keen to get this out soon.

gaearon:
Thanks! Sorry that getting it through seems frustrating 😞 Hope you’ll figure it out!

gaearon:
Is there anything here that community could help you with?

xzyfer:
Sorry, I bit off more than I could chew with my time. I'll double check
this works and ship it.

xzyfer:
I think I'm really close on this now. Should hopefully land this weekend.

xzyfer:
Tacking this in #2020. Should land any day now.

PR #1902, opened 2017-02-16
Add binding.js to Coveralls coverage

nschonni:
Notice the Travis post build was actually failing (doesn't fail the build) because the binding.js coverage file was missing.


xzyfer:
Nice catch

PR #1915, opened 2017-02-24
Update request version to avoid npm warn

arzafran:
## Background
_request 2.78.0 uses deprecated node-uuid, it has been updated in version 2.79.0 to use uuid package, an unwanted warn is shown in the console when npm installing node-sass_
## Changes done
_bump request dependency version to 2.79.0_

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "nan": "^2.3.2",
     "node-gyp": "^3.3.1",
     "npmlog": "^4.0.0",
-    "request": "^2.79.1",
+    "request": "^2.79.0",
     "sass-graph": "^2.1.1",
     "stdout-stream": "^1.4.0"
   },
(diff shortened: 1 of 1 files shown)


nschonni:
Because of the `^`, that is the version that should be installed anyway. I do understand trying to bump it up though to avoid a known issue

arzafran:
@nschonni It should work like that, but in several environments at my office we had issues when installing node-sass by npm install and it's a really minor change that would solve this warn.

danez:
It can sometimes happen that an other dependency depends on `request` version `<2.79 && >2.61`. And if this dependency gets resolved first, npm will use the older version of `request` also for node-sass as it also satisfies the dependency `^2.61`

xzyfer:
Apologies, I've been away. I agree with @nschonni that this isn't technically necessary but I also see how much people react to these messages - https://github.com/strongloop/fsevents/pull/163

Thanks @arzafran

Issue #1838, opened 2016-12-16
Visual Studio 2013 no longer available for download


cheapsteak:
This link now only contains download for vs 2017:
https://www.visualstudio.com/downloads/download-visual-studio-vs#d-express-windows-desktop

Should the readme be updated?
What would be the method for building node-sass on windows now?

nschonni:
I think it should probibly point to the node-gyp instructions since that is really what we require https://github.com/nodejs/node-gyp#installation. That said I want to say there was an issue at one point with the optimizations in 2015 that didn't play nice with XP?

saper:
@nschonni the reason we insisted on vs2013 was that the official node builds were build with vs2013 at the time. Mixing C++ runtime libs across the module border is never guarantee to work. Our XP users got hit with some particular difference in the way a thread local storage got initialized.

cheapsteak:
What do you think about replacing it with this?
`npm install --global --production windows-build-tools node-gyp`

Referenced from [node-gyp's instructions for windows](https://github.com/nodejs/node-gyp#installation), which references [Microsofts' NodeJS Guidelines](https://github.com/Microsoft/nodejs-guidelines/blob/master/windows-environment.md#environment-setup-and-configuration)

nschonni:
Maybe a cleaned up version of this? https://github.com/laverdet/node-fibers#from-source
It's probably good to look around at some of the native packages and see how they explain it.

xzyfer:
I'm all for recommending `windows-build-tools` over VS2013 if we can verify the compiled binary works as expected.

saper:
Good idea, we only need to check it can be cached (so we do not download it every time) and that my fancy PDB debugging support works :)

PR #1921, opened 2017-03-12
Update Link: Create a link to correct VS

hweeks:
I've updated the link to Visual Studio 2013 WD so that it is accurate. The old link now points to a more recent version.


nschonni:
Thanks, we may defer to the node-gyp instructions as discussed in #1838, but it's better we don't leave it in a broken state 😄

Issue #1758, opened 2016-10-24
Inconsistent behavior when using directory input with --watch and without --watch


herophuong:
**Current behavior**

- When using directory input without --watch, all .scss files will be compiled and *normal css files won't be compiled*.

- When using directory input with --watch, when *any css file changed, they will be compiled*.

**Suggested behavior**

When using directory input with --watch, we should watch .scss file only.

--- 

- NPM version (`npm -v`): 3.10.3
- Node version (`node -v`): 6.3.1
- node-sass version (`node -p "require('node-sass').info"`): 3.10.1

wuzhouyang:
resolved?

xzyfer:
if the issue still open it's not resolved.

nschonni:
Only watching SASS/SCSS files appears to be in the current code https://github.com/sass/node-sass/blob/master/bin/node-sass#L135

nschonni:
Nevermind, it looks like the watcher does look for `css` here https://github.com/sass/node-sass/blob/master/bin/node-sass#L243

xzyfer:
This looks like a bug I introduced in https://github.com/sass/node-sass/pull/1078. We should not be watching .css files.

kibbled:
Is there a workaround for this bug?

eugeneglova:
hey @xzyfer could you just make these [extensions](https://github.com/sass/node-sass/blob/master/bin/node-sass#L236) to be configurable so I can remove `css` from that list for as a workaround for this issue. Should I create PR for that?

xzyfer:
They cannot be configured. They live in LibSass. The decisions has been made to keep the current behaviour until Sass 4's module system is ready. Removing this now will break lots of people's code.

kibbled:
This bug sort of causes infinite compiles to happen when you watch a directory because after SCSS files compile to CSS, the watcher detects changes and recompiles the CSS which starts a vicious cycle Perhaps a recommendation in the documentation for using a different directory for your CSS files from your SCSS files since it's unusable in the current form.

RoyTinker:
@kibbled Perhaps using nodemon to watch & trigger node-sass in an npm script would be an acceptable workaround.

RoyTinker:
Looks like the best workaround is to use node-sass-chokidar.

xzyfer:
So I think the reasonable workaround for this is for node-sass to not treat .css as entry points.

[...]

PR #1925, opened 2017-03-15
Fix watching of CSS files to avoid infinite build cycle

Nateowami:
Previously when the src and dest dirs were the same, whenever a CSS file changed (either manually or as the result of a build) it was recursively recompiled in an infinite loop, unless a build finished in less than
500ms (Gaze's default file watching debounce). Now, when the input and output dirs are the same, CSS files are treated like includes (files starting with an underscore) in that they are never built into CSS, but
they are watched in case they are included in a Sass file.

When the input and output dirs differ, functionality is unchanged.

I'm not sure the best way to go about writing tests for this. The infinite cycle only starts when builds takes longer than Gaze's default debounce timeout of 500ms (which is happening in a project I'm working on). A small logic bug in Gaze ([incorrect operator short-circuiting](https://github.com/shama/gaze/blob/8db2a1ec30b75167e96fc24902ae57c4a2fbba67/lib/gaze.js#L43)) prevents setting the timeout to 0 (poking into `node_modules` is how I demonstrated this issue while testing my changes). The one symptom you could test for is that if you modify a CSS file node-sass will process it and reformat it.

Currently, for reasons unknown to me, all the watch tests are being skipped (except the check that node-sass doesn't immediately exit). This causes me think the authors have been unable to get the watch tests passing. As someone unfamiliar with the project I would most likely be unable to get watch tests passing either.

This is related to #1758 (Inconsistent behavior when using directory input with --watch and without --watch). It would make the most sense to not compile CSS files at all (or at least make it consistent with and without --watch). As, @xzyfer points out though, changing it now could break builds that rely on node-sass compiling CSS files to another directory. However, this change only affects the scenario where the input directory is the same as the output directory, and therefore should not cause a problem.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -274,7 +274,8 @@ function watch(options, emitter) {
       }
     });
     files.forEach(function(file) {
-      if (path.basename(file)[0] !== '_') {
+      var fileName = path.basename(file);
+      if (fileName[0] !== '_' && (!options.sameInputAsOutput || fileName.slice(-4) !== '.css')) {
         renderFile(file, options, emitter);
       }
     });
@@ -309,6 +310,8 @@ function run(options, emitter) {
     if (!isDirectory(options.output)) {
       emitter.emit('error', 'An output directory must be specified when compiling a directory');
     }
+
+    options.sameInputAsOutput = path.resolve(options.directory) === path.resolve(options.output);
   }
 
   if (options.sourceMapOriginal && options.directory && !isDirectory(options.sourceMapOriginal) && options.sourceMapOriginal !== 'true') {


xzyfer:
I appreciate you taking the time, however we cannot accept this change.

This is a significant breaking change for node-sass users. Additionally causes node-sass to differ from other LibSass bindings i.e. sassc, libsass-python, perl-sass etc..

This behaviour is due to a known bug in LibSass. The bug, although easy to fix, cannot be patched because it would unnecessarily disrupt the community. Sass 4 will introduce first class modules that will allow us to support importing of CSS files in an official way.

For now our advice for now is to not use the same directory for input and output. Alternative explicitly add a `.scss` extension to your `@import`s.

Nateowami:
Thanks for taking the time to look at this.

I wonder though whether I may not have explained the changes clearly, as I do not believe this is a breaking change. I tried to track down the bug you mentioned, but I'm not sure I found it. Perhaps sass/libsass#2096, or #1758?

In this PR, the behavior of watching CSS files remains unchanged. They must be watched because they can be included into Sass files. The only thing this changes is that when watching files, if the input dir is the same as the output dir, CSS files will not be built on top of themselves recursively. They'll still be watched, and can be included into other files. They become more like `_include` files (though only when watching, and only when input and output are identical directories—not in any other case). This is the same issue @kibbled [mentioned at the end of #1758.](https://github.com/sass/node-sass/issues/1758#issuecomment-258744171)

Livven:
@xzyfer Is this documented anywhere? This is a very surprising issue that is very hard to search for, e.g. [these people over at `create-react-app`](https://github.com/facebookincubator/create-react-app/issues/1939) don't seem to be aware that it is indeed a known bug.

RoyTinker:
AFAIK, the current best workaround for this issue is to use https://github.com/michaelwayman/node-sass-chokidar

PR #1958, opened 2017-04-29
Bump sass-graph@2.2.4

xzyfer:
- fixes issues with transitive dependency scss-tokenizer
- allows graph to follow symlinks


xzyfer:
Blocked on me not being on a plane

PR #1969, opened 2017-05-16
Update module version for Node 8

xzyfer:
Was 54 is now 55 will soon be 57.

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -70,6 +70,7 @@ function getHumanNodeVersion(abi) {
     case 48: return 'Node.js 6.x';
     case 51: return 'Node.js 7.x';
     case 55: return 'Node.js 8.x';
+    case 57: return 'Node.js 8.x';
     default: return false;
   }
 }
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -69,7 +69,7 @@ function getHumanNodeVersion(abi) {
     case 47: return 'Node.js 5.x';
     case 48: return 'Node.js 6.x';
     case 51: return 'Node.js 7.x';
-    case 54: return 'Node.js 8.x';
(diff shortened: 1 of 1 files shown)


lib/extensions.js
```diff
@@ -69,7 +69,8 @@ function getHumanNodeVersion(abi) {
     case 47: return 'Node.js 5.x';
     case 48: return 'Node.js 6.x';
     case 51: return 'Node.js 7.x';
-    case 54: return 'Node.js 8.x';
+    case 55: return 'Node.js 8.x';
```
nschonni: Minor, but we chould just fall through here

xzyfer: True. I did them as discrete commits so I didn't think about it. I want to update this handling so that it only shows as a _helpful_ warning rather than erring out.

xzyfer:
Context: https://github.com/nodejs/citgm/issues/389

lib/extensions.js
suryapratap: can't we use a dictionary lookup instead of a switch?

```js
function getHumanNodeVersion(abi) {
	var humanNodeVersions={
	   11:'Node 0.10.x',
     14:'Node 0.12.x',
     42:'io.js 1.x',
     43:'io.js 1.1.x',
     44:'io.js 2.x',
     45:'io.js 3.x',
     46:'Node.js 4.x',
     47:'Node.js 5.x',
     48:'Node.js 6.x',
     51:'Node.js 7.x',
     55:'Node.js 8.x',
     57:'Node.js 8.x'
	}
	var version = parseInt(abi || process.versions.modules, 10);
	return humanNodeVersions[version]||false;
}
```

xzyfer: We could, but I prefer the switch.

PR #1977, opened 2017-05-26
Build: Use Travis Stages

nschonni:
This is still beta, but I thought I'd give it a try.
Instead of running the full matrix at once, run the mainline node release linting and coverage, and then run the rest if that is ok.
I think in the end there could probably be a "docker" phase to maybe run some alpine stuff in the future. It could also be used to orchistrate the releases, although there would need to be some sync/handshake with AppVeyor

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,40 @@
 language: node_js
-node_js:
-  - "7"
-  - "6"
-  - "4"
-  - "0.12"
-  - "0.10"
-  - "iojs"
 
 compiler: gcc
 sudo: false
-
-os:
-  - linux
-  - osx
 
 env:
   global:
     - SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
 
-matrix:
-  fast_finish: true
-  exclude:
-    - node_js: iojs
-    - node_js: "0.10"
+jobs:
+  include:
+    - stage: test
+      node_js: "node"
+      os: linux
+      before_script: npm run lint || exit 1;
+      after_success: npm run-script coverage;
+    - stage: platform-test
+      node_js: "node"
       os: osx
-    - node_js: "0.12"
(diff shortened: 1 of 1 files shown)


nschonni:
I also removed "iojs" since it was already excluded from the matrix

xzyfer:
This is cool, nice one!

.travis.yml
```diff
@@ -1,31 +1,40 @@
...
 
-matrix:
-  fast_finish: true
-  exclude:
-    - node_js: iojs
-    - node_js: "0.10"
+jobs:
+  include:
+    - stage: test
+      node_js: "7"
```
xzyfer: Is it possible to use some kind of "latest" or "current" tag so we don't have to change this for node 8 etc..?

nschonni: Looks like there is a "node" key https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions
We would still have to change for Node 8 since we would need to add it to the matrix. I do like this idea though

xzyfer:
True we would still need to change it. At some point in the future we'll
only support LTS and current so using the node key is nice because it'll
always be the current.

nschonni:
Yeah, it would be good to see if the "LTS" could be used too. It looks like NVM does have that concept https://github.com/creationix/nvm#long-term-support

nschonni:
I swapped the `6` for `lts/boron` and `4` with `lts/argon`. It looks like we could use`lts/*` in that future case where you only want the latest & LTS though.

xzyfer:
Nice find. I'm still traveling but sounds good to me

nschonni:
I think landing this can wait a bit, just to see if Travis changes anything before it gets out of beta.

nschonni:
I'm just going to land this and then add in 8 separately

PR #1999, opened 2017-06-02
Add Electron ABI versions

jhermsmeier:
This adds the respective Electron ABI versions to `getHumanNodeVersion()`
to enable running `node-sass` under Electron where the ABI versions differ from Node's.


jhermsmeier:
@saper that was quick, thanks!

nschonni:
Think there should be something in here that warns users that we still don't support Electron and the binary download will still fail

Issue #636, opened 2015-01-22
Migrate to a better file watcher


paulmillr:
Hey. Your file watcher [(gaze)](https://github.com/shama/gaze/issues) has 53 opened issues. Many bugs on various platforms.

Can we make a pull request that migrates the watching to [chokidar](https://github.com/paulmillr/chokidar)? Chokidar is ultra-stable and is used in many popular projects.

kevva:
+1, I basically chose gaze because it's widely used in gulp and seems to work there. But I'm all for a change.

paulmillr:
Gulp is switching to chokidar ;) https://github.com/wearefractal/glob-watcher/pull/13

@es128

am11:
:+1: Lets make it a post-v2 (stable) change.

BTW, nice name for a watcher library: Chokidar (watchman in my lang) :p

sindresorhus:
:+1:

callumlocke:
FWIW, I've worked a lot with gaze, chokidar and [sane](https://github.com/amasad/sane). sane is the newest and least well known, but it is by far the most reliable in my experience. And I think it has the most sensible API. Worth a look.

am11:
@callumlocke, thanks for the suggestion. Will take a look.

jareware:
:+1: would really like this!

tuhoojabotti:
:+1: Gaze is using ancient deprecated dependencies.

saper:
Some help is needed to fix the watcher tests, we had to disable them because the way the CLI was called didn't really work.

adambuczynski:
+1 to swtich to chokidar, or at least update the gaze dependency, as the current  version used by node-sass contains very outdated deps:

```
├─┬ gulp-sass@2.2.0
│ └─┬ node-sass@3.4.2
│   ├─┬ gaze@0.5.2
│   │ └─┬ globule@0.1.0
│   │   └── lodash@1.0.2 
```

cody-greene:
Why does node-sass even include a `--watch` option? It just seems out of scope. In fact, other tools [including but not limited to] webpack need to set up their own file watchers to properly trigger auto reloading and hot swapping. Including this functionality in node-sass increases the surface area for bugs, , expands the required API and documentation, and inflates the dependency tree. And now the dependencies that `gaze` is using are printing multiple warnings because they're so old.

Why not improve the focus of this package and remove the `--watch` option entirely?

callumlocke:
[...]

Issue #1056, opened 2015-08-07
Exposing watch in API


baileyparker:
I've read through the issues about the `--watch` flag and appreciate the amount of effort that has been put into maintaining it! :smile: The [implementation](https://github.com/sass/node-sass/blob/34db870992bcae17de30dd0e1f0fd8ddb4bcc363/bin/node-sass#L217-L256) is impressively terse and as I understand it (the PR mentioning this was merged), has recently been overhauled.

Would it be possible to expose this in the Node API so that it could be invoked programmatically (from a task runner like Gulp or Grunt, or just in plain Node) through something like `sass.watch('./file.scss')`?

In the case of task runners, this could allow allow for an evented API that allows for additional build steps on recompile. Perhaps something like this:

    var sass = require('node-sass');

    sass.watch('./file.scss')
      .on('update', function(file) {  // where file is a Vinyl instance
        // Version image url() assets, for example
        file.contents = cssVersioner(file.contents);
      });

If this sounds reasonable, I could get started on a PR for it. Otherwise, I plan on creating a npm module for this.

SpaceK33z:
I would also like a way to use the watch functionality through an API. This has the big advantage that external build tools are not necessary if you want to have a callback when a file is compiled.

saper:
I think we should expose this. Maybe one day we will be able to [tell libsass which files need to be reloaded individually](https://github.com/sass/libsass/issues/1368).

schmod:
Agree, and I'd be willing to add hooks into [grunt-sass](https://github.com/sindresorhus/grunt-sass) once this is done.

monarchjhaveri:
would like this functionality.

niksy:
Any progress on this? It would be really nice to have this as part of Gulp/Grunt process.

DarkPark:
+1

a good example of implementation is webpack:

```js
var webpack = require("webpack");

// returns a Compiler instance
var compiler = webpack({
    // configuration
});

// regular build
compiler.run(function(err, stats) {
    // ...
});

// build + watch
compiler.watch({ // watch options:
    aggregateTimeout: 300, // wait so long for more changes
    poll: true // use polling instead of native watchers
    // pass a number to set the polling interval
}, function(err, stats) {
    // ...
});
```

genintho:
[...]

PR #2020, opened 2017-06-15
Decouple the graph and render logic from the fs watcher

### LINKED ISSUE #1891: watching newly created files

xzyfer:
This logic is all tightly coupled in the bin. This logic has been
notoriously impossible to test due to weird fs timing issues. By
extracting the actual logic we're now able to test it in isolation.

With this in place replacing Gaze (#636) becomes a viable option.
This PR not only massively increases our test coverage for the
watcher but also address a bunch of known edge cases i.e. orphaned
imports when a files is deleted.

Closes #1896
Fixes #1891

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -3,13 +3,13 @@
 var Emitter = require('events').EventEmitter,
   forEach = require('async-foreach').forEach,
   Gaze = require('gaze'),
-  grapher = require('sass-graph'),
   meow = require('meow'),
   util = require('util'),
   path = require('path'),
   glob = require('glob'),
   sass = require('../lib'),
   render = require('../lib/render'),
+  watcher = require('../lib/watcher'),
   stdout = require('stdout-stream'),
   stdin = require('get-stdin'),
   fs = require('fs');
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -75,12 +75,14 @@
   "devDependencies": {
     "coveralls": "^2.11.8",
     "eslint": "^3.4.0",
+    "fs-extra": "^0.30.0",
```
xzyfer: This is a super older version but it's the last version that supports Node 0.10.

xzyfer:
>Abstract all the gaze.xxx calls in bin/node-sass to functions in the watcher

Can you elaborate on how this differs from what I've done here?

>Extract the watcher to a new NPM module 

This is related to #1056. I'm open to it.

bin/node-sass
```diff
@@ -229,63 +229,41 @@ function getOptions(args, options) {
...
-    };
-
-    if (options.directory) {
-      graph = grapher.parseDir(options.directory, graphOptions);
-    } else {
-      graph = grapher.parseFile(options.src, graphOptions);
-    }
+  var handler = function(files) {
+    files.added.forEach(function(file) {
+      var watch = gaze.watched();
```
nschonni: Abstract was probably the wrong word, but here is a crude example example
```js
    var watch = watcher.watched()
      if (watch.indexOf(file) === -1) {
        watcher.add(file);
      }
    });
```
So the `watcher` module would export certain functions required in the bin

nschonni:
The other part of splitting it would be a conditional `require`
```
var watcher;
if (option.watcher !== null)
  watcher = require(option.watcher)
 // error handling
} else {
  watcher = require('node-sass-watcher');
}

File: bin/node-sass (hunk 2 above)
Comment (xzyfer) [inline bin/node-sass]: Sorry I'm still not following. 

If I'm understanding correctly wouldn't that then couple us to the fs watcher implementation's api? 
What happens if the next fs watcher doesn't expose a method like `watch()` for exposing all currently watched files?

The `watch.indexOf(file) === -1` is an optimisation to prevent calling `gaze.add` superfluously. Under the covers `gaze.add` also does deduping but there's not guarantee the next fs watcher would.

File: bin/node-sass [hunk 3]
```diff
@@ -229,63 +229,41 @@ function getOptions(args, options) {
...
+      if (watch.indexOf(file) === -1) {
+        gaze.add(file);
+      }
+    });
 
-    return graph;
+    files.changed.forEach(function(file) {
+      if (path.basename(file)[0] !== '_') {
```
xzyfer: We could move this logic into `lib/watcher` but if we were to extract it into it's own package I can see uses cases for people wanting to poke changed partials i.e. `url` rewriting, or some kind webpack magic.

bin/node-sass
```diff
@@ -229,63 +229,41 @@ function getOptions(args, options) {
...
+        renderFile(file, options, emitter);
+      }
+    });
+
+    files.removed.forEach(function(file) {
```
xzyfer: This could be

```
files.removed.forEach(gaze.remove)
```

but it's not likely the next watcher would have the same API

nschonni:
I think this is probably fine to land as-is as it gets us testing.
Looking at the `bin` again, if we do a second pass, I think all of the gaze/watcher may just be pulled out by changing https://github.com/xzyfer/node-sass/blob/e81d6b9f225be4ea124a244d6e287a3711044e9f/bin/node-sass#L312
```js
   if (options.watch) {
     watcher.handle(options, emitter);
   } else if (options.directory) {
     renderDir(options, emitter);
   } else {
     render(options, emitter);
}
```
Although I may be missing an obvious problem with passing it that way

xzyfer:
Yeah there will be more passes. I also want to extract the console emiiter stuff because it's the cause of a lot of issues and feature request.

PR #2064, opened 2017-08-09
Fixes tapdriver shebang

naroga:
Adds the recommended shebang syntax -- the previous way breaks on sample AWS Elastic Beanstalk PHP environments.

### DIFF
--- a/src/libsass/script/tap-driver
+++ b/src/libsass/script/tap-driver
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/usr/bin/env sh
 # Copyright (C) 2011-2013 Free Software Foundation, Inc.
 #
 # This program is free software; you can redistribute it and/or modify


nschonni:
❤️ , thanks but all this code is from https://github.com/sass/libsass. We used to use a submodule that made that clearer, but had to in line it at the request of the NodeJS canary build folks.
Take a look at the upstream, but there has been a bunch of activity in that repo sync the last tag we sync'd

saper:
I can imagine doing it for pearl or even bash (BSDs putting them in /usr/local/bin), but who is not having /bin/sh ?

naroga:
@saper AWS's Elastic Beanstalk provides ready-to-use environments where `sh` exists, but is not located in `/bin`. Some sandboxed environments also load `sh` from elsewhere. This is not really uncommon.

@nschonni Thanks for the info :)

PR #2065, opened 2017-08-10
Fix static analysis logic errors, memory leaks, and dead stores.

krisselden:


nschonni:
❤️ , thanks but all this code is from https://github.com/sass/libsass. We used to use a submodule that made that clearer, but had to in line it at the request of the NodeJS canary build folks.
Take a look at the upstream, but there has been a bunch of activity in that repo sync the last tag we sync'd

krisselden:
doh

PR #2070, opened 2017-08-17
[WIP] Make sure libsass headers are read first

saper:
In the case libsass is installed system-wide
(e.g. the headers are in /usr/local/include) *and*
node requires reading headers from the same path
for the unbundled shared libraries like libuv,
make sure -I/usr/local/include will not take
precedence over our local bundles libsass headers.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -35,8 +35,9 @@
         'GCC_ENABLE_CPP_EXCEPTIONS': 'NO',
         'MACOSX_DEPLOYMENT_TARGET': '10.7'
       },
-      'include_dirs': [
+      'include_dirs+': [
         '<!(node -e "require(\'nan\')")',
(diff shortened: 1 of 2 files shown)


binding.gyp
```diff
@@ -35,8 +35,9 @@
         'GCC_ENABLE_CPP_EXCEPTIONS': 'NO',
         'MACOSX_DEPLOYMENT_TARGET': '10.7'
       },
-      'include_dirs': [
+      'include_dirs+': [
```
nschonni: Could/should this be done below in the `libsass_ext` condition? Not too familar with the gyp syntax though

nschonni:
@saper I just flipped this to Draft since it had "WIP" in the title. If it's good to go, feel free to flip it back

PR #2103, opened 2017-09-27
Bump request@2.83.0

jamesfoster85:
Updating request due to tough-cookie security fix
More info: https://github.com/request/request/pull/2776

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "nan": "^2.3.2",
     "node-gyp": "^3.3.1",
     "npmlog": "^4.0.0",
-    "request": "^2.79.0",
+    "request": "^2.83.0",
     "sass-graph": "^2.2.4",
     "stdout-stream": "^1.4.0"
   },


xzyfer:
Unfortunately this patch would effectively break node < 4 support.

jamesfoster85:
Understood.  I'll close this out.

acoard:
Hey, first off, love node-sass and have used it for years, thanks for all the hard work you all do.

That being said, I'm a little alarmed that using node-sass will now be introducing potential vulnerabilities to my application.

My understanding is that the vulnerability is not that serious with a small attack surface. [Node-sass only uses the request package to download binary files during the installation phase and no-where else](https://github.com/sass/node-sass/blob/cb8774f0ea47fb472625515df04c53ddc9c41623/scripts/install.js#L54).  As such, the attack vector would require compromising the url destination and injecting malicious data into the cookie causing the regex slowdown. Potentially this could make npm install fail, however, if the the attacker has compromised the server hosting the binaries they could make the install fail regardless, e.g. by returning null.

Can anyone confirm/deny this?  If I've made any mistakes please let me know. :)

Nonetheless in order to be compliant we have a strict 0 vulnerabilities policy. Even though this vulnerability seems very minor we still must fix it. The only solution I can think of is forking and breaking node < 4 support.  Any comments?

Thanks again,
Adam.

nschonni:
The `^` should already be installing the latest anyway, we just can't pin or set the minimum above the last version of the lib that will bread older Node. EX: if you install the package today you will install a non-vulnerable version already.

acoard:
Thanks for the response, you're entirely correct.  

For the record, it turns out that I'm getting different results if I use `yarn` or `npm install`.  When I use yarn, it winds up installing the problematic version but it's through another package.  I'll leave the details below in case it helps anyone else, but this definitely isn't anything node-sass is doing. 

**npm install**
```bash
$npm list tough-cookie
npm info it worked if it ends with ok
npm info using npm@5.3.0
npm info using node@v8.6.0
project@1.1.0 /PROJECTDIR/
└─┬ node-sass@3.13.1
  └─┬ request@2.83.0
    └── tough-cookie@2.3.3

npm info ok
```

Then in between, I `rm -R node_modules/` and all lock files.

**yarn**
```bash
project@1.1.0 /PROJECTDIR/
├─┬ karma@1.4.1
│ └─┬ chokidar@1.7.0
│   └─┬ fsevents@1.1.2
│     └─┬ node-pre-gyp@0.6.36
│       └─┬ request@2.81.0
│         └── tough-cookie@2.3.2 # PROBLEM!
└─┬ node-sass@3.13.1
  └─┬ request@2.83.0
    └── tough-cookie@2.3.3
```

Anyways, thanks again. I'll be digging deeper and maybe opening an issue for Yarn if I can get a test case.

PR #1726, opened 2016-09-22
render: emit err.formatted

cody-greene:
fixes #1508 

The when encountering an error CLI will now print:

```
Error: Undefined variable: \"$variable\".
        on line X of X
>> color: $variable;
          ^
```

instead of:

```
{
  "formatted": "Error: Undefined variable: \"$variable\".\n        on line X of X\n>> \n   ----^\n",
  "message": "Undefined variable: \"$variable\".",
  "column": 1,
  "line": 1,
  "file": "FILE_NAME",
  "status": 1
} 
```

PR #2104, opened 2017-09-28
use error.formatted at CLI error if it exists

shokai:
Since it is difficult to read the error in JSON, I changed the `error.formatted` property to output error.
If there is no `formatted`, it will be output in JSON as before.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -105,7 +105,7 @@ module.exports = function(options, emitter) {
   };
 
   var error = function(error) {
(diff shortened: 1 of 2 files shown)


shokai:
CI has fallen, but I can not fix it.
In node v0.10 there is a code using `const` in `node_modules/request/`

xzyfer:
Thanks @shokai. This is a duplicate of #1726 which is blocked because it's a semver major change,

Issue #2081, opened 2017-08-27
Cannot use relative paths for sass binary path


wittibs:
If I specify a sass_binary_path variable with a relative path in a local .npmrc file, the install/build process won't pick it up properly.

From lib/extensions.js:
```javascript
function hasBinary(binaryPath) {
  return fs.existsSync(binaryPath);
}
```

`fs.existsSync` checks for existence of directory/file based on the directory in which node process was started.

From lib/binding.js:
```javascript
module.exports = function(ext) {
  if (!ext.hasBinary(ext.getBinaryPath())) {
    if (!ext.isSupportedEnvironment()) {
      throw new Error(errors.unsupportedEnvironment());
    } else {
      throw new Error(errors.missingBinary());
    }
  }

  return require(ext.getBinaryPath());
};
```

The `require()` near the bottom tries to load the binary with respect from the current directory binding.js exists in. This is at odds with the `ext.hasBinary()` call above.

OS/platform info below.

- NPM version (`npm -v`): 5.3.0
- Node version (`node -v`): 8.4.0
- Node Process (`node -p process.versions`):
{ http_parser: '2.7.0',
  node: '8.4.0',
  v8: '6.0.286.52',
  uv: '1.13.1',
  zlib: '1.2.11',
  ares: '1.10.1-DEV',
  modules: '57',
  nghttp2: '1.22.0',
  openssl: '1.0.2l',
  icu: '59.1',
  unicode: '9.0',
  cldr: '31.0.1',
  tz: '2017b' }
- Node Platform (`node -p process.platform`): darwin
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`): 4.5.3
- npm node-sass versions (`npm ls node-sass`): 4.5.3

wittibs:
As I mentioned in PR #2108, I don't think this should be a feature. The implementation to this would be non-deterministic and hackish at best.

rmeshksar:
I noticed that when I am using sass_binary_path in the .npmrc file, it works with relative path as long as there is No underscore ( _ ) in the path. 

Maybe the following line is causing, however it should not be called when using sass_binary_path.
https://github.com/sass/node-sass/blob/master/lib/extensions.js#L273

For now I used relative path to the binary file inside package.json like this:

  "nodeSassConfig": {
    "binaryPath": "../../node_modules/node-sass-binary/vendor/node-sass/4.7.2/win32-x64-57_binding.node"
  }

[...]

Issue #2080, opened 2017-08-26
Binary CLI arguments not picked up properly during install


wittibs:
According to the docs [binary config section](https://github.com/sass/node-sass#binary-configuration-parameters), I should be able to pass command-line arguments to the install process, like so:

```
npm install node-sass --sass-binary-path=./custom-built-binaries/binding.node
```

However, the sass extensions script isn't coded to pick up CL arguments this way:

```javascript
function getArgument(name, args) {
  var flags = args || process.argv.slice(2),
    index = flags.lastIndexOf(name);

  if (index === -1 || index + 1 >= flags.length) {
    return null;
  }

  return flags[index + 1];
}
```

It expects another CL argument after the first flag, something like this:

```
npm install node-sass --sass-binary-path ./custom-built-binaries/binding.node
```

However that [won't work](https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script#answer-14404223), as all custom CL arguments must have '--' prefixed.

The `getArgument` function needs to be modified to properly pick up CL arguments (probably following the --argument=value pattern specified in the docs).
This was briefly mentioned once in issue #1955  [here](https://github.com/sass/node-sass/issues/1955#issuecomment-297710919) by @xzyfer, but I didn't see any other issue / PR mentioning this problem.

OS/platform info below. FWIW - also confirmed on Windows 7 64-bit machine.

- NPM version (`npm -v`): 5.3.0
- Node version (`node -v`): 8.4.0
- Node Process (`node -p process.versions`):
{ http_parser: '2.7.0',
  node: '8.4.0',
  v8: '6.0.286.52',
  uv: '1.13.1',
  zlib: '1.2.11',
  ares: '1.10.1-DEV',
  modules: '57',
  nghttp2: '1.22.0',
  openssl: '1.0.2l',
  icu: '59.1',
  unicode: '9.0',
  cldr: '31.0.1',
  tz: '2017b' }
- Node Platform (`node -p process.platform`): darwin
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`): 4.5.3
- npm node-sass versions (`npm ls node-sass`): 4.5.3

nschonni:
The `--sass-binary-path` shouldn't include the `binding.node` in it, it's the path to the folder containing the binding. I had to look through the code to figure this one out, so I think the docs need some clarity. The `--sass-binary-name` is used if you want to override the binding file name.

[...]

PR #2108, opened 2017-10-04
Fixes for #2080 and #2081

wittibs:
Tested changes for #2080 and #2081 by publishing to npm and installing/running the package locally (specifically with --sass-binary-path CLI flag).

Fix for #2081 would be really nice. I work at a big corporation, and we have firewalls that prevent easy direct access to downloading sass binaries (which causes us some headaches). Being able to point to a compiled binary, using a *relative path*, would help us a lot.

Thank you!

### DIFF
--- a/lib/binding.js
+++ b/lib/binding.js
@@ -17,6 +17,8 @@ module.exports = function(ext) {
       throw new Error(errors.missingBinary());
     }
   }
-  if (!path.isAbsolute(binaryPath)) binaryPath = path.join('../../..', binaryPath);
+  if (!path.isAbsolute(binaryPath)) {
+    binaryPath = path.join('..', '..', '..', binaryPath);
+  }
   return require(binaryPath);
 };
--- a/lib/binding.js
+++ b/lib/binding.js
@@ -17,6 +17,6 @@ module.exports = function(ext) {
       throw new Error(errors.missingBinary());
     }
   }
-  if (!path.isAbsolute(binaryPath)) binaryPath = path.join('..', binaryPath);
+  if (!path.isAbsolute(binaryPath)) binaryPath = path.join('../../..', binaryPath);
   return require(binaryPath);
 };
--- a/lib/binding.js
+++ b/lib/binding.js
@@ -3,22 +3,18 @@
  */
 
 var errors = require('./errors');
-var path   = require('path');
 
 /**
  * Require binding
  */
 module.exports = function(ext) {
-  var binaryPath = ext.getBinaryPath();
-  if (!ext.hasBinary(binaryPath)) {
+  if (!ext.hasBinary(ext.getBinaryPath())) {
     if (!ext.isSupportedEnvironment()) {
       throw new Error(errors.unsupportedEnvironment());
     } else {
       throw new Error(errors.missingBinary());
     }
   }
-  if (!path.isAbsolute(binaryPath)) {
-    binaryPath = path.join('..', '..', '..', binaryPath);
-  }
-  return require(binaryPath);
+
+  return require(ext.getBinaryPath());
 };
--- a/lib/binding.js
+++ b/lib/binding.js
@@ -3,18 +3,20 @@
  */
 
 var errors = require('./errors');
+var path   = require('path');
 
 /**
  * Require binding
  */
 module.exports = function(ext) {
-  if (!ext.hasBinary(ext.getBinaryPath())) {
+  var binaryPath = ext.getBinaryPath();
+  if (!ext.hasBinary(binaryPath)) {
     if (!ext.isSupportedEnvironment()) {
       throw new Error(errors.unsupportedEnvironment());
     } else {
       throw new Error(errors.missingBinary());
     }
   }
-
-  return require(ext.getBinaryPath());
+  if (!path.isAbsolute(binaryPath)) binaryPath = path.join('..', binaryPath);
+  return require(binaryPath);
 };
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -165,7 +165,9 @@ function getArgument(name) {
     }
   }
 
-  if (!flag) return null;
+  if (!flag) {
+    return null;
+  }
 
   return flag.substr(flag.indexOf('=') + 1);
 }
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -154,15 +154,10 @@ function isSupportedEnvironment(platform, arch, abi) {
  * @api private
  */
 
-function getArgument(name, args) {
-  var flags = args || process.argv.slice(2),
-    index = flags.lastIndexOf(name);
-
-  if (index === -1 || index + 1 >= flags.length) {
-    return null;
-  }
-
-  return flags[index + 1];
+function getArgument(name) {
+  var flag = process.argv.slice(2).find(a => a.indexOf(name) !== -1);
+  if (!flag) return null;
+  return flag.substring(flag.indexOf('=') + 1);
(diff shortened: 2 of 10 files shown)


lib/extensions.js
```diff
@@ -154,15 +154,10 @@ function isSupportedEnvironment(platform, arch, abi) {
...
-  var flags = args || process.argv.slice(2),
-    index = flags.lastIndexOf(name);
-
-  if (index === -1 || index + 1 >= flags.length) {
-    return null;
-  }
-
-  return flags[index + 1];
+function getArgument(name) {
+  var flag = process.argv.slice(2).find(a => a.indexOf(name) !== -1);
```
xzyfer: Please don't use ES6. We still support older versions of node

lib/binding.js
```diff
@@ -3,18 +3,20 @@
...
+  if (!ext.hasBinary(binaryPath)) {
     if (!ext.isSupportedEnvironment()) {
       throw new Error(errors.unsupportedEnvironment());
     } else {
       throw new Error(errors.missingBinary());
     }
   }
-
-  return require(ext.getBinaryPath());
+  if (!path.isAbsolute(binaryPath)) binaryPath = path.join('../../..', binaryPath);
```
xzyfer: Please keep to the existing code style.

```js
if (!path.isAbsolute(binaryPath)) {
  binaryPath = path.join('../../..', binaryPath);
}
```

xzyfer: We can't assume *nix

```js
binaryPath = path.join('..', '..', '..', binaryPath);
```

xzyfer: What is the rationale for going up exactly 3 directories?

wittibs: I make the assumption that non-absolute binary paths are relative to the working project directory (and I specify this in the docs). Consider the following project structure:

```
project
|-index.html
|-index.js
|-my-node-binding
  |-binding.node
|-node_modules
  |-node-sass
    |-lib
      |-binding.js
|-package.json
|-package-lock.json
```

You can see binding.js is 3 directories deep relative to the working project directory (node-modules/node-sass/lib). The ``require()`` function loads modules/files relative to the calling file. Since binding.js ``require()``'s the binary, we have to go up 3 directories.

xzyfer: This doesn't account for nested `node_modules` for instance.

```
project
|-index.html
|-index.js
|-my-node-binding
  |-binding.node
|-node_modules
  |-node-sass@4.5.2
    |-lib
      |-binding.js
  |-gulp-sass
    |-node_modules
      |-node-sass@3.7.1
        |-lib
          |-binding.js
|-package.json
|-package-lock.json
```

npm does not guarantee flat node_modules, and it's not possible to make such a guarantee when version conflicts happen. I think `process.cwd()` is probably a better fit.

wittibs:
Summary:
- removed ES6 ``find()`` function, replaced with ES5-compliant solution
- made code changes to adhere to coding style
- made suggested cross-platform update

Please review. Thank you!

xzyfer:
A few note. Removing support for the current `--foo bar` style is a breaking change. To get this change over the line we'd need to support both the old and new style. This can be done by asserting whether the matched value contains `=`.

>However that [won't work][1], as all custom CL arguments must have '--' prefixed.

I'm not sure I follow why this wont work given we're finding the index in the arg list the number of args beforehand shouldn't matter.

```
npm install node-sass -- --foo bar
```


[1]: https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script#answer-14404223

wittibs:
All custom CLI arguments must have '--' prefixed. So, this will fail:
```
npm install node-sass -- --foo bar
```
But this would work:
```
npm install node-sass -- --foo --bar
```
When you consider passing a binary path, we would need to strip the `--` from the beginning of the path.
```
npm install node-sass -- --sass-binary-path --./node_binary/binding.node
```
However, that's weird. I'd rather align with industry standard of arg=val, like so:
```
npm install node-sass -- --sass-binary-path=./node_binary/binding.node
```

wittibs:
Summary:
* modified `extensions.getBinaryPath()` to return an absolute path *from a relative path* if possible
  * it does this by attempting to `resolve()` the relative path against the current working directory (as you suggested earlier), but traverses up the directory tree trying to `resolve()` the path. This is how we get around the arbitrarily-nested `node_modules` directory problem
* undid changes to binding.js since they were no longer necessary

Please review. Thanks!

saper:
@wittibs is getting the absolute path really necessary?

wittibs:
@saper Yes - it is critical to ensuring relative binary paths will work for any level of depth of nested node_modules sub-directories.
For example: if I provide a binary path `./my-node-binding/binding.node` (relative to my project directory), I should expect the binary path to be resolved correctly regardless of node-sass directory depth. See sample file structure below (taken from @xzyfer 's comment from outdated code changes above):
```
project
|-index.html
|-index.js
|-my-node-binding
  |-binding.node
|-node_modules
  |-node-sass@4.5.2
    |-lib
      |-binding.js
  |-gulp-sass
    |-node_modules
      |-node-sass@3.7.1
        |-lib
          |-binding.js
|-package.json
|-package-lock.json
```

rdubigny:
Thx @wittibs for this work.

@zxyger can we merge this ASAP ?  We need this feature for our deployment pipeline to work :)

wittibs:
Hey @xzyfer just following up - have you had a chance to review this solution? If so, any feedback or thoughts?

Let me know if there's anything I can do to help expedite the PR. Thanks!

saper:
I am going to give it a try - how can Inreproduce the problem using Win Server 2016?

wittibs:
@saper try to install node-sass package with npmrc variable sass_binary_path set to a relative path. (#2081)

Try using any sass binary CLI argument during install or runtime (#2080)

saper:
Interesting problem when trying to use absolute paths: #2149

wittibs:
I'm considering closing this PR because:

- The fix for #2081 is non-deterministic, which is buggy in nature
- I don't think #2081 can be solved without some hackish fix. I think it should be the consuming application's responsibility to provide an absolute path for sass_binary_path.

Let me know what you all think. At the very least, I can:

- keep the fix for #2080 
- provide error message for when relative path is provided to sass_binary_path

Thoughts?

saper:
I've tried this today and it seems to work:

````> npm install node-sass --sass-binary-path=/tmp/binding.node

> node-sass@4.6.1 install /tmp/node/node_modules/node-sass
> node scripts/install.js

node-sass build Binary found at /tmp/binding.node

> node-sass@4.6.1 postinstall /tmp/node/node_modules/node-sass
> node scripts/build.js

Binary found at /tmp/binding.node
Testing binary
Binary is fine
````

I think something strange happened to npm since I believe the space separated form worked as well.

Can you try without `--` and with `=` separator?

**Comment by wittibs** [comment]
I tried that and... it worked! Although I have no idea why .... I don't think I fully understand how CLI arguments get parsed & passed to node apps. :/

Well this has been a good learning experience for me. Thank you for your time.

**Comment by jmcmullen** [comment]
@saper @wittibs Running into the same issue, but that fix has thrown me off completely. Can you tell me what I'm doing wrong?

```
$ node
> node-sass@4.6.1 install /tmp/lib/node-sass/linux-x64-51_binding.node
SyntaxError: Invalid or unexpected token
```

```
node-sass@4.6.1 install /tmp/lib/node-sass/linux-x64-51_binding.node
zsh: command not found: node-sass@4.6.1
```

Issue #2100, opened 2017-09-19
Use of const in strict mode causing failures installing Node 0.12


Bugget:
- NPM version: 2.11.3
- Node version: 0.12.7
- Node Process: { http_parser: '2.3',
  node: '0.12.7',
  v8: '3.28.71.19',
  uv: '1.6.1',
  zlib: '1.2.8',
  modules: '14',
  openssl: '1.0.1p' }
- Node Platform: linux
- Node architecture: x64
./ui/node_modules/gulp-sass/node_modules/node-sass/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
const Hoek = require('hoek');
^^^^^
SyntaxError: Use of const in strict mode.
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/tbugge/mocat/ui/node_modules/gulp-sass/node_modules/node-sass/node_modules/request/node_modules/hawk/lib/index.js:5:33)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)

Looks like a dependency changed.  I took a quick look at the releases on hawk and boom and didn't see anything recent.  This stopped building with this error in the last hour or so.

kshay:
We just hit this too. It looks like it was caused by this update to request:

https://github.com/request/request/commit/479143d546d6dff378da0e5fac5801dd3bf01f15

Bugget:
Indeed that was it.  In our package.json file we just added a hard set of request to version 2.81.0 which is before they set hawk to 6.x and that fixed the build.  So long as all your tests pass that is a work around.

mirzazeyrek:
This error message started to appear recently. Is there anything you guys suggest may be ?

/node_modules/grunt-sass/node_modules/node-sass/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
11:16:32 const Hoek = require('hoek');
11:16:32 ^^^^^
11:16:32 SyntaxError: Use of const in strict mode.
11:16:32 at exports.runInThisContext (vm.js:73:16)
11:16:32 at Module._compile (module.js:443:25)
11:16:32 at Object.Module._extensions..js (module.js:478:10)
[...]

PR #2112, opened 2017-10-06
Remove Node 0.10 and 0.12 from CI config

xzyfer:
These will always fails now due to https://github.com/sass/node-sass/issues/2100. We manually verify support when building release binaries so there's no value in failing CI. It's scary to contributors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,6 @@ jobs:
     - stage: platform-test
       node_js: "lts/argon"
       os: osx
-    - stage: platform-test
-      node_js: "0.12"
-      os: linux
-    - stage: platform-test
-      node_js: "0.10"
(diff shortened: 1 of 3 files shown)


xzyfer:
I think a lot of people are still using 7. I know my work is, which is plenty enough reason for me to keep it around :)

xzyfer:
Added a note to the readme linking to the troubleshooting entry

nschonni:
OK, but for the node-sass@5, I think all the unsupported versions should go away. Especially since the odd number releases just roll into the even ones when they go stable

xzyfer:
Agreed. Active LTS and current is nice but forces us to bump the major every 12 months to drop odd number releases. I think this is ok, but time will tell.

PR #2119, opened 2017-10-21
Allow setting of SASS_BINARY_DIR without setting explicit binary name

kausters:
I'm at a company with people developing on multiple platforms behind a strict firewall. I want to host precompiled SASS bindings locally so setting SASS_BINARY_URL doesn't work for me. However I do not want to use SASS_BINARY_PATH because it requires giving explicit file name of the binding which doesn't work when some developers are on Macs, some on WIndows, and the whole project is compiled on Linux for production — it is quite a pain to set the full path as required in each environment, and as far as I know impossible within a static .npmrc file (.npmrc supports expansion of environment variables but I'd still need to set the SASS version, OS arch & Node Modules version).

Introducing: SASS_BINARY_DIR! Enables setting a directory where the binaries are stored but uses the default file name which SASS figures out by itself.

Also updated the `getBinaryPath` to use the new `getBinaryDir`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,6 +567,7 @@ Variable name    | .npmrc parameter | Process argument   | Value
 SASS_BINARY_NAME | sass_binary_name | --sass-binary-name | path
 SASS_BINARY_SITE | sass_binary_site | --sass-binary-site | URL
 SASS_BINARY_PATH | sass_binary_path | --sass-binary-path | path
+SASS_BINARY_DIR  | sass_binary_dir  | --sass-binary-dir  | path
 
 These parameters can be used as environment variable:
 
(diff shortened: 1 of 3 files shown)


xzyfer:
Thanks! Conceptually i support the idea, I'll need some time to look over
the implementation.

kausters:
Thanks! I mostly just reused the code for SASS_BINARY_PATH, let me know if I should change anything.

xzyfer:
All good. There's currently another active PR working on sass binary path,
and we're working the next major version of node sass is all

xzyfer:
Thanks for you work :) This will land in 4.8.0

PR #2136, opened 2017-11-03
docs: Update the Node version support message

nschonni:
Pointing them to the release notes is safer since it will accurately reflect the support for that version of node-sass


xzyfer:
:+1:

Issue #2157, opened 2017-11-20
scripts/install.js tries to traverse filesystem tree; breaks if access not possible


criess:
I try to install node-sass 4.7.1 in the following environment:

npm install in (any) local user folder.
npm 3.10.10
node v6.11.5

I can confirm this an 4.7 Bug, as 4.6.x installs just fine.

`npm install node-sass@"~4.6.0"` runs fine.
`npm install node-sass@"~4.7.0"` breaks stuff.

Somehow the installer script tries to traverse any parents of the current install location. I don't know why it does so. This breaks in my case where the traversing hits the vhost level of my hosting config (`/var/www/vhosts/myfancyvhost.com`)

<pre>
npm WARN prefer global node-gyp@3.6.2 should be installed with -g

> node-sass@4.7.1 install /var/www/vhosts/myfancyvhost.com/antrag/shared/bundle/ruby/2.4.0/bundler/gems/entwicklung.frontend-oa-75ea994727c8/node_modules/node-sass
> node scripts/install.js

/var/www/vhosts/myfancyvhost.com/antrag/shared/bundle/ruby/2.4.0/bundler/gems/entwicklung.frontend-oa-75ea994727c8/node_modules/true-case-path/node_modules/glob/sync.js:321
        throw er
        ^

Error: EACCES: permission denied, scandir '/var/www/vhosts/myfancyvhost.com'
    at Error (native)
    at Object.fs.readdirSync (fs.js:953:18)
    at GlobSync._readdir (/var/www/vhosts/myfancyvhost.com/antrag/shared/bundle/ruby/2.4.0/bundler/gems/entwicklung.frontend-oa-75ea994727c8/node_modules/true-case-path/node_modules/glob/sync.js:275:41)
    at GlobSync._processReaddir (/var/www/vhosts/myfancyvhost.com/antrag/shared/bundle/ruby/2.4.0/bundler/gems/entwicklung.frontend-oa-75ea994727c8/node_modules/true-case-path/node_modules/glob/sync.js:135:22)
    at GlobSync._process (/var/www/vhosts/myfancyvhost.com/antrag/shared/bundle/ruby/2.4.0/bundler/gems/entwicklung.frontend-oa-75ea994727c8/node_modules/true-case-path/node_modules/glob/sync.js:130:10)
[...]

PR #2149, opened 2017-11-10
Use true-case-path to fix re-entrancy issue on Windows

MLoughry:
Fixes #1895 

When using node-sass through multiple entry points (as can happen when using webpack) on Windows, the casing of the drive letter in the path can be inconsistent. Since Node is case-sensitive to a fault, this results in Node attempting to create a second instance of the native binding, which results in a `Module did not self-register` error.

The fix is to use `true-case-path` to normalize the binding path, to prevent unintended duplicate bindings from occurring.

PR #2161, opened 2017-11-20
Use trueCasePath only in Windows environments

arrkiin:
During the install process of node-sass in [termux](https://github.com/termux/termux-app) under android I got an error like in this issue here #2157. 
Like @criess mentioned in his issue, the change for issue #2149 caused the problem of breaking installation in restricted environments like [termux](https://github.com/termux/termux-app) or protected cloud hosting environments. Because this change is mainly focused towards Windows installations I integrate the corresponding check to use it only if executed on Windows.

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -277,7 +277,7 @@ function getBinaryPath() {
     return binaryPath;
   }
 
-  return trueCasePathSync(binaryPath) || binaryPath;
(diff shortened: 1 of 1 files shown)


xzyfer:
Thanks, we went with a slightly different approach. This should be fixed in v4.7.2.

PR #2163, opened 2017-11-22
Add a Git .mailmap with mgol's new name

mgol:
In this way his past contributions are mapped correctly, including collapsing
two different ways of writing "ę" into one.


xzyfer:
TIL. LGTM.

PR #2170, opened 2017-11-29
chore: update `request` package

evilebottnawi:
Fix security issues (see https://github.com/request/request/blob/master/CHANGELOG.md#v2810-20170309)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "nan": "^2.3.2",
     "node-gyp": "^3.3.1",
     "npmlog": "^4.0.0",
-    "request": "~2.79.0",
+    "request": "~2.83.0",
     "sass-graph": "^2.2.4",
     "stdout-stream": "^1.4.0",
     "true-case-path": "^1.0.2"


xzyfer:
This breaks node < 4 support and would be a very big major change for us.
This will happen in version 5.

jc1arke:
Just checking: since this is now being halted on downstream projects by NSP, when will version 5 be released?

Support for Node 4 as an LTS ends in basically 2 months (https://github.com/nodejs/Release#release-schedule), meaning that blocking the dependency update to cater for a soon-to-be-deprecated version just doesn't seem valid.

kumorig:
It should be mentioned that since `npm 6.0.0` anyone who uses node-sass in their project and run `npm install` or `npm update`  gets a big red `[!] 5 vulnerabilities found | Severity: 5 moderate` in their console. I'm not sure of the actual implication of those vulnerabilities, but it *looks* pretty bad from an optics standpoint.

tflori:
just my two cents: node-sass 4.9 requires node-gyp 3.3 and node-gyp 3.3 that also requires hawk 4.2.1 with the result that node-sass 4.9 already requires node >= 4 - so I don't understand the point in blocking an update because it is breaking node 0.10 support :confused:

nschonni:
node-gyp is only used if the binary download fails, which is done through request

Issue #1895, opened 2017-02-13
Error: Module did not self-register


MLoughry:
I'm in the process of upgrading my team's project to webpack 2, and I'm now encountering numerous "Module did not self-register" errors for `.scss` files, both locally and in our build system. Example errors for one file:

```
ERROR in c:/Agents/01/_work/1/s/obj/framework/owa-application/lib/styles/globalstyles.global.scss
Module build failed: ModuleBuildError: Module build failed: Error: Module did not self-register.
    at Error (native)
    at Object.Module._extensions..node (module.js:597:18)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at module.exports (C:\Agents\01\_work\1\s\node_modules\node-sass\lib\binding.js:19:10)
    at Object.<anonymous> (C:\Agents\01\_work\1\s\node_modules\node-sass\lib\index.js:14:35)
    at Module._compile (module.js:570:32)
    at C:\Agents\01\_work\1\s\node_modules\webpack\lib\NormalModule.js:141:35
    at C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:364:11
    at C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:170:18
    at loadLoader (C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\loadLoader.js:27:11)
    at iteratePitchingLoaders (C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:169:2)
    at iteratePitchingLoaders (C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:165:10)
    at C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:173:18
    at loadLoader (C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\loadLoader.js:36:3)
    at iteratePitchingLoaders (C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:169:2)
    at iteratePitchingLoaders (C:\Agents\01\_work\1\s\node_modules\loader-runner\lib\LoaderRunner.js:165:10)

[...]

Issue #2111, opened 2017-10-06
Breaking changes for version 5


xzyfer:
Current WIP: https://github.com/sass/node-sass/pull/2312

We have been planning to release node-sass@v5 with LibSass 3.5.0 stable when it's ready. With a 5.0 milestone on the horizon we want to take note of the breaking changes we're planning to make.

This is living, and incomplete. Please subscribe for updates.

## Node versions support 

Moving forward we'll only be actively supporting active LTS and current Node versions. In practical terms this means Node ~6~10+. 

- supporting old versions of Node and npm is a significant maintenance burden
- older versions of npm are notoriously troublesome when it comes to native modules - it's becoming impractical, and sometimes even unsafe, to continue using old packages.

See #2290
See #2286
See #2257
See #2170
See #2256
See #2205
See #2288
...the list goes on

## Switch watcher to node-chokidar

>Why? Because Gaze in docker and various virtual machines uses a lot of resources whereas chokidar does not. Read about the advantages of [chokidar](https://github.com/paulmillr/chokidar)

>[...] in docker for mac you will get really high CPU usage with com.docker.hyperkit and com.docker.osxfs (I've seen reports of up to 300%).

See #2208 

## Compile on watch by default

When using the watch flag we should do a compilation before watching.
This becoming the expected behaviour in JS (see webpack).

See https://github.com/sass/node-sass/issues/2300
See https://github.com/sass/node-sass/issues/1973
See https://github.com/sass/node-sass/issues/1369
See https://github.com/sass/node-sass/issues/1742

## Stop watching .css files

A LibSass has a feature/bug that allows it to `@import` .css files. 
This means our watcher must also watch for .css files.
This can cause infinite loops if the input and out directories are the same.
LibSass is deprecating this behaviour.

See https://github.com/sass/libsass/issues/2611
See #2184
See #2006
See #1933
See #1925
See #1867
See #1845

## Don't fail installation on unsupported environments

In order to reduce issues about installation issues we failed the installation if we detected an environment we didn't provide binaries for. 

This had the intended affect, but also had some unfortunate side-effects
- difficult or impossible to install on environments that could fallback to local compilation i.e. arm, electron
- required a version bump when new version of node landed
- couldn't back port support for new Node versions

In v5 we should still produce an informative error, but allow the installation to continue.

nschonni:
For the node support version here is my proposal. Since Node publishes their LTS and dev branch timeline on https://github.com/nodejs/Release#lts_schedule
[...]

PR #2179, opened 2017-12-05
Normalize drive letter

MLoughry:
This is a follow-up to #2149. 

I did not realize `true-case-path` did not normalize the drive letter, which can cause the same issues as described in #1895. The project does not seem to be maintained any longer, so I've removed the dependency and copied the code here with one small change to normalize the drive letter.

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -295,24 +295,26 @@ function trueCasePathSync(fsPath) {
   // !! As of Node v4.1.1, a path starting with ../ is NOT resolved relative
   // !! to the current dir, and glob.sync() below then fails.
   // !! When in doubt, resolve with fs.realPathSync() *beforehand*.
-  var fsPathNormalized = path.normalize(fsPath)
+  var fsPathNormalized = path.normalize(fsPath);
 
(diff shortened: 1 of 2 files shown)


nschonni:
Due to the license, we can't just copy/paste this in https://github.com/barsh/true-case-path/blob/master/LICENSE
The next major of node-sass will also likely be dropping all our custom download and binding code for node-pre-gyp, so I'm going to close this

MLoughry:
Hmmm. Your repo, your call; but the code for that package is sourced entirely from a StackOverflow post. I'm not sure how they can claim license over it.

https://stackoverflow.com/questions/33086985/how-to-obtain-case-exact-path-of-a-file-in-node-js-on-windows/33139702#33139702

MLoughry:
Is there a timeline for the next major version?

nschonni:
> Hmmm. Your repo, your call; but the code for that package is sourced entirely from a StackOverflow post. I'm not sure how they can claim license over it.

:man_shrugging:  https://github.com/barsh/true-case-path/issues/1

For the v5 stuff, we're discussing it over in https://github.com/sass/node-sass/issues/2111
I think the main thing is getting libsass to cut a stable 3.5 release

PR #2205, opened 2018-01-06
Update request -> 2.80.0

ydxia:
request 2.79.0 depends on an old version of har-validator which depends on a package that no longer exists (pinkie-promise)

2.80.0 bumps the version of har-validator which no longer depends on pinkie-promise


xzyfer:
Please search issues before for a new one. Update request breaks support
for Node < 4.

This was an npm issue that's now been resolved

https://status.npmjs.org/incidents/41zfb8qpvrdj

PR #2206, opened 2018-01-09
Add command that configure China mirror

gucong3000:


nschonni:
@gucong3000 our install process usually relys on GitHub releases for the binary downloads. Is that also something that is blocked? I know it can fall back to local build if necessary, I'm just curious of the experience with the alternate NPM install.

xzyfer:
It'd be better to general document settings a proxy using npm config or Env
cars rather than suggesting a specific library and proxy server

gucong3000:
> something that is blocked?

Yes, [Great Firewall](https://en.wikipedia.org/wiki/Great_Firewall) randomly dropp network packets from GitHub and Amazon S3 Service.

xzyfer:
This seems reasonable to me. Curious to hear @EdwonLim thoughts since he maintains [node-sass-china](https://github.com/EdwonLim/node-sass-china).

README.md
```diff
@@ -50,6 +50,13 @@ SyntaxError: Use of const in strict mode.
 
 **Having installation troubles? Check out our [Troubleshooting guide](/TROUBLESHOOTING.md).**
 
+### Install from mirror in China
+
+```shell
+npx mirror-config-china --registry=http://registry.npm.taobao.org
```
saper: Can this be done without adding `npx` dependency? Something like  

    npm i -g mirror-config-china --registry=https://registry.npm.taobao.org

gucong3000: Yes.

PR #2219, opened 2018-01-19
Fixed compilation issue under certain config

phg1024:
The missing scope identifier could lead to unresolved `distance` function under certain compiler configuration. Adding the `utf8::` to fix this issue.


nschonni:
Thanks @phg1024, unfortunately that file is part of https://github.com/sass/libsass/ so the patch would need to go there.

phg1024:
@nschonni Thanks. I'll propose the change for `libsass`.

nschonni:
Thanks. Linking #2545
Not that we don't pull in libsass often, so you might have to monkey patch for a bit if this is causing you alot of issues ❤️

Issue #2011, opened 2017-06-09
Compile to WebAssembly


mafintosh:
Since Node 8 supports WebAssembly it would be nice to compile libsass to that using emscripten so people wouldn't need to compile the c++ or depend on a prebuild.

nschonni:
There already is an official libsass Emscripten project https://github.com/medialize/sass.js

xzyfer:
Yeah but with wasm we can in theory entirely ditch vendor binaries for Node
8+

saper:
Dropping binaries would be fantastic... I wonder how well will that setup perform

xzyfer:
@mafintosh I'm starting look into this seriously, but I'm struggling with finding resources on how to do this for native extensions in Node. Do you know of any other projects that have made this switch to WASM that I can reference?

aduh95:
@xzyfer I tried to install the Emscripten SDK on my computer to compile libsass to WASM; good news, it compiles without changing the C++ code!

I follow the following steps:

1. [Activate the Emscripten SDK on my environment](http://webassembly.org/getting-started/developers-guide/#compile-emscripten-from-source)
2. On the latest commit of libsass, I run the makefile with the Emscripten compiler:
```bash
$ make clean && make --eval="CC=emcc
> CXX=emcc"
```
3. Then I compile the library into wasm: `emcc src/*.o lib/libsass.a -O2 -s WASM=1 -o libsass.js`

The `libsass.wasm` file should now contain the library, and can be run by node >=8
Also I try to change the `binding.js` file of node-sass to use the cross-platform WASM instead of the platform-dependent binary, but I suppose I'm missing something since node displays an error:

In `binding.js`, I changed:

```js
return ext.getBinaryPath()
```

by

```js
    return require('./libsass.js'); // libsass.js is the file generated by emcc
```

And it outputs the following:

```
TypeError: binding.libsassVersion is not a function
    at Object.getVersionInfo (.../node_modules/node-sass/lib/extensions.js:368:27)
    at Object.<anonymous> (.../node_modules/node-sass/lib/index.js:448:28)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
    at Module.require (module.js:517:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous>
```

I hope I'm helping!

[...]

Issue #1988, opened 2017-05-31
Look at N-API to see if it would help our binding portablitiy


nschonni:
Still experimental with 8, but it looks like they're trying to integrate the Nan approach (with some other stuff) into the platform https://nodejs.org/api/n-api.html

/cc @mgol  and @saper who were talking about this in Slack too

xzyfer:
I've been following N-API closely. My bet is the that the community is blocked until nan adds support. There is currently no tracking issue with nan yet.

saper:
I do not think that squeezing N-API into NAN is possible.

I am looking right now if our low-level uv callback stuff can be converted into N-API, which offers some callback and async job handling.

boingoing:
N-API is meant to be mostly a replacement for NAN in that N-API is designed to be non-leaky about V8 or engine details and provide a flat, stable ABI surface. An N-API enabled native module binary shouldn't have to be re-compiled to continue working with future versions of Node.js. But I'm not going to go completely into the sales pitch here, you can read about it. 😁 

We (N-API team) have built a preliminary port of node-sass on top of N-API as part of our effort to understand the API surface used by native modules. We used this port and the port of some other modules to determine which areas to focus on converting to N-API. This port has all of the NAN and V8-specific code removed and replaced by N-API equivalents.

I doubt this port meets the quality level to go in as-is, but it would certainly be a springboard if someone on node-sass was serious about doing the conversion to N-API. 😄 Please take a look if you're interested. I will be happy to answer questions about the port, help with coding it, point to documentation/code, etc.

https://github.com/boingoing/node-sass/tree/napi

nschonni:
@boingoing if Nan is replaced with the napi convention, would we still be able to produce builds compatible with at least the current LTS versions of Node (4+) or is this only going to work for 8+

boingoing:
The N-API enabled module binary can be loaded by older Node versions as well (4 and 6) by relying on an NPM package. There are more details in the documentation of the package:

https://github.com/nodejs/node-addon-api

saper:
Thank you @boingoing this is very helpful. From what I can see some `libuv` specific code is still left there (but that's no V8 and even not C++).

It looks fairly complete and maybe even works, nice! :)

jasongin:
> My bet is the that the community is blocked until nan adds support.

[...]

PR #2220, opened 2018-01-19
POC: Using WebAssembly build of libsass

kwonoj:
**Not to be merged**. More of proof of concept to illustrate high-level codeflows with pros & cons, and roadmaps to make it happen. This PR aims to provide proof-of-concept implementation of #2011, replaces current native libsass module into WebAssembly based binary.

### How to run demo

`npm install && node ./lib/loadModule.js`. I haven't spend time to node.js version compatibility for `bootstrap.ts`, so ensure it doesn't emit error (and `wget` is available on shell too).

you can see version information, and simplest sass compilation result.

image

### Code structure

Building `libsass` requires several native dependency & emsdk setup, but only necessary when need to bump up libsass itself. Instead of managing native build in single repo, this PR uses prebuilt binary via [docker-libsass-wasm](https://github.com/kwonoj/docker-libsass-wasm), preconfigured docker image generates specified libsass version of binary. In this PR, it builds latest beta 3.5.0.beta.3. `node-sass` uses tagged release from repo via `bootstrap` script, nothing more than just download binaries and put under correct location.

Prebuilt binary exposes C interface of `libsass` - in `loadModule.js`, `node-sass` creates corresponding bindings to C functions, wraps C interface to javascript functions with proper pointer interop, and finally exposes `loadSass` interface to compile wasm binary and build wrapped binding into `sass` module. In this PR, it exposes `sass.info` and naïve version of `renderSync` which only accepts data string.

### Pros & cons

Pros

- does not need native module anymore
- supports asm.js as fallback for node.js doesn't support native wasm

Cons

- Wasm compliation happens asynchronously, introduces breaking change interface `loadSass(): Promise<sass>` to prepare sass module correctly
- Basically this is rewriting of `node-sass`, dropping existing native bindings
- cabllback based async bindings (i.e `render`) is now won't be well supported, as it's basically not supported to create async fn via libuv queue. To provide interface level compatibility could trap out sync version into nexttick, but not ideal. (there are some experimental approach like `Emterpreter` but it is unsure it'll even work with libsass)

### Roadmap (if this is a thing to go)

- [ ] create wrapped interfaces / pointer interop for all necessary libsass interface
- [ ] migrate existing synchronous sass interface (`renderSync`, ...) 
- [ ] simple nexttick trapping for async interface (`render`, ...)
- [ ] evaluate empreter to replace ☝️ 
- [ ] migrate test cases to verify sanity
- [ ] explore possibility of creating correct async interface instead of trapping
- [ ] (good to have) run `libsass` test cases to wasm binary, to verify sanity of wasm binary itself

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ test/fixtures/watching-css-out-01
 test/fixtures/watching-css-out-02
 coverage
 package-lock.json
+libsass-asm.js
+libsass-wasm.js
--- a/bootstrap.ts
+++ b/bootstrap.ts
@@ -1,0 +1,75 @@
+/*tslint:disable:no-console*/
+
+/**
+ * Script to download libsass wasm binary from https://github.com/kwonoj/docker-libsass-wasm
+ */
+
+import * as crypto from 'crypto';
+import * as fs from 'fs';
+import * as path from 'path';
+import { exec, mkdir, rm } from 'shelljs';
+import { promisify } from 'util';
+//tslint:disable-next-line: no-require-imports no-var-requires
+const { config } = require('./package.json');
+
+// Package.json defines `libsass-version` under `config` section to find corresponding release version
+const version = config['libsass-version'];
+const readFile = promisify(fs.readFile);
+const writeFile = promisify(fs.writeFile);
+const asyncExec = promisify(exec);
+
+/**
+ * Generate sha512 checksum from given file.
+ */
+const calculateChecksumFromFile = async (filePath: string) =>
+  crypto
+    .createHash('sha512')
+    .update(await readFile(filePath))
+    .digest('hex');
+
+/**
+ * Get remote release checksum.
+ */
+const getRemoteChecksum = (url: string) => {
+  const { stdout } = exec(`wget -qO- ${url}.sha512`, { silent: true });
+  return (stdout as string).slice(0, (stdout as string).indexOf(' '));
+};
+
+/**
+ * Simple script to download compiled wasm binaries.
+ * wasm binary itself is being separately built & bumped up when libsass version changes,
+ * and this bootstarp script download corresponding version in package.json install time.
+ * Published package will ship with downloaded binary, download only occurs on prepare stage.
+ *
+ */
+(async () => {
+  const libPath = path.resolve('./lib');
+  const fileNames = ['libsass-asm.js', 'libsass-wasm.js'];
+
+  for (const fileName of fileNames) {
+    const localBinarypath = path.join(libPath, fileName);
+
+    const url = `https://github.com/kwonoj/docker-libsass-wasm/releases/download/${version}/${fileName}`;
+
+    //Create checksum validator
+    const remoteChecksum = getRemoteChecksum(url);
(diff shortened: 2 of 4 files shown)


kwonoj:
Feel freely close PR if this isn't way to go.

kwonoj:
few things need to be evaluated. current wasm design so far. at least doesn't have way to queue sync functions and control stack execution in async flow.

for emscripten side there's empreter which looks interesting, but that's something I haven't tried in this poc. I wouldn't say it's impossible forever, but safe to assume it'll be tricky.

kwonoj:
@xzyfer example of empreter-based async flow execution : https://github.com/curiousdannii/emglken/blob/a98cc6c75d8214184afaeeb7f6ccbed0cd79e3af/emglken/library.js#L390-L414

xzyfer:
[I've reached out] to the N-API to get some clarification whether the N-APIs place nicer with WASM

[1]: https://github.com/sass/node-sass/issues/1988#issuecomment-359129948

kwonoj:
It'd be interesting if n-api have way to support it. If there isn't, I believe emterpreter is way for immediate direction to go with wasm.

jayphelps:
> Wasm compliation happens asynchronously, introduces breaking change interface loadSass(): Promise<sass> to prepare sass module correctly

It shouldn't have to be, if that's not desired. 😄 [Emscripten supports](https://github.com/kripken/emscripten/wiki/WebAssembly#wasm-files-and-compilation) sync via [`new WebAssembly.Module(binary)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Module) `BINARYEN_ASYNC_COMPILATION=0`. As long as you don't need browser support, because browsers won't allow you to synchronously load binary data on the main thread.

kwonoj:
oh I had some issue when I tried on browsers, so it doesn't have any limitation to sync compile on nodejs? @jayphelps

jayphelps:
@kwonoj yep, works good in latest node. Dunno what the minimum version is off-hand though.

```js
const module = new WebAssembly.Module(new Uint8Array([
  0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00
]));
```

kwonoj:
Oh good to know. I still prefer async compilation but at least we have option here.

kwonoj:
@jayphelps uh.. second thought: sync loading works with asm.js build as well? 😅 I don't think `node-sass` can aim native wasm supported node.js version only.

jayphelps:
👍 asm.js compiles the same as JS, so to do it’s normaly sync but can be async with `<script async>` or in node with the vm module stuff.

xzyfer:
FWIW we have no intention to have browsers compatibility. There's already https://github.com/medialize/sass.js/ for that use case.

kwonoj:
I don't expect browser compat too. mostly aims recent version of node, including non-native wasm versions as well via asm.js.

kwonoj:
Closing as inactive. I believe PR itself has served sufficiently for proof verification for the possible road.

xzyfer:
Additional useful resource: https://github.com/medialize/sass.js/compare/feature/wasm

PR #2233, opened 2018-01-27
Fix TypeScript problem

DusterTheFirst:
With the .js at the end of the filename, TypeScript was unable to find that declarations from @types/node-sass


nschonni:
This isn't a TypeScript project, and that advice doesn't match NPMs https://docs.npmjs.com/getting-started/using-a-package.json

PR #2238, opened 2018-01-31
Auth parameters for binaries download in install.js

shiaart:
In some cases, when SASS_BINARY_SITE is used, site may have authentication/authorization,
For such cases I've added 2 params, to be able to pass username and password to request module (which is used for download) - https://www.npmjs.com/package/request#http-authentication


xzyfer:
Does the http scheme approach not work with `SASS_BINARY_SITE`?

```
SASS_BINARY_SITE=https://user:pass@mysite.com
```

shiaart:
Just checked.
```
SASS_BINARY_SITE=https://user:pass@mysite.com
```

Works fine for both cases. when I provide clear text password and encrypted password (my specific case).
Thanks for pointing it out!

Michenux:
The install script should not display the url with login/password.

PR #2268, opened 2018-03-04
Respect quiet option in Node API

### LINKED ISSUE #2265: add quiet as non-cli option?
i use [grunt-sass](https://github.com/sindresorhus/grunt-sass) which implements node-sass. i know that node-sass offers a `quiet` option, but when i pass `{"quiet": true}` as an option, i continue to get debug warnings during compilation. i believe this is because `quiet` is not actually implemented as an option outside the cli. that being the case, it would be wonderful if you could pass `quiet` as an option when instantiating node-sass. please consider implementing this enhancement.

thanks!

xzyfer:
- Emit normal output as info not warn
- Don't emit info output in quiet mode

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -329,7 +329,7 @@ function run(options, emitter) {
 function renderFile(file, options, emitter) {
   options = getOptions([path.resolve(file)], options);
   if (options.watch) {
-    emitter.emit('warn', util.format('=> changed: %s', file));
+    emitter.emit('info', util.format('=> changed: %s', file));
   }
   render(options, emitter);
 }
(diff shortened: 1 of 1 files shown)


xzyfer:
You're totally right. The issue is confusing.

xzyfer:
I agree this isn't related the the tagged issue (now removed), but it still the correct behaviour.

RiZKiT:
@xzyfer Thanks a lot for the 'warn' > 'info' change, the useless warning output in our build process was always confusing.

It would be nice to change it here, too: https://github.com/sass/node-sass/blob/master/lib/render.js

xzyfer:
Great suggestion. We're open to PRs, otherwise we'll try to get it sorted
in v5

xzyfer:
Addressed in #2344.

saper:
While troubleshooting CLI watcher tests (for https://github.com/sass/node-sass/issues/2560 and https://github.com/sass/node-sass/issues/2479 I have noticed that this and #2344 has a bad side-effect.

Watcher messages go out now to standard output instead of to standard error.

This breaks the following (seldom) usage:

     node-sass --output-style compressed --watch main.scss

Our (currently skipped) watcher tests expect that CSS output will be delivered to stdout.

I have filed #2755 about this

PR #2298, opened 2018-03-19
fix: propagate async context

ofrobots:
EDIT: This is an improvement on some of the fixes in #2295.

Starting with Nan 2.9.0, we have the ability to propagate async context
across async hops. Certain variants of Nan::Callback::Call are now
deprecated to encourage use of the context presevering alternatives.
Certain variants of Node's MakeCallback that were used internally are
going to be deprecated in Node 10.

Summary is that one should use Nan::Call for sync calls, and
Nan::Callback::Call for async. The latter expects an async resource
corresponding to the async operation to be provided at the call time.

This patch fixes things up so that 1) node-sass isn't using any
deprecated APIs, and 2) properly propagates async context for async
callbacks by creating async resources in the appropriate places.

Related: https://github.com/nodejs/nan/issues/755
/cc @xzyfer

### DIFF
--- a/src/binding.cpp
+++ b/src/binding.cpp
@@ -227,9 +227,16 @@ int GetResult(sass_context_wrapper* ctx_w, Sass_Context* ctx, bool is_sync = fal
   return status;
 }
 
+void PerformCall(sass_context_wrapper* ctx_w, Nan::Callback* callback, int argc, v8::Local<v8::Value> argv[]) {
+  if (ctx_w->is_sync) {
+    Nan::Call(*callback, argc, argv);
+  } else {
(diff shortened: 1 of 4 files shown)


ofrobots:
I just noticed https://github.com/sass/node-sass/pull/2295. Seems like @kkoopa beat me to it.

ofrobots:
Upon deeper inspection, I think this PR should still be preferred, as it avoids creating async resources and using MakeCallback (with its tick affecting behaviour) for synchronous calls. I can rebase if folks agree.

xzyfer:
Thanks for this @ofrobots. Looks the CI failure was transient.

PR #2312, opened 2018-04-03
Version 5.0

xzyfer:
See #2111

## Features

- [x] Reduce Node version support to [Active LTS and Current][1] (@xzyfer) **\*BREAKING\***
- [x] Replace [Gaze][2] with [Chokidar][3] (@xzyfer)
- [x] Compile on watch by default (@xzyfer) **\*BREAKING\***
- [x] Stop watching and compiling .css files (@xzyfer) **\*BREAKING\***
- [ ] Allow fallback to local compile on unsupported environments
- [ ] LibSass 3.6

## Misc

- [x] Update dependencies (@xzyfer) **\*BREAKING\***
- [ ] ~ES6-ify where it improves readability **\*BREAKING\***~
- [x] Remove deprecated `process.sass` API (@xzyfer) **\*BREAKING\***
- [x] Set `engines` to supported Node versions (#2316, @realityking) **\*BREAKING\***
- [x] Remove some Node < 6 work arounds (#2317, @realityking) **\*BREAKING\***
- [x] Remove `Object.assign` polyfill (#2318, @realityking)
- [x] Remove async-foreach dependency (@xzyfer)

## Maybe

- [ ] Official electron support

[1]: https://github.com/nodejs/Release
[2]: https://npm.im/gaze
[3]: https://npm.im/chokidar


package.json
```diff
@@ -54,31 +54,31 @@
   ],
   "dependencies": {
     "async-foreach": "^0.1.3",
-    "chalk": "^1.1.1",
-    "cross-spawn": "^3.0.0",
-    "gaze": "^1.0.0",
-    "get-stdin": "^4.0.1",
+    "chalk": "^2.3.2",
+    "chokidar": "^2.0.3",
+    "cross-spawn": "^6.0.5",
```
nschonni: Do we still need this one if we don't have old Node?

xzyfer: Good question. It works around a bunch of `child_process.spawn` issue, but they're maybe not relevant. Lets seen what CI thinks.

xzyfer: Removing it made [Windows CI sad](https://ci.appveyor.com/project/sass/node-sass/build/2776/job/2d3xrs03rgvtgb75#L8877). There is a lot of code to work around [Windows in cross-spawn](https://github.com/moxystudio/node-cross-spawn/blob/master/lib/parse.js#L33-L65) which I'm happy to not have to deal with :)

YasharF:
Would you be able to do a pre-release, release candidate, or beta release of v5, or even better the v5 itself, as is, with plans to fix/address the remaining issues in 5.1?

Thanks

xzyfer:
Rebased on master

paulschreiber:
What's the planned release date for v5?

hiepxanh:
I'm really hope about electron support, I'm really looking for this and I will waiting for the team. Best wish to team

xzyfer:
@nschonni @saper I've created the `next-major` branch. The plan is to merge all the 5.0 milestone changes into that branch so that we can start cutting 5.0 betas with the latest libsass, and trial N-API.

saper:
Is N-API here ready for the prime time? I think have already big changes in, let's not put too many...

xzyfer:
Lots of extensions have moved over. We had issues with node 9 initially but
with node >= 10 ci seems happy.

saper:
Sure the code works **now** - the question is can we release a new binding file with a new version number when the libuv gets upgraded (it gets upgraded pretty often, but rarely breaks compatibility).

I wrap by head around this: https://nodejs.org/api/n-api.html#n_api_implications_of_abi_stability

We include `<uv.h>` in  `src/callback_bridge.h` and we need it. That probably means we have to check our compatibility with `libuv` via some other means.

xzyfer:
N-API has async APIs we can use instead of direct libuv
https://nodejs.org/api/n-api.html#n_api_simple_asynchronous_operations

saper:
Yes, we should switch to that. Looks like a non-trivial rewrite to me :(

xzyfer:
Definitely. I think I considered before and ruled it out. I need to
refamiliarise myself with the implementation.

saper:
But that means to me we should drop it if we want to kick 5.0 out of the door quickly. Sorry if I am feeling pushy and not contributing enough... :(

xzyfer:
I appreciate your feedback. Node 8 will is in Jan. I agreed we need to
understand the trade offs of uv's abi, and do our best to limit the impact.
I'll keep digging into it.

saper:
Sure, I'll have a look at it too.

PR #2313, opened 2018-04-03
Auto Switch download mirror

gucong3000:


xzyfer:
This seems reasonable to me. Thoughts @nschonni ?

xzyfer:
Nice catch. Had a look at the module and I agree.

xzyfer:
Merged #2206 instead. Thanks heaps @gucong3000

PR #2316, opened 2018-04-04
[v5] Indicate the supported Node.js version in package.json

realityking:
I based the range on what's indicated in `.travis.yml` but since Travis always defaults to the newest release in each series a higher minor could be selected as well.

Especially Node.js 4.5 might be a good pick since it's the first release to support `Buffer.from`, `Buffer.alloc` and `Buffer.allocUnsafe`.

In #2111 I found this mention in the summary:

> Moving forward we'll only be actively supporting active LTS and current Node versions. In practical terms this means Node 6+.

But it's not currently reflected in the Travis config.

Relates to #2312

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "url": "http://andrew.github.com"
   },
   "engines": {
-    "node": ">=0.10.0"
+    "node": "^6.9 || ^8.9 || ^9 || ^10"
   },
   "main": "lib/index.js",
   "nodeSassConfig": {


realityking:
It's mostly machine readable documentation - npm will just print a warning, though yarn will actually error out when used on an unsupported Node.js version.

I followed  your suggestion partially and changed it to `^6.9 || ^8.9 || >= 9` because
- Node.js 6.5 actually contains a few new features around ES2015 support that v5 could then use at some point of time
- Excluding Node.js 11 - which will be released in 1.5 years - seems like a bad idea.

xzyfer:
The version choices seem arbitrary. Our support policy will be [Active LTS and Current](https://github.com/nodejs/Release#release-schedule). We don't get to choose which minor/patch we choose to support because it's convenient to us. 

We either support the whole range, or we start support from the current version with it went LTS (usually a couple months after the initial release). I was unable to find a list of at what specific version a release went LTS. For now please keep it at `^6 || ^8`. We'll cross the JS feature bridge when we come to it.

Additionally `>= 9` is too broad. We'll have dropped 9 before start supporting 11. Please stick to the suggested `^9 || ^10`, it's far easier for us to delete/update without thinking.

realityking:
Sorry, I should've mentioned how I arrived at those minor versions.

6.9 and 89 _are_ the releases where those versions transitioned into LTS. See these blog posts for reference:
- https://nodejs.org/en/blog/release/v6.9.0/
- https://nodejs.org/en/blog/release/v8.9.0/

As for the second point, at least for users using `yarn` this would mean any user using Node.js 11 could not use node-sass without you releasing a new version.

xzyfer:
Nice one, I was trying to find those blog post but missed it.

>any user using Node.js 11 could not use node-sass without you releasing a new version

This is mostly true regardless because we wouldn't have the binaries ready. 
We'd do a major bump between the 9 EOL and 11 Stable, in order to drop 9 support.
Users who want to live on the bleeding edge can use the `--ignore-engines` flag.

realityking:
Fair point about the binaries. Changed to `^6.9 || ^8.9 || ^9 || ^10`

xzyfer:
Looks like travis was having issues. Kicked off a rebuild.

xzyfer:
Thanks again ❤️

saper:
I wonder if we shouldn't add this to 4.x. too - even when using node 10 yarn has no way of knowing it needs to install newer version than specified

PR #2323, opened 2018-04-05
Move lodash.mergewith from dependencies to dev-dependencies.

realityking:
It's only used in a test.

**Note:** This is true in 4.8.3, if you plan on making another 4.x release I could rebase this branch against master instead.


xzyfer:
💓

nschonni:
:+1: might be worth cherry-picking this to master if we do one last 4.x release

PR #2349, opened 2018-04-25
Add Node 10 to CI

xzyfer:
:tada:


xzyfer:
It'll be interesting to see if people run into issues then.

Issue #2355, opened 2018-04-27
Security issue in dependencies


noraj:
## Update

Resolved in node-sass@4.9.1. Upgrade to quiet npm.

***

hoek@2.16.3 is vulnerable to [CVE-2018-3728](https://nvd.nist.gov/vuln/detail/CVE-2018-3728)

for node-sass the problem comes from requiring `request@2.79.0` in the package.json

dependency tree is as follow for 4.8.3
```
node-sass@4.8.3
|-request@2.79.0
 |-hawk@3.1.3
  |-hoek@2.16.3
```

and is the same for 4.9.0:

```
node-sass@4.9.0
|-request@2.79.0
 |-hawk@3.1.3
  |-hoek@2.16.3
```

# Fix

To fix this request@2.82.0 or superior is required.

# Context

- NPM version (`npm -v`): 5.8.0
- Node version (`node -v`): v9.11.1
- Node Process (`node -p process.versions`):
```
{ http_parser: '2.8.0',
  node: '9.11.1',
  v8: '6.2.414.46-node.23',
  uv: '1.19.2',
  zlib: '1.2.11',
  ares: '1.13.0',
  modules: '59',
  nghttp2: '1.29.0',
  napi: '3',
  openssl: '1.0.2o',
  icu: '61.1',
  unicode: '10.0',
  cldr: '33.0',
  tz: '2018c' }
```
- Node Platform (`node -p process.platform`): linux
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`):
```
node-sass	4.9.0	(Wrapper)	[JavaScript]
libsass  	3.5.4	(Sass Compiler)	[C/C++]
```
- npm node-sass versions (`npm ls node-sass`):
```
├─┬ gulp-sass@4.0.1
│ └── node-sass@4.9.0  deduped
└── node-sass@4.9.0 
```

# Related issues

request:

https://github.com/request/request/issues/2926
https://github.com/request/request/issues/2874

node-sass:

https://github.com/sass/node-sass/issues/2352
https://github.com/sass/node-sass/issues/2288
https://github.com/sass/node-sass/issues/2262
https://github.com/sass/node-sass/issues/2252
https://github.com/sass/node-sass/pull/2170
https://github.com/sass/node-sass/pull/2256

# Problem

xzyfer in #2352 
> It cannot be fixed without break node < 4 support

I also see in #2288 that the problem is solved in node-sass v5.

[...]

Issue #2288, opened 2018-03-14
Security issue: Prototype pollution attack(Introduced by the request@2.79.0 => hawk@3.1.3 => hoek@2.16.3)


MichaelTsengLZ:
The latest version node-sass@4.7.2 uses request@~2.79.0. However, request@~2.79.0 has a vulnerability which is introduced by hoek@2.16.3. More information is here:

https://nodesecurity.io/advisories/566
https://snyk.io/vuln/npm:hoek:20180212

Could you please update request to the latest version to solve the vulnerability? Thank you so much.

- NPM version (`npm -v`):  5.4.2
- Node version (`node -v`): v8.8.1
- Node Process (`node -p process.versions`):
```
{ 
  http_parser: '2.7.0',
  node: '8.8.1',
  v8: '6.1.534.42',
  uv: '1.15.0',
  zlib: '1.2.11',
  ares: '1.10.1-DEV',
  modules: '57',
  nghttp2: '1.25.0',
  openssl: '1.0.2l',
  icu: '59.1',
  unicode: '9.0',
  cldr: '31.0.1',
  tz: '2017b' 
}
```
- Node Platform (`node -p process.platform`): darwin
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`):
```
node-sass	4.7.2	(Wrapper)	[JavaScript]
libsass  	3.5.0.beta.2	(Sass Compiler)	[C/C++]
```
- npm node-sass versions (`npm ls node-sass`): 
```
cla-assistant@1.4.1 /Users/microsoft/Michael/MicrosoftRepo/cla/cla-assistant
└─┬ node-sass-middleware@0.11.0
  └── node-sass@4.7.2
```

nschonni:
Duplicate of a bunch of existing tickets

philwareham:
Just a heads-up: GitHub is now flagging up this vulnerability on any repos that have Node-Sass in the package-lock.json (at least in Node-Sass 4.9.0).

asommer70:
@philwareham I installed the v5 branch with:

```
npm install --save https://github.com/sass/node-sass/tarball/v5
```

And that fixed the hoek problem due to the request version being updated in node-sass.  Everything seems to work fine for me, but my project is pretty simple as far as SASS is concerned.

Your mileage may vary...

jgmcelwain:
Any chance we get the v5 branch on npm as `node-sass@next`?

xzyfer:
[...]

PR #2356, opened 2018-04-27
require request 2.85.1

paulschreiber:
request 2.85.1 requires hawk 6.0.2, which requires boom 4.x.x, which requires hoek 4.x.x. (hoek 2.16.3 has a security vulnerability).

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "nan": "^2.10.0",
     "node-gyp": "^3.3.1",
     "npmlog": "^4.0.0",
-    "request": "~2.79.0",
+    "request": "~2.85.1",
     "sass-graph": "^2.2.4",
     "stdout-stream": "^1.4.0",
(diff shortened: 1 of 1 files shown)


paulschreiber:
If you don't want to use current `request`, how do you propose getting a current hoek? How will you address #2355?

xzyfer:
Please respect our time by search for issues before opening a new one. Your questions have all been answered many times.

paulschreiber:
I see from #2288 that this has been known for at least six weeks. Sigh.

xzyfer:
It's know and unpatched on purpose because request broke semver by dropping
node < 4 support.

YasharF:
For anyone that is looking at this issue, the dependency upgrade is in the works with the release of node-sass 5.  See https://github.com/sass/node-sass/pull/2312

Issue #2370, opened 2018-05-05
link to troubleshooting guide returns 404.


rupav:
On https://www.npmjs.com/package/node-sass link to `troubleshooting guide` returns [404](https://www.npmjs.com/package/TROUBLESHOOTING.md#installing-node-sass-4x-with-node--4) .

xzyfer:
Thanks for the report. Npm scrapes readme but doesn't adjust the link when
displaying it in their site. We need to update the links on the readme to
be absolute URLs to our GitHub page.

This a good first contribution for anyone interested.

rupav:
I am making a PR then 😃

xzyfer:
Fixed in #2371

PR #2371, opened 2018-05-06
Updated links to absolute path instead of relative

rupav:
solves #2370 
with relative path to troubleshoot guide, npm was unable to direct to correct page. Now its corrected using absolute path.


rupav:
@xzyfer its corrected now.

nschonni:
👍 thanks!

PR #2379, opened 2018-05-13
Update issue templates

nschonni:


nschonni:
Was trying to add a 4th issue template type, but apparently it's restricted to only Bug, Feature, and Custom right now. This was a second "Custom" so it tried to overwrite it.
I'm just going to add this content into the Bug template instead

Issue #2022, opened 2017-06-16
Compile sass Error: File to read not found or unreadable


sbfkcel:
First make sure the file is there.

IDE：vscode
node：7.10.0
node-sass：4.5.3
system：win10

When the sass file is modified for the first time will be prompted to save the error. Not for the second time to save the normal ... ...

Using notepad.exe also does not cause the problem.

After testing, the problem does not occur on other operating systems.

xzyfer:
This sounds like a filesystem issue that's out of our control. See this documentation on configuring your editor to do atomic writes - https://webpack.js.org/guides/development/#adjusting-your-text-editor

marcosbozzani:
I've created a PR that fix this problem: https://github.com/sass/node-sass/pull/2386

If want to try this fix, you can patch your node-sass local copy replacing 'node_modules/node-sass/lib/render.js' with this version: https://github.com/marcosbozzani/node-sass/blob/bug-vscode-watch/lib/render.js

The diff is here: https://github.com/marcosbozzani/node-sass/commit/266861bf5761dcafd5d53d123264b965c861887c#diff-a485abf5b8f49de7f313d7799df3faf4

Adimvicky:
Thanks @marcosbozzani ! it's totally working for me now.

maniac0s:
I have the same issue but that fix doesn't change anything.

node-sass	4.9.3	(Wrapper)	[JavaScript]
libsass  	3.5.4	(Sass Compiler)	[C/C++]

Fatal error: File to read not found or unreadable: /var/www/projectthemes.src/pager.src/sass/base
Completed in 0.762s at Wed Sep 19 2018 08:30:41 GMT+0200 (Central European Summer Time) - Waiting...

/var/www/projectthemes.src/pager.src/sass/base is there and readable to the user. I don't understand why this error appears it totally doesn't make any sense. sass/base is not a file, it's a directory
 ls -la sass/
total 28
drwxr-xr-x  7 user sudo 4096 Sep 14 11:10 .
drwxr-xr-x 10 user sudo 4096 Sep 19 08:29 ..
drwxr-xr-x  2 user sudo 4096 Sep 19 08:30 base
drwxr-xr-x  2 user sudo 4096 Sep 14 11:10 components
drwxr-xr-x  2 user sudo 4096 Sep 14 11:10 config
drwxr-xr-x  2 user sudo 4096 Sep 17 16:05 layout
[...]

Issue #1894, opened 2017-02-12
"File to read not found or unreadable" with -w under Windows


pocesar:
- NPM version (`npm -v`): 4.2.0
- Node version (`node -v`): v7.5.0
- Node Process (`node -p process.versions`): 
```
{ http_parser: '2.7.0',
  node: '7.5.0',
  v8: '5.4.500.48',
  uv: '1.10.2',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  modules: '51',
  openssl: '1.0.2k',
  icu: '58.2',
  unicode: '9.0',
  cldr: '30.0.3',
  tz: '2016j' }
```
- Node Platform (`node -p process.platform`): win32
- Node architecture (`node -p process.arch`):  x64
- node-sass version (`node -p "require('node-sass').info"`):
```
node-sass       4.5.0   (Wrapper)       [JavaScript]
libsass         3.5.0.beta.2    (Sass Compiler) [C/C++]
```
- npm node-sass versions (`npm ls node-sass`):
```
`-- node-sass@4.5.0
```

---

Randomly when using node-sass directly from command line (without gulp), it fails with the error:

```
=> changed: G:\www\styles\styles.scss
{
  "status": 3,
  "message": "File to read not found or unreadable: G:/www/styles/styles.scss",
  "formatted": "Internal Error: File to read not found or unreadable: G:/www/styles/styles.scss\n"
}
=> changed: G:\www\styles\styles.scss
Rendering Complete, saving .css file...
Wrote CSS to G:\www\styles\styles.css
Wrote Source Map to G:\www\styles\styles.css.map
```

the file exists all the time, the HDD is fine and healthy. it fails around 30% of the time, and I have to re-save the file for the watch to recompile. sometimes it fails 4-5 times in a row. the ownership of the files are fine, no symlinks are being used, if it matters, I'm using VSCode (but the problem happens with EmEditor as well)

tuanngominh:
I face the same issue with v4.5.0 on Windows 10. node-sass watch compiles for the first change then show error for the second change.

The same node-sass version run well on Mac OSX 10.11.6

mj420:
I also have this issue on Windows 10 and node v6.9.5 when calling from node script

```
...
function style() {
        sass.render({
            file: path.join(path_style, 'main.sass')
        }, function (err, result) {
            if (err)
                console.log(err);
            else
                fs.writeFileSync(
                    path.join(path_dist, `style.css`),
                    result.css
                );
[...]

PR #2386, opened 2018-05-18
Fix 'unreadable file error' in watch mode after a save in vscode

marcosbozzani:
This fix detects 'File to read not found or unreadable' and 'File to import not found or unreadable' errors that occasionally happens after a file save in Visual Studio Code when node-sass is running on watch mode.

If those errors are detected, a retry mechanism is applied. It verifies that the file actually exists and limits the maximum number of retries to 1000. In my local tests, it usually takes less than 5 retries. The limit is there as a safety measure to prevent infinite recursion and the value is high enough for those cases where the machine is really slow or it has to access a storage with high latency, like a remote one.

Those errors happen with 'node-sass (watch mode)', 'chokidar + node-sass' and 'node-sass-chokidar (watch mode)' too.

It is also possible to change the retry mechanism for one with exponential backoff. For example:

Change this code:
```js
if (isFileUnreadable(error) && retries < 1000) {
      retries++;
      sass.render(renderOptions, renderCallback);
}
```

For this one:
```js
if (isFileUnreadable(error) && retries < 4) {
      retries++;
      setTimeout(() => sass.render(renderOptions, renderCallback), Math.pow(retries, 5));
}
```

Related issues:
https://github.com/sass/node-sass/issues/1894
https://github.com/sass/node-sass/issues/2022
https://github.com/Microsoft/vscode/issues/20491
https://github.com/facebook/create-react-app/issues/2531
https://github.com/michaelwayman/node-sass-chokidar/issues/14
https://github.com/michaelwayman/node-sass-chokidar/issues/22
https://stackoverflow.com/questions/50395998/vscode-wont-work-with-filewatchers


marcosbozzani:
Sorry if I wasn't clear enough in the PR. I know it's not node-sass fault. Yet this cause a very annoying issue for node-sass users. This could serve as a temporary workaround that could be removed once the real issue was solved. It doesn't affect those editors that do that right thing and mitigate the problem until the other editors can be fixed.

Chokidar fails as well (chokidar + node-sass and node-sass-chokidar), as you can see in the related issues. :confused:

I'm just sharing the solution that worked for me, so anyone can use it. :wink:

If this is really not the right place for this patch, I'm sorry for the trouble. :sweat_smile:

maartenraes:
@marcosbozzani worked form me by replacing the file localy, many thanks! 👍

seradork:
In my environment it did not work well.
I modified it as follows and it worked.

line 126-129:
```js
    if (unreadable) {
      var dir = path.dirname(error.file);
      var file = error.message.split('not found or unreadable: ')[1].trim().slice(0,-1);
      var fileDir = path.dirname(file);
      var fileName = path.basename(file);
      return (
        fs.existsSync(path.join(dir,fileDir,fileName)) ||
        fs.existsSync(path.join(dir,fileDir,"_" + fileName)) ||
        fs.existsSync(path.join(dir,fileDir,fileName + ".scss")) ||
        fs.existsSync(path.join(dir,fileDir,"_" + fileName + ".scss"))
      );
    }
```

- win7 64bit
- node-sass 4.9.2
- vscode 1.25.1

cyberpro4:
Just leaving here a simple solution found.

Since the problem is a "race condition" between the editor save operation and the watch itself, adding --use-polling option will break the concurrence. For ex:

`node-sass-chokidar src/ -o src/ --watch --recursive --use-polling --polling-interval 1000`

Ness70:
@marcosbozzani, If I may ask, where and in which file do you add this code.

ehindola:
please anybody how do i use this  change the retry mechanism for one with exponential backoff?

MattixNow:
Up ?

ayofef:
where do i paste this line of code:
if (isFileUnreadable(error) && retries < 4) {
      retries++;
      setTimeout(() => sass.render(renderOptions, renderCallback), Math.pow(retries, 5));
}

Agasthian:
@ayofef  you can patch this code in your project's node-sass local 
Location => 'node_modules/node-sass/lib/render.js' 
Line no => 114 in render.js file.
This solution works for me.

nschonni:
You could look at using something like https://www.npmjs.com/package/patch-package to apply patches like this. (Haven't tried, just searched for "npm patches")

ayofef:
Thank you

wfabio:
> @marcosbozzani worked for me by replacing the file localy, many thanks! 👍

Hello Marcos, Can you, please, tell me how to replace this file? I have no idea how to do it.

marcosbozzani:
@wfabio, locate the file `node-sass/lib/render.js` inside your local `node_modules` folder and replace with this code:

https://github.com/marcosbozzani/node-sass/blob/bug-vscode-watch/lib/render.js

You may need to redo this if you reinstall the library with `npm` or `yarn`

saper:
Can somebody explain me what exactly vscode is doing? What is the sequence of operations we are seeing when saving the file?

eastlin7:
I added this to node-sasss and it fixed my issue on my windows 10 machine.

saper:
@maartenraes can you come up with some reasonable value lower than 1000 ?

saper:
Could anyone experiencing this add the following patch to the libsass code (`src/libsass/src/file.cpp`) then rebuild node-sass and see what error message is generated?

```
diff --git a/src/libsass/src/file.cpp b/src/libsass/src/file.cpp
index 32d4a7c63..ead19c947 100644
--- a/src/libsass/src/file.cpp
+++ b/src/libsass/src/file.cpp
@@ -417,7 +417,10 @@ namespace Sass {
         if (rv > 32767) throw Exception::OperationError("Path is too long");
         if (rv == 0) throw Exception::OperationError("Path could not be resolved");
         HANDLE hFile = CreateFileW(resolved, GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
-        if (hFile == INVALID_HANDLE_VALUE) return 0;
+        if (hFile == INVALID_HANDLE_VALUE) {
+             std::cerr << "Error opening file:" << resolved << "LastError was [" << GetLastError() << "]" << std::endl; return 0; 
+            return 0;
+        }
         DWORD dwFileLength = GetFileSize(hFile, NULL);
         if (dwFileLength == INVALID_FILE_SIZE) return 0;
         // allocate an extra byte for the null char
```

marcosbozzani:
A ready to use patch equivalent to this PR is available here: https://github.com/marcosbozzani/patch-node-sass-watch

PR #2417, opened 2018-06-14
dropping node-gyp as direct dependency

Rohithzr:
referencing 
https://github.com/nodejs/node-gyp/issues/1439#issuecomment-397255452
and 
https://github.com/sass/node-sass/issues/2355#issuecomment-397395712

node-gyp should be used as provided with the node installations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ before_install:
   - g++ --version
 
 install:
+  - npm install -g node-gyp
   - npm install
 
 script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -75,6 +75,7 @@
     - ps: Install-Product node $env:nodejs_version $env:platform
     - node --version
     - npm --version
+    - npm install -g node-gyp
     - npm install
 
   test: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -75,7 +75,6 @@
     - ps: Install-Product node $env:nodejs_version $env:platform
     - node --version
     - npm --version
-    - npm install -g node-gyp
     - npm install
 
   test: off
@@ -160,6 +159,7 @@
     - ps: Install-Product node $env:nodejs_version $env:platform
     - node --version
     - npm --version
+    - npm install -g node-gyp
     - npm install
 
   test_script:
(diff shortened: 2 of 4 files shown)


Rohithzr:
referencing this https://github.com/sass/node-sass/issues/2355#issuecomment-397395712 probably there needs a discussion on how node-gyp should be used. @bnoordhuis maybe you can shed some light on using the implicit node-gyp.

bnoordhuis:
If there's a binding.gyp in the top-level directory, then npm should automatically start node-gyp.

Rohithzr:
@bnoordhuis there is a binding.gyp, but I am still unable to access `node-gyp` in the [scripts/build.js](https://github.com/sass/node-sass/pull/2417/commits/4dfd395f8b86dde0fc79d792f8e349e9975b7a65#diff-ea1c4abd09127cfb80f898b2bf12b0a4) I am now planning to find the npm installation and locate the node-gyp.js manually.

but i am still wondering why am i not able to access node-gyp.

bnoordhuis:
Ah, I understand now.  What I said about npm starting node-gyp is only true when there is no postinstall task in package.json but node-sass has one.

I think you should still be able to exec `node-gyp rebuild` though.  It's supposed to be on the PATH that the build script is spawned with.

saper:
Does this work with yarn as well? Why dropping node-gyp dependency improves things?

Rohithzr:
@bnoordhuis not really, i tried running node-gyp rebuild through the scripts it failed

@saper not sure yet, node-gyp is not a direct dependency for yarn, it was attempted because there are vulnerability issues in node-gyp and according to @bnoordhuis who is a node-gyp maintainer, it is recommended to use the node-gyp that was installed by npm.

the problem is that node-sass@5 is the only way to use the next node-gyp version as both of them will drop support for node<4.

but i was thinking if there was a way to use the latest possible version of node-gyp that can be allowed before the release of node-sass@v5

saper:
@Rohithzr yes but a project using yarn exclusively will not have node-gyp available. I think that explicit dependency is better.

Rohithzr:
@saper i know, i am just trying to find a way for it to work with yarn as well, if it doesn't then obviously it cannot be merged.

Rohithzr:
difficult to support it for yarn and some versions of npm. closing

PR #2128, opened 2017-10-31
Fix Some Leaks

stefanpenner:
[FIXES #2098]

We noticed our ember-cli builds for linkedin.com were consuming ~10gb of memory, [after some investigation](https://github.com/sass/node-sass/issues/2098) node-sass (via eyeglass) was identified as the primary culprit. With this PR, memory usage dropped by from ~10gb -> ~1gb (saving about 9gb of ram)

**NOTE: This does not fix all the leaks, more PR's to come**

### Fixes so far:

* `create_string` must be paired with a free (this catches a few of the easy ones)
  * number constructor's unit (`sass_make_number` copies the unit)
  * string constructor's value (`sass_make_string` also copies the string)
  * Bindings `ExtractOptions` function signature (`sass_make_function` copies the `signature`)
* `SassTypes::Value` are now subclasses of `Nan::ObjectWrap`, and we utilize its machinery for memory management. This ensures our wrapper objects no longer leak [Nan::ObjectWrap docs](https://github.com/nodejs/nan/blob/master/doc/object_wrappers.md#api_nan_object_wrap)

### Tests:
* add some isolated memory-test examples, which are handy to diagnose/fix memory leaks
* back-port tests for `sass.types.*`
* in sass-eyeglass/eyeglass
* in the build system for Linkedin.com

### Notes:

I realize this PR is quite large, the only drastic change was utilizing Nan::ObjectWrap memory management machinery. The rest is a mixture of tests, and cleanup to simplify the memory issues.

I look forward to you code reviews!

PR #2440, opened 2018-07-08
[WIP] N-API

xzyfer:
Heavy work in progress of replacing NAN with N-API.

Attempting to combine the prior work https://github.com/boingoing/node-sass/pull/1 with the drift in master, namely #2128 #2298.

### DIFF
--- a/src/binding.cpp
+++ b/src/binding.cpp
@@ -1,4 +1,4 @@
-#include <nan.h>
+#include <napi.h>
 #include <vector>
 #include "sass_context_wrapper.h"
 #include "custom_function_bridge.h"
@@ -25,28 +25,30 @@ union Sass_Value* sass_custom_function(const union Sass_Value* s_args, Sass_Func
   CustomFunctionBridge& bridge = *(static_cast<CustomFunctionBridge*>(cookie));
 
   std::vector<void*> argv;
-  for (unsigned l = sass_list_get_length(s_args), i = 0; i < l; i++) {
+  for (size_t l = sass_list_get_length(s_args), i = 0; i < l; i++) {
     argv.push_back((void*)sass_list_get_value(s_args, i));
   }
 
   return bridge(argv);
 }
 
-int ExtractOptions(v8::Local<v8::Object> options, void* cptr, sass_context_wrapper* ctx_w, bool is_file, bool is_sync) {
-  Nan::HandleScope scope;
+int ExtractOptions(napi_env e, napi_value options, void* cptr, sass_context_wrapper* ctx_w, bool is_file, bool is_sync) {
+  // TODO: Enable napi_close_handle_scope during a pending exception state.
+  //Napi::HandleScope scope(e);
 
   struct Sass_Context* ctx;
 
-  v8::Local<v8::Value> result_ = Nan::Get(
-    options,
-    Nan::New("result").ToLocalChecked()
-  ).ToLocalChecked();
-  if (!result_->IsObject()) {
-    Nan::ThrowTypeError("\"result\" element is not an object");
+  napi_value result_;
+  CHECK_NAPI_RESULT(napi_get_named_property(e, options, "result", &result_));
+  napi_valuetype t;
+  CHECK_NAPI_RESULT(napi_typeof(e, result_, &t));
+
+  if (t != napi_object) {
+    CHECK_NAPI_RESULT(napi_throw_type_error(e, nullptr, "\"result\" element is not an object"));
     return -1;
   }
 
-  ctx_w->result.Reset(result_.As<v8::Object>());
+  CHECK_NAPI_RESULT(napi_create_reference(e, result_, 1, &ctx_w->result));
 
   if (is_file) {
     ctx_w->fctx = (struct Sass_File_Context*) cptr;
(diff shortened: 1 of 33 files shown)


xzyfer:
This is getting pretty close. There's a segfault in one of the tests I still need to debug.

It's worth noting this unwinds the work @stefanpenner's to implement `Nan::ObjectWrap` (#2128) largely because I couldn't get my head around how to make `Napi::ObjectWrap<T>` work. May this is something you're interested in looking at again @stefanpenner?

xzyfer:
This potentially unwinds @kkoopa's work to propagate async context (#2295) because I couldn't find the equivalent in N-API.

xzyfer:
I have it compiling and running most of the tests. There is a segfault that seems semi random. I'll need to investigate.

I've update the the install scripts to use the N-API version instead of module version. Have confirmed binary compiled Node 10 executes on Node 6, 8, 9, 10. Although 9 displays an annoying warning that people will definitely files issues about because...

xzyfer:
I've resolved the segfault. All tests are passing locally (OSX) on all Node LTS and current.

We'll need to rejig CI a bit since the binary no longer builds on Node < 10.

xzyfer:
We have our first green N-API builds 
https://travis-ci.org/sass/node-sass/builds/404487187

Just for kicks the current hacked CI setup shows that binaries built on Node 10 work on other LTS Node.

xzyfer:
@nschonni note this means we probably can't have node-gyp fallback to build

nschonni:
No, the node-pre-gyp handles napi vs the module version in it's download and fallback

xzyfer:
Sorry what I mean is that with N-API we can no longer compile the binary on Node < 10. So if the download fails there's no point node-pre-gyp falling back to build locally.

nschonni:
Ah, gotcha, but I think it is backported to v6 & v8. I think older stuff may be able to have support through https://github.com/nodejs/node-addon-api/blob/master/doc/setup.md but I'm not :100:

xzyfer:
In my experiments binaries built on 10 run back until 6, but I couldn't get
it to build on < 10 because the have different napi versions. I could be
missing something though

xzyfer:
I have a call with some napi folk later in the week. I'll get some
clarification

stefanpenner:
> It's worth noting this unwinds the work @stefanpenner's to implement Nan::ObjectWrap (#2128) largely because I couldn't get my head around how to make Napi::ObjectWrap<T> work.

As long as the [memory leak scripts](https://github.com/sass/node-sass/tree/master/memory-tests) I wrote, don't demonstrate a leak we should be ok

stefanpenner:
@xzyfer I can try and find some time later this week, to see if this regresses any of the issues I saw earlier.

Switching to the napi is nice, as it is much nicer then NaN. But we should be sure not to regress on those leaks.

xzyfer:
>As long as the memory leak scripts I wrote, don't demonstrate a leak we should be ok

I can try running them locally as a verification.

> I can try and find some time later this week, to see if this regresses any of the issues I saw earlier. Let me know if I should.

This would be great. I'm not planning to change the N-API implementation at the moment.

> Assuming supported nodes support the needed napi's

It supports our reduced Node support as of v5 which is where I'd hope to land this.

realityking:
Looking at https://nodejs.org/api/n-api.html#n_api_n_api_version_matrix, I think the minimum version supported has to be raised to 6.14.

Maybe TravisCI should be adjusted to test on the minimum supported version of each LTS line?

xzyfer:
Hey @realityking, a couple updates from today
- I had call with some of the N-API folks this morning, and 
- had a chat with @nschonni 

It's now my expectation that we'll only rollout the N-API binding to Node 10+. Requiring minimum Node versions is simply too fragile. So moving forward we will be using nan for Node < 10 and N-API for Node 10+.

realityking:
That's very interesting

> Requiring minimum Node versions is simply too fragile

Is that because npm doesn't bail out when the the minimum node version isn't met, instead just printing a warning?

xzyfer:
Partly, but largely because the typical user of node-sass either isn't in
control of their node version or doesn't know how/why to change versions.

It also introduces another barrier to adoption. We're really focused on
making the installation process more forgiving. It's the single largest
source of issues.

aduh95:
Node.js team has added support for N-API on Node 6 and Node 8, what difference does that make for users to use N-API or `nan`? If the concern is about users using older versions, they can keep using `node-sass@4` (SEMVER major are about breaking changes after all), I don't really see the point of keeping `nan` when we could use N-API everywhere.

xzyfer:
Picking this up again as we plan to drop support for Node < 10 in the next semver major.

src/binding.cpp
```diff
@@ -65,100 +67,175 @@ int ExtractOptions(v8::Local<v8::Object> options, void* cptr, sass_context_wrapp
...
-  ctx_w->linefeed = create_string(Nan::Get(options, Nan::New("linefeed").ToLocalChecked()));
-  ctx_w->include_path = create_string(Nan::Get(options, Nan::New("includePaths").ToLocalChecked()));
-  ctx_w->out_file = create_string(Nan::Get(options, Nan::New("outFile").ToLocalChecked()));
-  ctx_w->source_map = create_string(Nan::Get(options, Nan::New("sourceMap").ToLocalChecked()));
-  ctx_w->source_map_root = create_string(Nan::Get(options, Nan::New("sourceMapRoot").ToLocalChecked()));
+    indent_width_len,
+    indent_type_len == 1 ? '\t' : ' '
+  ).c_str());
+
+ napi_value propertyLinefeed;
```
saper: indentation problem?

src/callback_bridge.h
```diff
@@ -2,30 +2,32 @@
 #define CALLBACK_BRIDGE_H
 
 #include <vector>
-#include <nan.h>
 #include <algorithm>
 #include <uv.h>
```
saper: Including `<uv.h>` breaks guarantees of ABI stability offered by N-API. https://github.com/sass/node-sass/pull/2312#issuecomment-552043349

src/sass_types/value.h
```diff
@@ -1,41 +1,17 @@
...
 
-#include <nan.h>
 #include <sass/values.h>
+#include "../common.h"
+#include <napi.h>
 
 namespace SassTypes
 {
   // This is the interface that all sass values must comply with
-  class Value : public Nan::ObjectWrap {
```
saper: We probable need to rebuild this without inheritance at all. Maybe this should become a template? We definitely need https://github.com/nodejs/node-addon-api/blob/master/doc/object_wrap.md here

gabrielschulhof:
Is anyone still working on this 👋

PR #2441, opened 2018-07-09
[WIP] Node pre gyp for v5

nschonni:
Think this still needs some cleaning up. I was going to do this against the v5 branch but rebased from master and haven't gone back to cherry pick it so it is clean there again.
Think I'd still like to incorporate the libsass version into the binding name.


xzyfer:
A couple things I'd like to confirm (and update the documentation accordingly):
- [ ] we can continue to support fire walled environments (`SASS_BINARY_*`)
- [ ] we can continue to support a user defined libsass library 
  - https://github.com/sass/node-sass/blob/master/scripts/build.js#L60

nschonni:
> we can continue to support fire walled environments (SASS_BINARY_*)

I believe that would be through https://github.com/mapbox/node-pre-gyp#download-binary-files-from-a-mirror

nschonni:
Think there is a problem with the download for Node 0.10 and 0.12 because of a use of `const` in the `tar` dependency. Probably fine for v5 when we drop older Node though

nschonni:
Closing this as discussed on Slack, will continue discussion on #2452

saper:
Can you give short info for those no longer on slack..?

nschonni:
Just rebasing it to the v5 branch. The other issues about NAPI and NAN are also discussed on the other issue

PR #2442, opened 2018-07-09
chore: Remove memory-tests

nschonni:
Just noticed these when I trying opening up the ESLint globs. These look specific to the work down in 4.7 memory performance stuff. Since we don't actually run this as part of any part of the build process I think they can be dropped.


xzyfer:
Let's keep these. I'm using them now for the N-API work to prevent
regressions.

PR #2452, opened 2018-07-19
feat: Switch to node-pre-gyp

nschonni:
Just rebased this, haven't tested it so I'll see what the CI says


xzyfer:
So I ran into these failures during my N-API work and was scratching my head for a good while. Turns out a bunch of tests have always been broken due to a subtle bug introduced into the CLI test 4 years ago (https://github.com/sass/node-sass/commit/535b5249c9b2b701f7e56578ad7f76fb66b5da0e#diff-583d616f4b01bf1fa153e71c0cef7de7R98).

For some reason a recent change to some Node LTS versions, or AppVeyor itself has caused the run order of tests to change which surfaced this error.

I have applied a patch the v5 branch (https://github.com/sass/node-sass/pull/2312/commits/4b05387bacbb7bbe47d7d18f22add770808be98c). Re-rebase this PR and your CI troubles should go away.

xzyfer:
Given our chat is it correct that is blocked on https://github.com/mapbox/node-pre-gyp/issues/402?

nschonni:
Depends if we want to try and support both at the same time

xzyfer:
Yeah we do. With a minimum napi version of 3

springmeyer:
> Given our chat is it correct that is blocked on mapbox/node-pre-gyp#402?

This is now released in node-pre-gyp@0.11.0

nschonni:
Came across https://github.com/JCMais/node-libcurl which has a good node-pre-gyp setup, but they're not doing N-API right now

codler:
Will this replace `node-gyp`?

nschonni:
@codler no, it is just as an alternate to our custom build scripts to pull down pre-built binaries. All native modules in node use node-gyp as the fallback build.

PR #2477, opened 2018-08-14
don't pack the test directory into the npm-module

Meir017:
we don't really need the test files in order to use this module 😉

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "lib",
     "scripts",
     "src",
-    "test",
     "vendor"
   ],
   "keywords": [


nschonni:
@xzyfer is this needed for the Canary in the Mine testing by the NodeJS build team?

xzyfer:
Correct. We need to publish tests for https://github.com/nodejs/citgm

Meir017:
Is that module used to run the test in the CI?

xzyfer:
It's a system used by the Node core team to ensure they don't accidentally
break important modules before doing a Node version release.

Meir017:
Hmm, express don't pack their tests https://github.com/expressjs/express/blob/master/package.json#L84

Issue #2491, opened 2018-09-10
Imported files outside the source directory built when changed in watch mode


paulcpederson:
I have a fix for this, will try to pr shortly. It's kind of an edge case, but replication is fairly straight forward.

Assume we have the following file tree:

```
src/
  file1.scss
  file2.scss
dist/
library/
  include.scss
```

And `file1` imports the library, like this:

```
@import "library/include";
```

We build all the sass files in `src/` with the CLI using `node-sass src/ -o dist/". This works correctly.
If, we watch for changes and change the files in `src`,  it also still works correctly. 

But if we change `library/include.scss`, the CLI will _build a new file at `library/include.css`_. Then, node-sass will throw an error:

```
It's not clear which file to import for '@import "library/include". 
Candidates:
    library/include.scss
    library/include.css
```

Even though it is outside the source directory, it's included in the watched files, and because it  doesn't have an underscore preceding it's name, we render it to the default location (right next to itself). 

The fix is pretty simple, basically when using a directory for multiple file compilation, when files change, check that they are actually inside the directory before rendering. I added multiple file compilation once upon a time so this is mostly my fault 😬 I'll try to get a pr for this fix in soon.

---

- NPM version (`npm -v`): 6.4.0
- Node version (`node -v`): v8.2.1
- Node Process (`node -p process.versions`):
```
{ http_parser: '2.7.0',
  node: '8.2.1',
  v8: '5.8.283.41',
  uv: '1.13.1',
  zlib: '1.2.11',
  ares: '1.10.1-DEV',
  modules: '57',
  openssl: '1.0.2l',
  icu: '59.1',
  unicode: '9.0',
  cldr: '31.0.1',
  tz: '2017b' }
```
- Node Platform (`node -p process.platform`): darwin
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`):
```
node-sass	4.9.3	(Wrapper)	[JavaScript]
libsass  	3.5.4	(Sass Compiler)	[C/C++]
``` 
- npm node-sass versions (`npm ls node-sass`): node-sass@4.9.3

paulcpederson:
pr submitted

swervo:
I think I may be seeing a related issue. My command:

`node-sass --include-path 'node_modules/@salesforce-ux/design-system/scss' --source-map true sass/ -wo app/styles`

[...]

Issue #2479, opened 2018-08-18
CLI watcher crashes when it encounters empty tags/files or missing brackets


Jacky2611:
- NPM version (`npm -v`): 5.6.0
- Node version (`node -v`): 8.11.3
- Node Process (`node -p process.versions`):
- Node Platform (`node -p process.platform`): win32
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`): node-sass 4.9.3, libsass 3.5.4
- npm node-sass versions (`npm ls node-sass`): 4.9.3

I recently switched over to sass and am now in the process of updating my build scripts. However, I can't seem to get the following command to work.
"watch-sass": "node-sass src/ -o dist/ -w",
It works fine in most cases, but whenever I create an empty .scss file it crashes for good and I have to manually restart the watch process.
 
An example for a scss file that causes such a crash would be the following:
```
h1 {

}
```
I have also managed to produce this problem by using sass syntax instead of scss inside one of my files or by not closing one of my brackets before saving.

Here is the error log:
```
> node-sass src/ -o dist/ -w

D:\Development\Projekte\AssertShareWeb\node_modules\scss-tokenizer\lib\tokenize.js:265
                if (ident.test(css) && (ident.lastIndex = pos || 1) && ident.exec(css).index === pos) {
                                                                                      ^

TypeError: Cannot read property 'index' of null
    at tokenize (D:\Development\Projekte\AssertShareWeb\node_modules\scss-tokenizer\lib\tokenize.js:265:87)
    at Object.scss.tokenize (D:\Development\Projekte\AssertShareWeb\node_modules\scss-tokenizer\lib\entry.js:18:35)
    at parseImports (D:\Development\Projekte\AssertShareWeb\node_modules\sass-graph\parse-imports.js:4:26)
    at Graph.addFile (D:\Development\Projekte\AssertShareWeb\node_modules\sass-graph\sass-graph.js:69:17)
    at D:\Development\Projekte\AssertShareWeb\node_modules\sass-graph\sass-graph.js:54:13
    at arrayEach (D:\Development\Projekte\AssertShareWeb\node_modules\lodash\lodash.js:516:11)
    at Function.forEach (D:\Development\Projekte\AssertShareWeb\node_modules\lodash\lodash.js:9342:14)
    at new Graph (D:\Development\Projekte\AssertShareWeb\node_modules\sass-graph\sass-graph.js:53:7)
[...]

PR #2492, opened 2018-09-10
when watching entire directory, don't rebuild files outside said directory

paulcpederson:
#2491 - small fix to the CLI so that non-underscored files outside the watched directory (but imported as part of the tree) are not built.

@xzyfer I've written a test for this with a fixture but `skipped` it as it looks like you're skipping most of the async tests? Let me know if you want me to add that back in to the list.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -246,7 +246,9 @@ function watch(options, emitter) {
     });
 
     files.changed.forEach(function(file) {
-      if (path.basename(file)[0] !== '_') {
+      var srcDir = options.directory && path.join(process.cwd(), options.directory);
+      var isInclude = srcDir && file.indexOf(srcDir) === -1;
+      if (path.basename(file)[0] !== '_' && !isInclude) {
         renderFile(file, options, emitter);
       }
     });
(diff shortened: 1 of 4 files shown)


paulcpederson:
@nschonni do you know who could take a look at this?

saper:
It looks to me that https://github.com/sass/node-sass/issues/2491 https://github.com/sass/node-sass/issues/2504 https://github.com/sass/node-sass/issues/2560 and partially https://github.com/sass/node-sass/issues/2479 have the similar root cause.

We constantly maintain a graph of dependencies and we try to rebuild everything on that graph that does not start with underscore. 

This is probably not correct - we should maintain two trees:
* dependency graph
* expanded original input specification

We should react to all events on the dependency graph,  but only files on the original input specification should ever be compiled. 

Looking at that code:

https://github.com/sass/node-sass/blob/8d0accabd61ee5cb16248474f2989209b6200e50/bin/node-sass#L248-L252

we should not fire compilation of the *changed* file as a response to the `changed` event - only affected original files should be subjected to rendering.

Any thoughts @xzyfer , @nschonni ? If yes, this would mean a major change in our watching code.

PR #2517, opened 2018-10-20
Upgrade node 10

bc-zoharmuzafi:


nschonni:
@bc-zoharmuzafi your company is running an old fork of node-sass before Node 10 support was added. Please update to use the official releases instead of opening these PRs

PR #2553, opened 2018-12-04
Suppress -std=c++0x option for zOS

joransiu:
The njsc/c++ compiler used for Node.js on z/OS has C++11 support by
default.  The -std=c++0x option currently triggers a FSUM3210 error,
hence, updating gyp files to suppress use of this option on the
platform (similar to what's currently done for Windows).

Signed-off-by: Joran Siu <joransiu@ca.ibm.com>

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -80,7 +80,7 @@
             },
           },
         }],
-        ['OS!="win"', {
+        ['OS!="win" and OS!="zos"', {
           'cflags_cc+': [
             '-std=c++0x'
           ]
--- a/src/libsass.gyp
+++ b/src/libsass.gyp
@@ -103,7 +103,7 @@
              }]
           ]
         }],
-        ['OS!="win"', {
+        ['OS!="win" and OS!="zos"', {
           'cflags_cc+': [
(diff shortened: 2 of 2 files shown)


nschonni:
@xzyfer don't think this hurts (event if we don't support zOS), but I'm wondering about the upstream libsass change to swap this old flag for the newer `c++11` version

xzyfer:
I need to validate we can still compile on CentOS 5 with this change. I
vaguely recall that's why this hasn't been made before.

We've made this change in LibSass so we should do it in v5 for sure.

saper:
Around 3.5.0 (upgrade to node-gyp after 3.2.0) we have lost ability to build on CentOS 5.
I tried it for fun and there are a couple of things - to run tests, one needs to install git from non-official repositories; https://github.com links no longer work due to TLS issues; python 2.4.3 is the latest version, and gyp needs 2.5 in the version we are using as of 4.12.0

Therefore I believe we should drop this altogether, for all platforms.

nschonni:
These are dropped now for all platforms

Issue #2334, opened 2018-04-16
Disable "DEPRECATION WARNING"s


janrembold:
We're using `gulp-sass` package that is only a wrapper for `node-sass`. 

We encounter lots of deprecation warnings like:
```
DEPRECATION WARNING: Passing a string to call() is deprecated and will be illegal
in Sass 4.0. Use call(get-function("variable-exists")) instead.
```

This is a valid warning and is related to an old version of Susy Grid. But it is extremely annoying during our development process because it produces hundreds of warnings on each execution of the Sass task. 

It would be really great to be able to disable these warnings somehow. Or at least show them only once and not on each execution.

Is there a chance to do that or do you already plan to update this behavior?
Thanks in advance!

xzyfer:
There's no way for the compiler to differentiate between code you own and code you don't.

These deprecation warnings are important. They're telling you that your code will cease to work very soon, and you should take steps to resolve the warnings asap.

If it's code you don't control then we recommend opening issues on the code's repo, or submitting a PR to resolve the warnings.

There is simply no way to disable these warning. If you don't want to see them I suggest downgrading to an older version of Sass and never updating again.

xzyfer:
We'll look at what's involved in deduping the deprecation warnings but we can't make any promises at this time.

igorlakic:
I get the same warning when using `gulp-sass` with SingularityGS... I saw a [fix for the same problem](https://github.com/davidkpiano/sassdash/issues/37) in another project - can something like this be a good solution here too?

xzyfer:
Yes, this is exactly what needs to be done for the libraries causing these deprecation warnings. It should be easy enough to do for anyone with the time.

esteinborn:
@xzyfer I'm wondering if there is a way to print the file that those deprecation warnings are generated from during compilation? I have a large project with many references to functions this way, but only generating a few of the deprecation messages.

xzyfer:
So I don't believe we can know the file, and/or location of the offending import for various reasons. However I believe we can at least provide the path and name of the file being imported that is triggering the deprecation warning.

sebszocinski:
Any updates on this?

jwahler:
We have work around this issue by updating our susy grid to [2.2.14](https://www.npmjs.com/package/susy/v/2.2.14). All our warnings were produced by susy. Give it a try. good luck.

sebszocinski:
Thanks mate

Cheers,

SEB SZOCINSKI
— Web Developer

[...]

PR #2587, opened 2019-01-30
Added an environment variable to suppress DEPRECATION WARNS - Issue #2334

luiscla27:
Added env variable called `NODE_SASS_NO_DEPRECATION`, so everyone can be able to disable deprecation warnings.

Warnings will stop being shown by setting the variable `NODE_SASS_NO_DEPRECATION` to `0`

This is related to issue #2334, you may modify variable name and configuration value as you wish, the idea of this PR is just to give everyone _some hope_ of a possible solution.

### DIFF
--- a/src/libsass/src/error_handling.cpp
+++ b/src/libsass/src/error_handling.cpp
@@ -181,42 +181,48 @@ namespace Sass {
 
   void deprecated_function(std::string msg, ParserState pstate)
   {
-    std::string cwd(Sass::File::get_cwd());
-    std::string abs_path(Sass::File::rel2abs(pstate.path, cwd, cwd));
-    std::string rel_path(Sass::File::abs2rel(pstate.path, cwd, cwd));
-    std::string output_path(Sass::File::path_for_console(rel_path, abs_path, pstate.path));
+    if (!!process.env.NODE_SASS_NO_DEPRECATION && process.env.NODE_SASS_NO_DEPRECATION !== '0') {
+      std::string cwd(Sass::File::get_cwd());
+      std::string abs_path(Sass::File::rel2abs(pstate.path, cwd, cwd));
+      std::string rel_path(Sass::File::abs2rel(pstate.path, cwd, cwd));
+      std::string output_path(Sass::File::path_for_console(rel_path, abs_path, pstate.path));
 
-    std::cerr << "DEPRECATION WARNING: " << msg << std::endl;
-    std::cerr << "will be an error in future versions of Sass." << std::endl;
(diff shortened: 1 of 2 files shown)


luiscla27:
Checks fail because the NEW variable needs to be added to the test enviroment.

nschonni:
Sorry, we can't make changes to the files in the libsass folder as that is just pulled in from upstream

luiscla27:
So there's any way to achieve the same results by modifying something else?? this PR was intended as a suggestion, 

Where is the upstream folder @nsams ?

nschonni:
The upstream project is https://github.com/sass/libsass
I'm not 💯 but I think the next pull of libsass will have removed the things that the deprecation warnings were talking about

luiscla27:
Thank you @nschonni! the deprecation warnings are being shown on sass implementations (not sass itself), i'll try to make this PR there :)

xzyfer:
This is not a feature that will be accepted upstream. The acceptable
solution to silencing deprecation warnings to fix the deprecated code.

In this rare case we're rolling back the deprecation warning because there
isn't yet a reasonable way to resolve the warning.

luiscla27:
But @xzyfer ... as stated at issue #2334, how do i modify code that doesn't belong to me?? 

I don't even know which library is the one with the issue... i understand the need to force everyone to update their deprecated stuff, but that's not feasible. I think every developer is responsible for their own application and taking the risk to be outdated is on their own, i'll not make de PR, instead i'll just add a feature request to solve this.

xzyfer:
You modify the code by opening a PR. Leaving the code unfixed is a time
bomb. The code will stop working in a future releases. The deprecation
warnings a signals l to the community at large.

If you cannot find the code causing the deprecation warning that's a
different issue. We should make that easier. That's is a feature we would
consider. However controlling the output of deprecation warnings is not
something that will change.

luiscla27:
Most fwk's/libraries already expose warnings this same way, i understand the need to force everyone to update their deprecated stuff, as you said... the code will stop working on future releases. Thats same reason that every other lib "warns" developers of the risk. 

@xzyfer, the library responsibility is to provide a solution by upgrading the code... taking the risk to not upgrade is not something to concern to the libraries/fwk.

Tofandel:
But they're warning, so we get them once and then we know we'll need to update the codebase or a dependency when we want to update to the next major release, but in the minor releases there is so much things being deprecated that those warnings from other libs are just polluting the build process if we update minor, so yes a way to disable them would be welcome

PR #2592, opened 2019-02-05
Add .cirrus.yml for FreeBSD (amd64, i386) CI/CD

am11:
CirrusCI has [added FreeBSD support](https://forums.freebsd.org/threads/cirrus-ci-support-for-freebsd.68644/) in December 2018. Currently
only the amd64 images are provisioned.

`cbsd` is used to create an i386 jail environment. Kudos to Oleg
Ginzburg (@olevole), wouldn't have been possible without his
help (https://github.com/cbsd/cbsd/issues/367).

---

For each pull request, it triggers a job consists of 'node.js version x
architecture' matrix.
When a tag is pushed, it creates a release idempotently, and publishes
artifacts to the release.

The versions of node.js and npm packages are selected from package
management. This is because node.js official distribution does not
(yet) offer FreeBSD binaries and `nvm` spends too much time compiling
node.js and npm from source.

`pkg` does not support installation of old version of packages with
`pkg-install` command, so the `pkg-add` approach has been elected to
install older node.js. For that matter, the exact version of node.js
and npm packages are resolved using pattern matching.


am11:
A successful job run for normal commit looks like this: https://cirrus-ci.com/build/5765193800876032, for tag https://cirrus-ci.com/build/5684006109577216 and release artifacts get uploaded to GH artifacts in node-sass desired format: https://github.com/am11/node-sass/releases/tag/testbsd-x5.

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
...
+
+  env:
+    matrix:
+      - abi: freebsd:11:x86:64
+      - abi: freebsd:11:x86:32
+        jailName: j11i386
+        execPrefix: cbsd jexec jname=j11i386
+
+  env:
+    GH_API_TOKEN: ENCRYPTED[5e482f417304528148bb96eca8d030eacd6ab9972d482485fc4d42907283b995f658b351e0676e9493a37d815398f541]
```
am11: @xzyfer, it seems like we would need to login to CirrucCI with organization admin's account and encrypt the personal access token using the following view:

image

This is only used for publishing artifacts. 8-)

nschonni:
Nice!
Tagging @saper since he has been doing the FreeBSD builds so far

am11:
Thanks @nschonni! @saper, highly appreciate if you could review. :)

Was waiting for https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235412 review to go through, so we could use cbsd 12.0.4 from ports and clean up some code in the jail creation shell script. Currently, it is cloning the tag ref from cbsd repo for i386 jail creation. The longest job is less than eight minutes; create i386 jail, install tools, build node-sass and upload artifacts. `amd64` jobs are about four minutes on average.

nschonni:
@saper since you've been looking at the FreeBSD build again, maybe you can review and push any changes needed here

ltning:
If this is required for FreeBSD builds to be automated .. *poke* ?

nschonni:
@saper is this worth getting sorted out for the Node 14 builds?

appveyor.yml
```diff
@@ -166,7 +166,7 @@
       - nodejs_version: 11
         GYP_MSVS_VERSION: 2015
         APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-        
+
   install:
```
saper, line 170: spurious change

package.json
```diff
@@ -76,8 +76,8 @@
     "eslint": "^3.4.0",
     "fs-extra": "^0.30.0",
     "istanbul": "^0.4.2",
-    "mocha": "^3.1.2",
-    "mocha-lcov-reporter": "^1.2.0",
+    "mocha": "^3.5.3",
+    "mocha-lcov-reporter": "^1.3.0",
     "object-merge": "^2.5.1",
```
saper, line 81: unrelated change?

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
...
+        npmTxzKey: npm-6
+      - node_js: 10
+        nodeTxzKey: node10
+        npmTxzKey: npm-node10
+      - node_js: 8
+        nodeTxzKey: node8
+        npmTxzKey: npm-node8
+      - node_js: 6
+        nodeTxzKey: node6
+        npmTxzKey: npm-node6
```
saper, line 19: no longer there

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
+freebsd_instance:
+  image: freebsd-11-2-release-amd64
+
+freebsd_build_task:
+
+  env:
+    matrix:
+      - node_js: 11
```
saper, line 8: need to add node 13, 12

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
...
+        nodeTxzKey: node-11
```
saper, line 9: ugly and unreadable name

scripts/configure_freebsd_ci_jail.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env sh
+
+set -o xtrace
+
+jailName=$1
+skelDirectory=$2
+cbsd_workdir=/usr/jails
+jail_arch="i386"
+jail_ver="11.2"
```
saper, line 9: we are now at a 11.3

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
+freebsd_instance:
+  image: freebsd-11-2-release-amd64
```
saper, line 2: shall we use

     image_family: freebsd-11-3-snap

instead?

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
...
+  publish_script:
+    - |
+      if test "$CIRRUS_TAG" != ""; then
+        for file in `ls vendor/**/*.node`; do
+          parent=${file%/*};
+          name=${parent##*/};
+          fullyQualifiedName="$(pwd)/$parent/${name}_binding.node";
+          mv "$file" "$parent/${name}_binding.node";
+          echo -e "New filename\072 $fullyQualifiedName";
+          ./scripts/upload-github-release-asset.sh github_api_token=$GH_API_TOKEN owner=sass repo=node-sass tag=$CIRRUS_TAG filename=$fullyQualifiedName;
```
saper, line 67: How do we know it works? Should the pipeline fail if that fails? I used your key (could not be decrypted) and the publish script finished successfully.

scripts/upload-github-release-asset.sh
```diff
@@ -0,0 +1,60 @@
...
+GH_RELEASES="$GH_REPO/releases"
+GH_TAGS="$GH_RELEASES/tags/$tag"
+AUTH="Authorization: token $github_api_token"
+
+if test "$tag" == "LATEST"; then
+  GH_TAGS="$GH_REPO/releases/latest"
+fi
+
+# Validate token.
+curl -o /dev/null -sH "$AUTH" $GH_REPO || { echo "Error: Invalid repo, token or network issue!";  exit 1; }
```
saper, line 42: Assuming empty token:

    curl -o /dev/null -sH "Authorization: token " https://api.github.com/repos/sass/node-sass && echo OK

This works, so no error here

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
...
+  env:
+    GH_API_TOKEN: ENCRYPTED[5e482f417304528148bb96eca8d030eacd6ab9972d482485fc4d42907283b995f658b351e0676e9493a37d815398f541]
+
+  prepare_script:
+    - sed -i '' 's/quarterly/latest/g' /etc/pkg/FreeBSD.conf
+    - |
+      if test "$abi" = "freebsd:11:x86:32"; then
+        ./scripts/configure_freebsd_ci_jail.sh $jailName $CIRRUS_WORKING_DIR;
+        $execPrefix sed -i -- '' 's/quarterly/latest/g' /etc/pkg/FreeBSD.conf;
+        $execPrefix pkg update -f;
```
saper, line 37: Why are we doing this only for 32 bit? Maybe this is the problem with having not update repository?

.cirrus.yml
```diff
@@ -0,0 +1,69 @@
...
+    - $execPrefix node --version
+    - $execPrefix npm --version
+    - $execPrefix clang++ --version
+
+  build_script:
+    - |
+      if test "$abi" = "freebsd:11:x86:32"; then
+        changeDir="cd /etc/skel &&"
+      fi
+      echo "$changeDir npm install --unsafe-perm" | $execPrefix /bin/sh
```
saper, line 56: we should skip download here to ensure the binary is always built

nschonni:
It might be nice to get this running for v5. Right now that's just looking at a simpler matrix of "Currently Supported Node", so this matrix could be smaller and might be easier to deal with for Cirrus. The GitHub App is setup for Cirrus now, but I'm not sure if it's going to work for it coming from a fork or not

am11:
In another project, I was working with GitHub Actions and cross compilation for FreeBSD and illumos. Once the cross compilation script is setup properly (basically ~10 lines of shell script), and a docker image is built and published with that cross compilation toolchain, that container can run on any CIaaS which supports kubernetes or more generally, CaaS.
Thanks to `hub(1)`, uploading artifacts from is simple enough, e.g. this is currently working out quite nicely for my own stuff: https://github.com/am11/runtime/blob/4a95ab23d3a09262955f252e7869633989601d76/.github/workflows/main.yml#L29.

In short, we do not need FreeBSD VM to build FreeBSD binaries in the modern age of mankind. 😎

nschonni:
@am11 interesting, I didn't think FreeBSD supported Docker since it used jails instead of LXC

am11:
Yup, native Docker port is still under development for FreeBSD. However, the cross compilation  toolchain can be built for number of Unix-like OS, and it can execute on Linux. For instance, rust and dotnet runtime build their official packages with Linux.x64-based (glibc) containers that have cross compiling toolchains setup for alpine (muscle libc), arm64 etc. they also have FreeBSD CI legs using the same cross-compilation infra. I also used the same mechanics to configure a container that has cross compilation toolchain setup for illumous (derivative of SunOS), then used that container in GitHub Actions CI for dotnet runtime build (yaml link above). In order to run tests on the target platform, however, we would need a VM.

PR #2598, opened 2019-02-20
fix(cli): option --quiet suppresses output expect for error

### LINKED ISSUE #2599: --quiet should suppress log output except on error
### Current Behavior

The `--quiet` option is supposed to "Suppress log output except on error" per the documentation:
https://github.com/sass/node-sass/blame/master/README.md#L547
however, it currently is only suppressing `warn` and `info` level and lets `log` level through.

If you ran `node-sass --quiet /some/file.scss` you would expect this to only emit on error, however, this currently will output the entire resulting .css to stdout.  An example of when you would not want the resulting css but would only want an error if one occurred is if you were using the command as a CI step to verify that the scss was valid.

### Suggested Behavior
`--quiet` should suppress log output except on error

zieka:
### Background

The `--quiet` option is supposed to "Suppress log output except on error" per the documentation:
https://github.com/sass/node-sass/blame/master/README.md#L547
however, it currently is only suppressing `warn` and `info` level and lets `log` level through.

If you ran `node-sass --quiet /some/file.scss` you would expect this to only emit on error, however, this currently will output the entire resulting .css to stdout.  An example of when you would not want the resulting css but would only want an error if one occurred is if you were using the command as a CI step to verify that the scss was valid.

### Solution
This PR makes the `--quiet` option properly suppress log output except on error.

Closes #2599

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -155,19 +155,17 @@ function getEmitter() {
     }
   });
 
-  emitter.on('warn', function(data) {
-    if (!options.quiet) {
+  if (!options.quiet) {
+    emitter.on('warn', function(data) {
       console.warn(data);
-    }
-  });
+    });
 
-  emitter.on('info', function(data) {
-    if (!options.quiet) {
+    emitter.on('info', function(data) {
       console.info(data);
-    }
-  });
+    });
 
-  emitter.on('log', stdout.write.bind(stdout));
+    emitter.on('log', stdout.write.bind(stdout));
+  }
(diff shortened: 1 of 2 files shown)


xzyfer:
>If you ran `node-sass --quiet /some/file.scss` you would expect this to only emit on error, however, this currently will output the entire resulting .css to stdout

This assumption is incorrect. The default behaviour if not output flag is supplied is to output the compiled css to stdout.

PR #2608, opened 2019-03-04
https-ify sass-lang.com urls

asottile:
See https://github.com/sass/sass-site/issues/217

Committed via https://github.com/asottile/all-repos

### DIFF
--- a/src/libsass/src/operators.cpp
+++ b/src/libsass/src/operators.cpp
@@ -65,7 +65,7 @@ namespace Sass {
       );
 
       std::string msg("The operation `" + lsh + " " + op_str + " " + rhs + "` is deprecated and will be an error in future versions.");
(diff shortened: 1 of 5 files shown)


src/libsass/src/operators.cpp
```diff
@@ -65,7 +65,7 @@ namespace Sass {
       );
 
       std::string msg("The operation `" + lsh + " " + op_str + " " + rhs + "` is deprecated and will be an error in future versions.");
-      std::string tail("Consider using Sass's color functions instead.\nhttp://sass-lang.com/documentation/Sass/Script/Functions.html#other_color_functions");
+      std::string tail("Consider using Sass's color functions instead.\nhttps://sass-lang.com/documentation/Sass/Script/Functions.html#other_color_functions");
```
nschonni: Please undo this one since it is part of the libsass code, and will get updated when the subtree is updated

PR #2613, opened 2019-03-13
Fix: alternate build targets

TomSputz:
The old build scripts didn't use the given arguments to discover the binary's path, among other fixes.


TomSputz:
There seems to be an issue causing `arch` to be undefined when `npm explore node-sass -- npm run build` is run, causing the binary to be written to the wrong folder. I can't begin to figure out why it's happening

Issue #2625, opened 2019-04-11
[Security] Vulnerability in `tar`


asbjornh:
# Do not open a PR. We appreciate the enthusiasm but the fix is more complicated than it appears. We're considering our options.

See https://www.npmjs.com/advisories/803

> Versions of node-tar prior to 4.4.2 are vulnerable to Arbitrary File Overwrite. Extracting tarballs containing a hardlink to a file that already exists in the system, and a file that matches the hardlink will overwrite the system's file with the contents of the extracted file.

Caused by `node-gyp`. I guess this depends on https://github.com/nodejs/node-gyp/issues/1714 being fixed first. As far as I can tell, to fix this `node-sass` needs to to upgrade to `node-gyp@4.x.x` once they've resolved the issue on their part.

Output from `yarn audit`:

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Arbitrary File Overwrite                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ tar                                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=4.4.2                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ node-sass                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ node-sass > node-gyp > tar                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/803                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
1 vulnerabilities found - Packages audited: 16503
Severity: 1 High
```

loblawdevforsale:
Dealing with the same issue. Tried `npm update node-sass --depth 999`, `npm i tar --save`, and `npm update tar --depth999` none of that helped updating `tar` for node-sass

mohsenari:
Dealing with the same issue. Tried npm update node-sass --depth 999, npm i tar --save, and npm update tar --depth999 none of that helped updating tar for node-sass

mohsenari:
Anyone who's looking for a temporary workaround until this gets fixed, I managed to update `tar` version using **izogfif**'s answer here: https://stackoverflow.com/questions/15806152/how-do-i-override-nested-npm-dependency-versions

You need to remove `tar` form required section in node-gyp in package-lock.json
then replace this in the `dependencies` section in the same place:
```
"dependencies": {
[...]

PR #2629, opened 2019-04-20
Update node-gyp to eliminate tar vulnerability

aayush420:
node-gyp depends on tar. High security vulnerability was discovered in tar : https://hackerone.com/reports/344595. This is fixed in tar 4.4.2.  node-gyp 4.0.0 uses tar 4.4.8, thus eliminating the vulnerability.


nschonni:
This breaks compatibilty with older Node versions. We'll wait for the resolution of https://github.com/nodejs/node-gyp/pull/1718
As mentioned https://github.com/sass/node-sass/issues/2625#issuecomment-483085986 node-sass shouldn't be affected by the vulnerablity anyway

uniibu:
besides, there's no `v4.0.0` of node-gyp yet (as of the time of this post, see image below) So I'm not sure where did `v4.0.0` came from? Anyhow, the issue is not about node-sass is affected or not, but its about CI failing due to a failed npm/yarn audit. 
node-gyp

PR #2253, opened 2018-02-19
Add ARM CPU architectures

jhermsmeier:
This adds armv7 and armv8 cpu architectures to `getHumanArchitecture()`
in order to support running node-sass on ARM.

PR #2630, opened 2019-04-22
Node-sass:Providing the support for arm64.

ossdev07:
As per the discussion in PR #2253, provided the support for arm64.

### DIFF
--- a/lib/extensions.js
+++ b/lib/extensions.js
@@ -44,6 +44,11 @@ function getHumanArchitecture(arch) {
     case 'ia32': return '32-bit';
     case 'x86': return '32-bit';
     case 'x64': return '64-bit';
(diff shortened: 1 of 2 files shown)


xzyfer:
Please take time to read the linked issue for why this change alone is not sufficient.

ossdev07:
Hi @xzyfer 

I agree with you, these changes are not sufficient.

As per the conversation https://github.com/sass/node-sass/pull/2253#issuecomment-382236477
If someone provides the CI/CD ( Drone.io) support, then the binary could be released for the arm64.

I have successfully able to create a drone.yml, with which the package can run on the arm64 platform.
If it could solve the problem I will be pleased to provide one.

What is your suggestion for it?

PR #2635, opened 2019-04-24
Vulnerability Patch – node-gyp

Manbearpixel:
A recently published vulnerability in the `tar` package, of which is depended on by `node-gyp` and is then depended on by `node-sass` was identified.

The `tar` package has been resolved in versions `>= 4.4.2` and `node-gyp` was recently updated to `4.0.0` which includes a `tar` dependency of `4.4.8` which should resolve this chain.

---
- [`node-gyp` package source](https://github.com/nodejs/node-gyp/blob/master/package.json)
- [Vulnerability report](https://www.npmjs.com/advisories/803)


xzyfer:
Duplicate https://github.com/sass/node-sass/issues/2625.

See the latest update in the pinned issue. This breaks BC.

Manbearpixel:
@xzyfer I see your comments in #2625 but they literally just release a new version with an updated `tar` dependency an hour ago: https://www.npmjs.com/package/node-gyp

Could `node-sass` not just use `4.0.0` or will that break something in `node-sass`?

xzyfer:
It would require a major version bump since we'd need to drop support for older node versions. Although this is something we are planning to do in #2312 we aren't quite ready. We'd prefer not to do multiple major version bumps in a short period because it's very disruptive to the community.

We're keeping an eye on the upcoming node TSC meeting to seeing a non breaking node-gyp release is likely to happen.

PR #2640, opened 2019-04-24
Update Pkg To Latest Node-Gyp

arthurbernierjr:
Node Gyp depends on TAR and Tar version prior to 4.4.2 have a security vulnerability.


xzyfer:
See the pinned issue.

PR #2659, opened 2019-05-01
fix(deps): upgrading node-gyp to fix vulnerability

sketchidea:
this fixes the security vulnerability of the node-gyp dependancy "tar", that was present in the previous version of node-gyp. NPM audit referred to the vulnerability as being a High severity vulnerability and the risk being caused by "Arbitrary File Overwrite".

I have tested this upgrade locally in my own application that uses node-sass, and from manual testing this works well and doesn't cause any visual issues, nor are there any errors in the console.


nschonni:
Duplicate of a bunch of PRs. Follow #2625

PR #2675, opened 2019-05-23
Update libsass version to 3.6

PolyPik:


xzyfer:
Sorry it's much more complicated than this.

PolyPik:
Very well then I'll leave the job up to the rest of you.

PR #2681, opened 2019-05-31
bump node-gyp@4.0.0

protonate:
Current version of `node-gyp@3.8.0` uses `fstream` which is reporting a security vulnerability. `node-gyp@4.0.0` removes the `fstream` dependency.


xzyfer:
Please search before creating new issues.

npm update fstream

protonate:
sorry its a little more complicated then that

`fstream` is no longer used with `node-gyp@4.0.0` so `npm update fstream` is pointless and leaves a vuln in your project.

xzyfer:
We can't use node-gyp@4 because it because old node support. When fstream is patches It'll be available via npm update.

nschonni:
> When fstream is patches It'll be available via npm update.

It's already available, the reason why someone might be seeing this is if they have a lockfile with the old version and haven't updated the dependency as you mentioned

PR #2688, opened 2019-06-07
bump version of node-gyp up to 4.0.0

steevhise:
So that we're not using insecure version of `tar` anymore.  (see https://nvd.nist.gov/vuln/detail/CVE-2018-20834 )


xzyfer:
Please search closed issues before opening new ones. There is a dozen identical closed issue.

steevhise:
My apologies, I did do a brief search and found nothing, but maybe I didn't look for the right thing...

steevhise:
So, the fact that so many people did not and still have not seen what you're doing about this problem would seem to indicate that another way of communicating it is in order.  Maybe a brief note in the readme? Maybe leaving an issue open? Maybe...?

nschonni:
@steevhise we did have the pinned issue and I had it in the PR template (but removed it after the upstream issue was resolved), but neither seemed to discourage people from opening PRs. Do you want to suggest a PR to the Troubleshooting.md maybe?

steevhise:
oh I wasn't aware of the resolution. Github still gave me an alert on one of my repos. But now I see that after I re-installed I get node-gyp 3.8.0 which uses a safe version of tar.  sorry! thanks.

steevhise:
I think the issue is really not yours but with the security audit systems (for both npm and github). there should be more information about troubleshooting coming from them, rather than just an alert and an "autofix" button (which I don't trust and i assume a lot of others don't either).

nschonni:
no problem, thanks for the feedback. Maybe we'll keep the issue pinned for longer after the resolution. Not sure what a reasonable timeframe is, but we can keep experimenting

PR #2690, opened 2019-06-08
deps(chalk): Update to version 2.4.2

tomciopp:
Updates chalk to newest version. There shouldn't be any code changes needed for this update.


nschonni:
Looks like this breaks on old node, so we can't do this till v5

PR #2692, opened 2019-06-10
deps: update nan, node-gyp

jbergstroem:
This fixes build issues with node 12.x and NaN, most namely `IsNearDeath`.

Newer versions of node (v8) removed it after its deprecation period.
Upstream discussion/removal here: https://github.com/nodejs/nan/pull/842

An example build traceback would look something like:
```
In file included from ../src/binding.cpp:1:
In file included from ../../nan/nan.h:2657:
../../nan/nan_object_wrap.h:24:25: error: no member named 'IsNearDeath' in 'Nan::Persistent<v8::Object,
      v8::NonCopyablePersistentTraits<v8::Object> >'
    assert(persistent().IsNearDeath());
           ~~~~~~~~~~~~ ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include/assert.h:93:25: note: 
      expanded from macro 'assert'
    (__builtin_expect(!(e), 0) ? __assert_rtn(__func__, __FILE__, __LINE__, #e) : (void)0)
```


nschonni:
Thanks, but the nan version is still in the `^` range already. We can't bump node-gyp because it breaks old Node (bunch of closed PRs for this already)

jbergstroem:
So, `0.x` builds seems to fail; not sure why those are even in the matrix since even gyp 3.x (currently leaned on) is [node 4 or newer](https://github.com/nodejs/node-gyp/blob/v3.x/package.json#L38).

jbergstroem:
> We can't bump node-gyp because it breaks old Node (bunch of closed PRs for this already)

Posted while I was writing. Thanks for that. I'll look at what dependency chain that breaks for me then.

PR #2708, opened 2019-07-11
Bump sass-graph@^3.0.0

### LINKED ISSUE #2614: [Security] Vulnerability in lodash below 4.17.5 used by sass-graph@2
https://www.npmjs.com/advisories/782

> Versions of lodash before 4.17.5 are vulnerable to prototype pollution.
>
> The vulnerable functions are 'defaultsDeep', 'merge', and 'mergeWith' which allow a malicious user to modify the prototype of Object via {constructor: {prototype: {...}}} causing the addition or modification of an existing property that will exist on all objects.
>
> Remediation
> Update to version 4.17.11 or later.

Resolution: update to sass-graph@latest. The currently used version of sass-graph is the last at v2, so some testing will be needed to upgrade to v3

I'm happy to investigate a PR, though would be helpful to understand the usage of sass-graph before I get too deep. Is there anything that has prevented the upgrade in the past?

eatonphil:
Sass-graph is pulling in an old lodash version that has security alerts out for it. This bumps sass-graph to 3.0.0. [The only breaking change is that it doesn't read .css files by default and it requires Node >= 6.](https://github.com/xzyfer/sass-graph/releases/tag/v3.0.0)

Closes #2614

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "node-gyp": "^3.8.0",
     "npmlog": "^4.0.0",
     "request": "^2.88.0",
-    "sass-graph": "^2.2.4",
+    "sass-graph": "^3.0.0",
     "stdout-stream": "^1.4.0",
     "true-case-path": "^1.0.2"
   },


xzyfer:
We can't break BC with old node. The solution is to release a version of
sass-graph@2 with a patched lodash.

Issue #2709, opened 2019-07-18
test directory included in package


omrilotan:
https://github.com/sass/node-sass/blob/master/package.json#L44
<img src="https://user-images.githubusercontent.com/516342/61464219-d0231400-a96d-11e9-9abe-e0fa494946e3.png" width="200">

It appears that this directory includes only test files. I think it should be excluded from NPM package.
This way test frameworks will also not try to parse it when running without excluding node_modules explicitly (which fails because of missing devDependencies, of course).

| item | value
| - | -
| NPM version | 6.4.1
| Node version | v10.15.3
| Node Process | `{ http_parser: '2.8.0', node: '10.15.3', v8: '6.8.275.32-node.51', uv: '1.23.2', zlib: '1.2.11', ares: '1.15.0', modules: '64', nghttp2: '1.34.0', napi: '3', openssl: '1.1.0j', icu: '62.1', unicode: '11.0', cldr: '33.1', tz: '2018e' }`
| Node Platform | darwin
| Node architecture | x64
| node-sass version | node-sass	4.12.0 (Wrapper) [JavaScript], libsass  	3.5.4 (Sass Compiler) [C/C++]
| npm node-sass versions | `└── node-sass@4.12.0`

xzyfer:
This is by design for necessary 3rd party integration systems.

PR #2710, opened 2019-07-18
Remove test from published package. Address #2709

omrilotan:
#2709 content:

> https://github.com/sass/node-sass/blob/master/package.json#L44
> 
>
> It appears that this directory includes only test files. I think it should be excluded from NPM package.
> This way test frameworks will also not try to parse it when running without excluding node_modules explicitly (which fails because of missing devDependencies, of course).

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "lib",
     "scripts",
     "src",
-    "test",
     "vendor"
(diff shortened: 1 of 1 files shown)


xzyfer:
This is by design.

PR #2714, opened 2019-07-24
Libsass 2b8a17a

ItsLeeOwen:
Updating to latest libsass 

node-sass mocha tests don't pass, I'm not sure how to reconcile node-sass tests against libsass changes between Nov 11, 2018 and current libsass#2b8a17a. Unless I'm misunderstanding the failing node-sass tests I'm not confident that node-sass should have it's own set of render output tests, it should probably depend on libsass tests if it's not.

  5935 passing (18s)
  753 pending
  143 failing


package.json
```diff
@@ -1,87 +1,87 @@
...
+		"coveralls": "^3.0.2",
+		"eslint": "^3.4.0",
+		"fs-extra": "^0.30.0",
+		"istanbul": "^0.4.2",
+		"mocha": "^3.1.2",
+		"mocha-lcov-reporter": "^1.2.0",
+		"object-merge": "^2.5.1",
+		"read-yaml": "^1.0.0",
+		"rimraf": "^2.5.2",
+		"sass-spec": "https://github.com/sass/sass-spec.git#dc2d573",
```
saper: There seems to be a bit too much whitespace change, cna you check?

You need to figure out the right commit from the https://github.com/sass/sass-spec repository, it will be probably close to the master branch.

PR #2718, opened 2019-08-12
Fix handling of build flag and empty env

kapouer:
`./scripts/build.js --libsass_ext=auto` resets `libsass_*` whereas it should not.
This patch drops the env vars that have not been set at all.
Incidentally it fixes the build when using shared libsass.


scripts/build.js
```diff
@@ -56,10 +56,14 @@ function afterBuild(options) {
...
 
 function build(options) {
-  var args = [require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')), 'rebuild', '--verbose'].concat(
-    ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].map(function(subject) {
-      return ['--', subject, '=', process.env[subject.toUpperCase()] || ''].join('');
-    })).concat(options.args);
+  var libargs = [];
+  ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].forEach(function(subject) {
+    var env = process.env[subject.toUpperCase()];
+    if (env != null) libargs.push(['--', subject, '=', env].join(''));
```
xzyfer:
```diff
-if (env != null)
+if (env)
```

scripts/build.js
```diff
@@ -56,10 +56,14 @@ function afterBuild(options) {
  */
 
 function build(options) {
-  var args = [require.resolve(path.join('node-gyp', 'bin', 'node-gyp.js')), 'rebuild', '--verbose'].concat(
-    ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].map(function(subject) {
-      return ['--', subject, '=', process.env[subject.toUpperCase()] || ''].join('');
-    })).concat(options.args);
+  var libargs = [];
```
xzyfer: Please use reduce

```js
var libargs = ['libsass_ext', 'libsass_cflags', 'libsass_ldflags', 'libsass_library'].reduce(function(prev, cur) {
  var env = process.env[subject.toUpperCase()];
  if (env) {
    return prev.push(['--', subject, '=', env].join(''));
  } else {
    return prev;
  }
});
```

PR #2748, opened 2019-10-08
Updated mocha by 3 major versions, fixing audit vulnerabilities

Antrikshy:
I noticed the package had two vulnerabilities listed under `npm audit`, one of them critical. Not that big of a deal because they were both in dev-dependencies, through mocha. But it was (seemingly) completely straightforward to bump up mocha to the latest version. The console output has a slight visual refresh, but is largely the same as before. I went ahead and updated the coverage reporter to its latest version too, but there isn't much to it, looking at the changelog.

Tests still pass for me, and now we have zero scary vulnerabilities.


nschonni:
Sorry, there were previous PRs for this, but it doesn't support older node versions that we won't be dropping till the next major release

Antrikshy:
Ah, I see. No worries!

PR #2754, opened 2019-10-17
Fix #2394: sourceMap option should have consistent behaviour

### LINKED ISSUE #2394: sourceMap option documentation is inaccurate
The documentation for the `sourceMap` option claims that

> Setting the `sourceMap` option requires also setting the `outFile` option

However, in practice, if you set `sourceMap` to a string and don't set `outFile`, you still get a `map` buffer in the result. For example:

```js
var sass = require('node-sass');

var result = sass.renderSync({
  data: 'a {b: c}',
  sourceMap: 'out.css.map'
});
console.log(result.map.toString());
```

prints

```json
{
        "version": 3,
        "file": "stdin.css",
        "sources": [
                "stdin"
        ],
        "names": [],
        "mappings": "AAAA,AAAA,CAAC,CAAC;EAAC,CAAC,EAAE,CAAC,GAAE"
}
```

saper:
render() and renderSync() should return "map" property in the results
only if source map has been enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,9 +310,11 @@ Used to determine how many digits after the decimal will be allowed. For instanc
 * Type: `Boolean | String | undefined`
 * Default: `undefined`
 
-**Special:** Setting the `sourceMap` option requires also setting the `outFile` option
+Enables source map generation during `render` and `renderSync`.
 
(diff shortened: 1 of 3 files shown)


saper:
Related https://github.com/sass/node-sass/pull/2501

xzyfer:
This is potentially a brrakky change but I'm happy to move forward since a it's how the docs always said it worked

saper:
Thanks. I hope to get documentation to the level where it can actually be used as the requirements document :)

PR #2769, opened 2019-10-23
Fix #2621: Report libsass version 3.5.5

### LINKED ISSUE #2621: v4.11.0 and v4.12.0 binaries say using libsass 3.5.4 instead of 3.5.5
What the title says... at least for me, after doing a clean install of node-sass:  

```
> npx node-sass --version
node-sass       4.11.0  (Wrapper)       [JavaScript]
libsass         3.5.4   (Sass Compiler) [C/C++]
> npm --version 
6.4.1
> node --version
v11.2.0
```

Please release a new version with 3.5.5 (or later) due to security vulnerabilities. 

And at a minimum the documentation should be updated to state the version it actually uses. 

If needed I can try to submit a PR. 

(Windows 10 Enterprise, vr 10.0.16299, 64bit)

saper:
@xzyfer can we push 4.13.0 to include this? unfortunately the binaries have to be re-built

Fixes #2621

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "node-sass",
   "version": "4.13.0",
-  "libsass": "3.5.4",
+  "libsass": "3.5.5",
   "description": "Wrapper around libsass",
   "license": "MIT",
   "bugs": "https://github.com/sass/node-sass/issues",


saper:
I have my doubts about stability of 3.6 given massive changes imported from Dart, but maybe we should jump that ship. Your call :)

xzyfer:
3.6.1 is relatively safe and lands some features people have been asking
for like _index.scss

srcereress:
We're trying to get a newer version of libsass because Whitesource is complaining about issues below 3.5.5. We don't use this directly but our project requires it. Any eta when this might get approved? Thanks for the great work!

nschonni:
@srcereress that sounds like a bug with Whitesource then, because this isn't changing the libsass version, just what the CLI prints out

saper:
> @srcereress that sounds like a bug with Whitesource then, because this isn't changing the libsass version, just what the CLI prints out

I don't they have any way to do better - `libsass_version()` in C interface will also wrongly return `3.5.5`.

srcereress:
@nschonni and @saper - thanks. There's actually another open tix here about upgrading libsass 3.6.0. I wasn't sure if the version reporting would cause Whitesource to trigger an issue. If this reporting got fixed, maybe 1 of 2 of the issues would go away. It was a shot in the dark quite frankly. :-)

saper:
@srcereress "do as we do and not as we say" :)

franklsm1:
Is this on track to be included into a release soon? Whitesource is still complaining about this due to the way they look at versions, and even though it is not actually using that version it is rather annoying to have to request approvals to bypass the Whitesource false positive each time we use node-sass in a new app.

nschonni:
@xzyfer it might be worth landing this and releaseing a 4.13.2 just because of the binding cache issues with alpine people might already have on 4.13.1

xzyfer:
I had it in my head that this required rebuilding the binaries so I've been putting it off. My bad.

xzyfer:
Oh never mind the binaries do have to elbe rebuilt. I jump on plane for vacation in 24hrs. I'll kick off a build shortly

nschonni:
Don't worry about it, I think it could wait till after your break

PR #2787, opened 2019-11-17
Fix render example syntax

ZoranPandovski:
Remove Unexpected token ) on the render example.


nschonni:
@ZoranPandovski Thanks!

Issue #2822, opened 2020-01-16
"Binary has a problem" with 4.13.1


ckaczor:
We're getting the following during builds with 4.13.1 - if we go back to 4.13.0 everything is fine.

2020-01-16T17:04:44.3993932Z > node-sass@4.13.1 postinstall /node_modules/node-sass
2020-01-16T17:04:44.4010867Z > node scripts/build.js
2020-01-16T17:04:44.4016913Z 
2020-01-16T17:04:44.5530663Z Binary found at /node_modules/node-sass/vendor/linux_musl-x64-64/binding.node
2020-01-16T17:04:44.5548706Z Testing binary
2020-01-16T17:04:44.6559773Z Binary has a problem: Error: Error relocating /node_modules/node-sass/vendor/linux_musl-x64-64/binding.node: _ZNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEC1Ev: symbol not found
2020-01-16T17:04:44.6609834Z     at Object.Module._extensions..node (internal/modules/cjs/loader.js:718:18)
2020-01-16T17:04:44.6623255Z     at Module.load (internal/modules/cjs/loader.js:599:32)
2020-01-16T17:04:44.6640744Z     at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
2020-01-16T17:04:44.6653951Z     at Function.Module._load (internal/modules/cjs/loader.js:530:3)
2020-01-16T17:04:44.6667669Z     at Module.require (internal/modules/cjs/loader.js:637:17)
2020-01-16T17:04:44.6681340Z     at require (internal/modules/cjs/helpers.js:22:18)
2020-01-16T17:04:44.6696639Z     at module.exports (/node_modules/node-sass/lib/binding.js:19:10)
2020-01-16T17:04:44.6713582Z     at Object.<anonymous> (/node_modules/node-sass/lib/index.js:14:35)
2020-01-16T17:04:44.6730091Z     at Module._compile (internal/modules/cjs/loader.js:689:30)
2020-01-16T17:04:44.6743279Z     at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)

nschonni:
Did you clear your node_modules or vendor folder? The binding code was changed so it won't be able to reusing the binding.node from 4.13.0

rwoody:
We got this in a docker build as well. We aren't using any lock files. Changed the package.json from `^4.9.3` to `4.13.0` and everything works again.

[...]

PR #2823, opened 2020-01-18
chore: Add GitHub Actions for Alpine CI

nschonni:
Adds some CI for the Alpine images to speed up releases.
If this works out, I might take a look at converting some of the others.
This also uploads the artifacts as a zip on successful builds. This could be modified to only run on tags later

Note: ran into an issue with the Node 8.17.0 alpine images. You can see an example log here https://gist.github.com/nschonni/491cd2ae724e1f24db3b76b37698ab1c
Main issues is probably around
```
2020-01-18T03:50:16.3975372Z npm ERR! Command failed: git clone --mirror -q https://github.com/sass/sass-spec.git /github/home/.npm/_cacache/tmp/git-clone-dbd532b1/.git
2020-01-18T03:50:16.3975793Z npm ERR! /github/home/.npm/_cacache/tmp/git-clone-dbd532b1/.git: Permission denied
```
where the git sass-spec is failing. Tried a bunch of troubleshooting, but just decided to pin to the previous version that was working.

PS: I don't think the CI will show up till this lands, but you can see the build here https://github.com/nschonni/node-sass/commit/c288e6067650d9ebd11980c6e320e8b409f422d8/checks?check_suite_id=407548884

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -1,0 +1,49 @@
+name: Build bindings for Alpine releases
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: node:${{ matrix.node }}-alpine${{ matrix.alpine }}
+      env:
+        SKIP_SASS_BINARY_DOWNLOAD_FOR_CI: true
+    strategy:
+      matrix:
+        node:
+          - 6
+          - 8.16.2
+          - 10
+          - 12
+          - 13
+          - 14
+        include:
+          - node: 6
+            alpine: ""
+          - node: 8.16.2
+            alpine: "3.9"
+          - node: 10
+            alpine: "3.9"
+          - node: 12
+            alpine: "3.9"
+          - node: 13
+            alpine: "3.10"
+          - node: 14
+            alpine: "3.10"
+    steps:
+      - name: Install Alpine build tools
+        run: apk add --no-cache python make git gcc g++
+
+      - uses: actions/checkout@v2
+
+      - name: Install packages
+        run: npm install --unsafe-perm
+
+      - name: Run tests
+        run: npm test
+
+      - uses: actions/upload-artifact@v1
+        with:
+          name: ${{ matrix.node }}
+          path: vendor/


xzyfer:
This looks like something would do easily enough with Travis CI but I'm curious to see how this compares.

.github/workflows/alpine.yml
```diff
@@ -0,0 +1,33 @@
...
+      volumes:
+        - $GITHUB_WORKSPACE/vendor:/__w/node-sass/node-sass/vendor
+      env:
+        SKIP_SASS_BINARY_DOWNLOAD_FOR_CI: true
+    strategy:
+      matrix:
+        node:
+          - 6
+          - 8.16.2 # Something in 8.17.0 is breaking NPM cache install
+          - 10
```
xzyfer: Give https://github.com/sass/node-sass/issues/2822 we'll want to use 10.16

nschonni: I didn't run into this one in the testing, can you give an example of what Alpine version you hit the issue with?

xzyfer: In the end the gist of the problem appears to be that binaries built on newer alpine version didn't work on older alpine versions. I think most alpine docker images come 3.9, 3.10, 3.11, so we'll probably want to build binaries in whatever the lowest is for the specific node version.

nschonni: OK, there are tags that include the Alpine version, so I can take a look at aligning that details to the run

nschonni: I've pinned them to the oldest versions now

nschonni:
Added v14, but it seems like GitHub is using an image cache that is missing it right now https://github.com/nschonni/node-sass/runs/607236373

xzyfer:
I'm having the same issue trying to build the binary locally. Might be dockerhub struggling after today's EC2 issues.

xzyfer:
I've got the Linux and pax binaries built locally. I'll cut a minor release tonight.

nschonni:
Looks like it's happy now https://github.com/nschonni/node-sass/actions/runs/85070002

xzyfer:
Feel free to ship this when you're ready @nschonni

xzyfer:
This is super cool btw.

nschonni:
Lets give it a shot. If it works out, we can cherrypick something for v5. It's possible that Actions could replace most of the CI once we don't need the old VS & Node versions there

Issue #2716, opened 2019-07-29
node-sass using node-gyp v3.8.0 when v5.0.3 is out is there a reason?


jasonwilliams:
as title.

node-sass seems to be using an old version of node-gyp (which then pulls in old dependencies), but looks like node-gyp is now on v5.0.3.

Can node-sass bump to this version?

https://github.com/sass/node-sass/blob/master/package.json#L67

xzyfer:
4.0.0 breaks support for old node.

jeffrson:
Well, newer node-gyp can use Python3 - which is important when Python2 will be end-of-life at the end of 2019. So you definitely should move on.

omarlara:
@xzyfer any plan to upgrade node-gyp 6.0.0?  Python2 will be losing support at the end of 2019.

jkim7:
Yes I'd also like this upgraded, it introduces a file cb-never-called-1.0.1.tgz which throws a malware warning on my antiviruse

mitchellmebane:
@xzyfer node-gyp@6.0.1 requires Node >= 6.0.0.  Node 4.x has been EOL since April 2018, and Node 5.x has been EOL since June 2018.  If you're concerned about strictly following semver, could you please consider cutting a 5.x release with just a node-gyp bump?  The Python 2 dependency is going to start being a very big pain point for many users in the coming months.

LC43:
can this be reconsidered?

there's two vulnerabilities related to old version. Latest node-gyp@6.0.1 uses tar@4.4.12 which clears :
- https://www.npmjs.com/advisories/803
- https://www.npmjs.com/advisories/886

qnighy:
What is blocking this from happening other than compatibility things, which can be addressed by what @mitchellmebane said?

boggsey:
Python 2 is now EOL and the latest versions of node-gyp now support Python 3. Would love to see node-sass bump this up.

kpcyrd:
Any news here? We're going to drop python2 from Arch Linux in the foreseeable future.

LC43:
@xzyfer , should we open a new ticket to review this update?

KingOfCramers:
I believe the old node-gyp version is breaking installs, for example I've run a clean docker install and this is causing failures, see [this](https://github.com/sass/node-sass/issues/2822) thread

xzyfer:
5.0 has been released with latest node-gyp

saper:
[...]

PR #2841, opened 2020-02-03
Drop EOL python2 dependency

kpcyrd:
This bumps the node-gyp dependency to a version that supports python3. Please consider releasing this asap since python2 is already end-of-life.


Geo1088:
Ref: #2716

nschonni:
This is bumped already in the v5 branch, it can't be updated in master because it drops support for old NodeJS versions

kpcyrd:
What's the ETA for v5? It seems it's being worked on since 2017 and had very little activity during 2019.

PR #2843, opened 2020-02-05
Bump node-gyp: match node lts built-in or greater

imatlopez:
node-gyp 5 supports node 6, so this will be semver major


xzyfer:
This is being done in v5. Please check recent closed PRs.

PR #2862, opened 2020-03-16
Support up to Electron 8

glenn2223:
Bring support for Electron 5-8

Needs contributors to build to full support, currently only Windows x32 & x64, OSX x64 and Linux x64


nschonni:
@glenn2223 thanks for kicking this off!
Few comments:
- Can you try adding some CI for this? I'd suggest taking a look at https://github.com/sass/node-sass/pull/2823 as an example
- I think just updating the `case` statements with the module should be fine

glenn2223:
As you suggested, just updated the switch.

New to CI, am I taking a step in the right direction here? I'll read more on the GitHub docs over the coming days. Or should I be doing something else (maybe even altering the appveyor?)

```yaml
name: Electron Rebuild CI

on: [create]

jobs:
  build:

    runs-on: ${{ matrix.os }}

    strategy:
      matrix:
        node: ['13']
        os: [windows-latest, ubuntu-latest, macOS-latest]

    steps:
    - name: ${{ matrix.os }} - Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        os: ${{ matrix.os }}
    - run: npm ci
    - run: npm run build --if-present
    - run: npm install --save-dev electron-rebuild
    - name: Run on ${{ matrix.os }}
      if: matrix.os != 'macOS-latest'
        run:|
          ./node_modules/.bin/electron-rebuild -v=8 -w node-sass --arch=ia32
          ./node_modules/.bin/electron-rebuild -v=7 -w node-sass --arch=ia32
          ./node_modules/.bin/electron-rebuild -v=6 -w node-sass --arch=ia32
          ./node_modules/.bin/electron-rebuild -v=5 -w node-sass --arch=ia32
          ./node_modules/.bin/electron-rebuild -v=8 -w node-sass --arch=x64
          ./node_modules/.bin/electron-rebuild -v=7 -w node-sass --arch=x64
          ./node_modules/.bin/electron-rebuild -v=6 -w node-sass --arch=x64
          ./node_modules/.bin/electron-rebuild -v=5 -w node-sass --arch=x64
    - name: Run on ${{ matrix.os }}
      if: matrix.os == 'macOS-latest'
        run:|
          ./node_modules/.bin/electron-rebuild -v=8 -w node-sass
          ./node_modules/.bin/electron-rebuild -v=7 -w node-sass
          ./node_modules/.bin/electron-rebuild -v=6 -w node-sass
          ./node_modules/.bin/electron-rebuild -v=5 -w node-sass
    - run: npm test
      env:
        CI: true
```

nschonni:
Looks like a good start! There are a few things I'd probably suggest changing, but if you start by adding the file to the PR, it will be easier to comment directly on the file.

Here are a few other suggestions:
- Create separate YML files per environment since it seems like there might be some different handling. EX: electron-osx.yml, electron-win.yml, electron-linux.yml
- The binding.node files should be removed from the PR
- Not sure what the best place is for the instructions, maybe just the root of the repo for now, as the `.github` is for special GitHub related files
- The `on` should probably be `[push, pull_request]` instead of `create`
- Add the actions/checkout@v2 as the first `step`
- Use the electron version as part of your matrix as an array. The `arch` looks like it might only need to be in the Windows build. EX: 
````yml
    strategy:
      matrix:
        electron:
        - 5
        - 6
        - 7
        - 8
        arch:
        - ia32
        - x64
....
    - name: Run on ${{ matrix.os }}
      if: matrix.os != 'macOS-latest'
        run:|
          ./node_modules/.bin/electron-rebuild -v=${{ matrix.electron }} -w node-sass --arch=${{ matrix.arch}}
```
- You can comment out the individual electron version lines to shrink the amount of builds kicked off while you're iterating
- You can take a look at the Alpine PR for an idea of how caching/uploading the binaries might work, but I'd leave that till the end

Hope this helps you iterate one this

glenn2223:
@nschonni Thanks for all your help, I really appreciate it. One question in comments, if you can take a look

Built it for windows and getting that to work, then I'll duplicate for other OS's

Stumbled onto an alternate method, just testing the outputted file now

.github/workflows/electron-win.yml
```diff
@@ -0,0 +1,43 @@
...
+        os: ${{ matrix.os }}
+    - name: Node Ver (For debugging)
+      run: node -v
+    - name: Install packages
+      run: npm install --unsafe-perm
+    - name: Run on ${{ matrix.os }} - Electron v${{ matrix.electron }}
+      run: node ".\node_modules\node-gyp\bin\node-gyp.js" rebuild --target=${{ matrix.electron }} --arch=${{ matrix.arch }} --dist-url=https://electronjs.org/headers
+    - uses: actions/upload-artifact@v1
+      with:
+        name: win32_${{ matrix.arch }}_${{ matrix.electron }}.node
```
glenn2223: How can I replace `${{ matrix.electron }}` with the node version commented above (line 13-16)

glenn2223:
Looking at docs, it seems we can target the below straight off the bat
  - Windows x32 (ia32)
  - Windows x64
  - OSX (darwin x64)
  - Linux ia32
  - Linux x64

But how do we target FreeBSD and Linux/musl? Will it have to be self-hosted runners?

package.json
```diff
@@ -64,7 +64,7 @@
     "meow": "^3.7.0",
     "mkdirp": "^0.5.1",
     "nan": "^2.13.2",
-    "node-gyp": "^3.8.0",
+    "node-gyp": "^6.1.0",
```
nschonni: This can't be updated till v5 because it breaks support for old Node versions. If you need this as part of this setup, please edit the PR to target the v5 branch where it should already be bumped

glenn2223: Is there a planned release date for v5?

.github/workflows/electron-linux.yml
```diff
@@ -0,0 +1,53 @@
...
+          - electron: 5
+            nodeVersion: 70
+          - electron: 6
+            nodeVersion: 73
+          - electron: 7
+            nodeVersion: 75
+          - electron: 8
+            nodeVersion: 76
+        arch:
+          - ia32
```
nschonni: I'd probably drop this, since official x86 Node support dropped a couple of versions back

.github/workflows/electron-linux.yml
```diff
@@ -0,0 +1,53 @@
...
+        os: ${{ matrix.os }}
+    - name: Node Ver (For debugging)
+      run: node -v
+    - name: Install packages
+      run: npm install --unsafe-perm
+    - if: matrix.arch == 'ia32'
+      name: Run extra command on ia32
+      run: sudo apt-get install gcc-multilib g++-multilib
+    - name: Run on ${{ matrix.os }} - Electron v${{ matrix.electron }} (Node ${{ matrix.nodeVersion }})
+      run: node "./node_modules/node-gyp/bin/node-gyp.js" rebuild --target=${{ matrix.electron }}.0.0 --arch=${{ matrix.arch }} --dist-url=https://electronjs.org/headers
```
nschonni: Not sure if the `electron-rebuild` didn't work out, but you could also try running it through NPX like
```
npx electron-rebuild -v=${{ matrix.electron }} -w node-sass
```
That skips the need for the global install

.github/workflows/electron-linux.yml
glenn2223: Just THANK YOU!!! Having real problems with linux

.github/workflows/electron-linux.yml
glenn2223: Switched back and now can be back into the master!!! 🎉

package.json
glenn2223: Un-bumped

glenn2223:
@nschonni What do you think? Are there any other steps I need to take?

https://github.com/glenn2223/node-sass/actions?query=is%3Asuccess

.github/workflows/electron-linux.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: npm install --unsafe-perm
+    - name: Install electron-rebuild
+      run: npm install --save-dev electron-rebuild@latest
+    - name: Run on ${{ matrix.os }} - Electron v${{ matrix.electron }} (Node ${{ matrix.nodeVersion }})
+      run: ./node_modules/.bin/electron-rebuild -v=${{ matrix.electron }} -w node-sass
+    - uses: actions/upload-artifact@v1
+      with:
+        name: linux_x64_${{ matrix.nodeVersion }}_binding.node
+        path: build/Release/binding.node
+    # npm test failing on pre-existing master issues, commenting out
```
nschonni: What the pre-existing failures?

nschonni:
> But how do we target FreeBSD and Linux/musl? Will it have to be self-hosted runners?

I think those can be skipped. Alpine is a container only OS, unless you think people might be running something through Electron that way, it can be left out. 
Currently @saper manually builds FreeBSD binaries, so I wouldn't worry about it. If they decide to build it then they can upload it at the same time

.github/workflows/electron-linux.yml
glenn2223: If you see this run
https://github.com/glenn2223/node-sass/runs/514109989?check_suite_focus=true#step:9:11601

glenn2223:
I want to use this in a vscode extension, so support for as many of the below OS's is my aim

https://code.visualstudio.com/download

.github/workflows/electron-linux.yml
nschonni: OK, the other issue I could think that it's running into is that the `npm test` is still using the regular system NodeJS version to execute, but the binding is setup for an electron environment. Not familiar with how to properly test Electron, but is there something like `export NODE=electron_binary`, similar to what python tests do with `export PYTHON=python3` for making sure a particular version is used.

If it really is a CLI issue with Electron, but Electron won't use the test, maybe there is a way to just skip the test on Electron, but run the rest.

.github/workflows/electron-linux.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        os: ${{ matrix.os }}
+    - name: Install packages
+      run: npm install --unsafe-perm
+    - name: Install electron-rebuild
+      run: npm install --save-dev electron-rebuild@latest
+    - name: Run on ${{ matrix.os }} - Electron v${{ matrix.electron }} (Node ${{ matrix.nodeVersion }})
+      run: ./node_modules/.bin/electron-rebuild -v=${{ matrix.electron }} -w node-sass
```
nschonni: If you use `npx` like `npx electron-rebuild -v=${{ matrix.electron }} -w node-sass` you can skip the previous global install.
Don't feel stronly though, since it looks like this works

.github/workflows/electron-win.yml
```diff
@@ -0,0 +1,54 @@
...
+      run: ./node_modules/.bin/electron-rebuild -v=${{ matrix.electron }} -w node-sass
+    - uses: actions/upload-artifact@v1
+      with:
+        name: win32_${{ matrix.arch }}_${{ matrix.nodeVersion }}_binding.node
+        path: build/release/binding.node
+    - uses: actions/upload-artifact@v1
+      with:
+        name: win32_${{ matrix.arch }}_${{ matrix.nodeVersion }}_binding.pdb
+        path: build/release/binding.pdb
+    # npm test failing on pre-existing master issues, commenting out
```
nschonni: If we figure out the test issues, the test step should probably be before the upload. If we end up just skipping it, then we can just delete this rather than comment it out

nschonni:
> I want to use this in a vscode extension, so support for as many of the below OS's is my aim

That's what I figured :), and I think this will cover off all those now

saper:
> > But how do we target FreeBSD and Linux/musl? Will it have to be self-hosted runners?
> 
> I think those can be skipped. Alpine is a container only OS, unless you think people might be running something through Electron that way, it can be left out.
> Currently @saper manually builds FreeBSD binaries, so I wouldn't worry about it. If they decide to build it then they can upload it at the same time

FreeBSD has electron4, 6 and 7 in their ports tree. I could look at mixing that with my build machinery and provide binaries for them as well.

.github/workflows/electron-osx.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        os: ${{ matrix.os }}
+    - name: Install packages
+      run: npm install --unsafe-perm
+    - name: Install electron-rebuild
+      run: npm install --save-dev electron-rebuild@latest
+    - name: Run on ${{ matrix.os }} - Electron v${{ matrix.electron }} (Node ${{ matrix.nodeVersion }})
+      run: ./node_modules/.bin/electron-rebuild -v=${{ matrix.electron }} -w node-sass
```
saper: How does `electron-rebuild` really work? What is used to finally kick off the build? `gyp`? Something else?

glenn2223: You're right. It's kind of like a wrapper over `node-gpy` (the latest version).

You can get the same result [here](https://github.com/sass/node-sass/blob/499e06685c71f446d8314673ece3617020ab6b88/.github/workflows/electron-linux.yml#L42) with node-gyp but it requires a bump, this tweak doesn't and can be applied to the current release 🎆🎉

saper:
This is a CI-only change. I think we should also support building from the command line, so possibly https://github.com/sass/node-sass/blob/master/scripts/build.js should be modified to to support electron. 

Probably we only need to modify `node-gyp` invocation:

https://github.com/sass/node-sass/blob/b54053a1b50fd97e951eb0311a7fb818683a8e99/scripts/build.js#L59-L62

It could be as simple as https://github.com/electron/electron/blob/master/docs/tutorial/using-native-node-modules.md#manually-building-for-electron

     HOME=~/.electron-gyp node-gyp rebuild --target=1.2.3 --arch=x64 --dist-url=https://electronjs.org/headers

and therefore would not even require adding `electron-rebuild` as a dependency.

.github/workflows/electron-linux.yml
glenn2223: > I'll add the data from a local run tomorrow (in GMT here) as I think the error was different on that

```console
  6084 passing (31s)
  753 pending
  2 failing

  1) cli node-sass sass/ --output css/ should not error if output directory is a symlink:
     Error: EEXIST: file already exists, mkdir 'C:\Users\Glenn\Documents\GitHub\node-sass\test\fixtures\input-directory\css'
      at Object.mkdirSync (fs.js:860:3)
      at Context.<anonymous> (test\cli.js:582:10)
      at processImmediate (internal/timers.js:456:21)

  2) cli node-sass --follow --output output-dir input-dir should compile with the --follow option:
     Error: EEXIST: file already exists, mkdir 'C:\Users\Glenn\Documents\GitHub\node-sass\test\fixtures\follow\input-dir'
      at Object.mkdirSync (fs.js:860:3)
      at Context.<anonymous> (test\cli.js:623:10)
      at processImmediate (internal/timers.js:456:21)



npm ERR! Test failed.  See above for more details.
```

**The above is the result when I run `npm test` on the original repository (no changes)**  
Have it open in VSCode Winx62, using vendor node from win32-x64-79 (I have Node 13.10.1)

glenn2223:
> ```
>  HOME=~/.electron-gyp node-gyp rebuild --target=1.2.3 --arch=x64 --dist-url=https://electronjs.org/headers
> ```

Using `node-gyp` requires a bump (fine for v5), that's why I fell on `electron-rebuild` for 4.13

.github/workflows/electron-linux.yml
glenn2223: Just to make sure I understood you correctly, I removed the below step and ran your suggestion
```
npm install --unsafe-perm
```


If so, it produced this error. It needs to `install` to create the `binding.node` it's going to clone

```console err
internal/modules/cjs/loader.js:979
  throw err;
  ^

Error: Cannot find module 'nan'
Require stack:
- C:\Users\Glenn\Documents\GitHub\node-sass\[eval]
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:976:15)
    at Function.Module._load (internal/modules/cjs/loader.js:859:27)
    at Module.require (internal/modules/cjs/loader.js:1036:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at [eval]:1:1
    at Script.runInThisContext (vm.js:128:20)
    at Object.runInThisContext (vm.js:319:38)
    at Object.<anonymous> ([eval]-wrapper:10:26)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at evalScript (internal/process/execution.js:94:25) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'C:\\Users\\Glenn\\Documents\\GitHub\\node-sass\\[eval]' ]
}
gyp: Call to 'node -e "require('nan')"' returned exit status 1 while in binding.gyp. while trying to load binding.gyp
gyp ERR! configure error
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (C:\Users\Glenn\Documents\GitHub\node-sass\node_modules\electron-rebuild\node_modules\node-gyp\lib\configure.js:351:16)
gyp ERR! stack     at ChildProcess.emit (events.js:316:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
gyp ERR! System Windows_NT 10.0.18363
gyp ERR! command "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\Glenn\\Documents\\GitHub\\node-sass\\node_modules\\electron-rebuild\\node_modules\\node-gyp\\bin\\node-gyp.js" "rebuild" "--target=8.0.0" "--arch=x64" "--dist-url=https://www.electronjs.org/headers" "--build-from-source"
gyp ERR! cwd C:\Users\Glenn\Documents\GitHub\node-sass
gyp ERR! node -v v13.10.1
gyp ERR! node-gyp -v v6.1.0
gyp ERR! not ok
```

saper:
> Using `node-gyp` requires a bump (fine for v5), that's why I fell on `electron-rebuild` for 4.13

What happens if we try with 3.8.0? I do not see too many electron related changes (except for maybe for https://github.com/nodejs/node-gyp/pull/1566). What was used to build modules when 3.8.0 was current?

.github/workflows/electron-linux.yml
glenn2223: > OK, the other issue I could think that it's running into is that the `npm test` is still using the regular system NodeJS version to execute, but the binding is setup for an electron environment. Not familiar with how to properly test Electron, but is there something like `export NODE=electron_binary`, similar to what python tests do with `export PYTHON=python3` for making sure a particular version is used.
> 
> If it really is a CLI issue with Electron, but Electron won't use the test, maybe there is a way to just skip the test on Electron, but run the rest.

TBH, I don't know why I'm running `npm test` as we can't test it in that environment (unless I missed something)

The test is running on the system Node but still using the binary it downloaded from GitHub to test. This is because running `electron-rebuild` places the electron binary in the `bin` & `build/Release` folders, completely separate from where the current binary is (obv).

I tried (even though I knew the result) to replace the binary in the `vendor` folder and see what would happen; you guessed it, I got the below error: (essentially trying to run Electron 8 on Node 13)
```console err
Binary has a problem: Error: The module '\\?\C:\Users\Glenn\Documents\GitHub\node-sass\vendor\win32-x64-79\binding.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 76. This version of Node.js requires
NODE_MODULE_VERSION 79. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```

I think we should drop the `npm test` as the binary it is rebuilding off must be functional otherwise it wouldn't be accessible in a public release (or at least that's what we all hope 😅)

Do you agree?

glenn2223:
> > Using `node-gyp` requires a bump (fine for v5), that's why I fell on `electron-rebuild` for 4.13
> 
> What happens if we try with 3.8.0? I do not see too many electron related changes (except for maybe for [nodejs/node-gyp#1566](https://github.com/nodejs/node-gyp/pull/1566)). What was used to build modules when 3.8.0 was current?

See this [error line on GitHub action](https://github.com/glenn2223/node-sass/runs/513436323?check_suite_focus=true#step:6:55).  All 3 builds (that weren't cancelled) failed with the below message:
```console
gyp ERR! This is a bug in `node-gyp`.
gyp ERR! Try to update node-gyp and file an Issue if it does not help:
```

saper:
node-gyp fails because `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\15.0\Bin\MSBuild.exe` cannot be found. Why?

glenn2223:
> `Microsoft Visual Studio\2019\Enterprise\MSBuild\15.0\Bin\MSBuild.exe` cannot be found

Found this: https://github.com/nodejs/node-gyp/issues/1663  
Reading now

glenn2223:
So node-gyp was searching for MSBuild\15.0\Bin\. VS19 changed this to MSBuild\Current\Bin\
This seems to have been addressed with this pull: https://github.com/nodejs/node-gyp/pull/1762

saper:
Looking at `appveyor.yml` files in the https://github.com/electron/electron/ it seems to me (correct me if I am wrong) that electron 5, 6 and 7 are built using VS2017. Electron 8 upwards use VS2019. 

We should use the same compiler (or rather PlatformToolkit version) that has been used to build electron itself for the module to avoid potential ABI compatibility issues like https://github.com
/sass/node-sass/issues/1283

If GitHub Action can use only VS2019, we should probably build only 7 and 8 (if my assumptions about electron build are correct).

Could setting `npm config set msbuild_path "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\MSBuild.exe"` workaround the issue (until we switch to v5)?

glenn2223:
> Looking at `appveyor.yml` files in the https://github.com/electron/electron/ it seems to me (correct me if I am wrong) that electron 5, 6 and 7 are built using VS2017. Electron 8 upwards use VS2019.
> 
> We should use the same compiler (or rather PlatformToolkit version) that has been used to build electron itself for the module to avoid potential ABI compatibility issues like https://github.com
> /[/issues/1283](https://github.com/sass/node-sass/issues/1283)
> 
> If GitHub Action can use only VS2019, we should probably build only 7 and 8 (if my assumptions about electron build are correct).

Don't think it's a problem.

**Context:**  
I created an app with electron 5.0.11, installed `node-sass`, created a render script and ran it. Got the below result.
Windows - Electron Fail 70 not supported

I then retrieved the new node I created from [my action here](https://github.com/glenn2223/node-sass/actions/runs/57616816) and manually added it to the `/node_modules/node-sass/vendor/` folder. Re-ran the app and... success!
Windows - Electron Fail 70 Supported - YAY

***Quick sidebar:** I noticed that the install script used my node rather than electrons. May it be worth checking the start script in the `package.json` file for `"scripts: { start: "electron *" }` (if it's even possible)*

saper:
> Don't think it's a problem.

Sure, a script or two may ran fine, but believe me, ABI compatibility problems for C++ components are hard to debug. 

Microsoft says that [2015, 2017 and 2019 are binary compatible](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2017), something they could not guarantee before. But they say one should use the latest "redistributable" runtime library. Will Electron 5,6,7 bring only 2017 runtime library with itself? If yes, we should make build with 2017 to make sure 2019 is not required.

glenn2223:
> > Don't think it's a problem.
> 
> Sure, a script or two may ran fine, but believe me, ABI compatibility problems for C++ components are hard to debug. 

You're right. When you're right, you're right!

---

> Microsoft says that [2015, 2017 and 2019 are binary compatible](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2017), something they could not guarantee before. But they say one should use the latest "redistributable" runtime library. Will Electron 5,6,7 bring only 2017 runtime library with itself?***(1)*** If yes, we should make build with 2017 to make sure 2019 is not required.***(2)***
1) How can I find out?  
2) Do we need to move away from GitHub's runners then?

---

***Quick sidebar:** When I was running my test I noticed that the install script used node's `modules` rather than electron's `electron`. May it be worth checking for `process.versions.electron` and getting the correct binary on building? Again, only Electron 5+, as they have different `modules` to node (from what my builds have shown me anyway)*

glenn2223:
> > Microsoft says that [2015, 2017 and 2019 are binary compatible](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2017), something they could not guarantee before. But they say one should use the latest "redistributable" runtime library. Will Electron 5,6,7 bring only 2017 runtime library with itself?_**(1)**_ If yes, we should make build with 2017 to make sure 2019 is not required._**(2)**_
> 
> 1. How can I find out?

Don't know if this answers the question or not?  
Quote from [Electron v5 prerequisites](https://github.com/electron/electron/blob/v5.0.0/docs/development/build-instructions-windows.md#prerequisites): "Visual Studio 2017 15.7.2 or higher"

.github/workflows/electron-linux.yml
glenn2223: Got what you meant 🤦‍♂️. Switched it, looks cleaner.

glenn2223:
> > Microsoft says that [2015, 2017 and 2019 are binary compatible](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2017), something they could not guarantee before. But they say one should use the latest "redistributable" runtime library. Will Electron 5,6,7 bring only 2017 runtime library with itself?_**(1)**_ If yes, we should make build with 2017 to make sure 2019 is not required._**(2)**_
> 
> 1. How can I find out?
> 2. Do we need to move away from GitHub's runners then ([ref](https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners))?

After a bit more rummaging I can see that GitHub runners (Windows Server 2016) has [Visual Studio 2017](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2016-Readme.md#visual-studio-2017-enterprise). If we're happy that `"Electron 5,6,7 bring only 2017 runtime library"` then I can target that OS and we're ready??

saper:
I have filed https://github.com/sass/node-sass/issues/2868 to work on the general Visual Studio 2019 issue.

glenn2223:
> Could setting `npm config set msbuild_path "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\MSBuild.exe"` workaround the issue (until we switch to v5)?

Ref lines [153-155](https://github.com/nodejs/node-gyp/blob/v3.8.0/lib/configure.js#L153-L155) from v3.8.0 configure script & my [many failed actions](https://github.com/glenn2223/node-sass/actions?query=workflow%3A%22Electron+Rebuild+CI+-+Windows+%28x32+%26+x64%29%22)
```js
// Copied from v3.8.0's configure script //

  // GYP doesn't (yet) have support for VS2017, so we force it to VS2015
  // to avoid pulling a floating patch that has not landed upstream.
  // Ref: https://chromium-review.googlesource.com/#/c/433540/
```
As you can see my attempts fail and the comments (copied in above) seem to indicate that it shouldn't be overridden.

This was my last ditch effort, but failed:
```cmd
setx PATH "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin;%PATH%"
set path="C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin;%PATH%"
```

nschonni:
I'm OK with this being a CI only setup, there may be some edge cases where local fallback fails, but we can consider this experimental support for v4.
If you really want to try to get the specific toolsets for the versions changed, I'd suggest changing the `runs-on` values between `windows-2016` for VS 2017 and `window-2019` for the VS 2019 with a CI only install of newer node-gyp. Not sure this is worth it unless you're finding that the version produced by CI isn't working for your local VS Code install.

glenn2223:
> I'm OK with this being a CI only setup, there may be some edge cases where local fallback fails, but we can consider this experimental support for v4.

Me too!!

> If you really want to try to get the specific toolsets for the versions changed, I'd suggest changing the `runs-on` values between `windows-2016` for VS 2017 and `window-2019` for the VS 2019 with a CI only install of newer node-gyp.

From saper's comment and reading the [linked Microsoft docs](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2017), it seems that if we target 5, 6 and 7 with 2019 then we're forcing people who build apps with those electron versions to use a 2019 toolset/VS 2019. This is because those electron builds are compiled with VS 2017 and the binaries require the use of the highest level toolset from any component (To dummy it down, not meaning to be condescending or anything. `Electron: 2017 build` + `binary: 2019 build` => `2019 toolset required`).  
_**That's even if people are still using VS 2017 for their electron 5, 6 or 7 apps?**_

Do you think it's worth it? \*shrug\* just in case? Not sure, maybe overkill?

>Not sure this is worth it unless you're finding that the version produced by CI isn't working for your local VS Code install.

The binary works locally with VSCode (latest) on Windows x64, but you can download the [forked VSCode extension](https://github.com/glenn2223/vscode-live-sass-compiler) and be my guinea pig if you like 😆?

xzyfer:
FWIW I think it makes sense to target this patch for v5. Then only need to support Node 10, 12, 13

glenn2223:
> FWIW I think it makes sense to target this patch for v5. Then only need to support Node 10, 12, 13

Don't suppose there's an eta for v5?

xzyfer:
If there's no objects from the team I'm keen to tighten the bolts and ship it asap. Thoughts? (Sorry can't @ on my phone)

saper:
I am against pushing CI-only setup. `npm rebuild` will no longer work. People will start coming at us when upgrading electron versions. Once we start shipping binaries and electron will become a supported engine I do not see a way for the users to rebuild their bindings themselves if needed.

glenn2223:
**Not related to above**

Just as a personal learning experience I created a [new branch](https://github.com/glenn2223/node-sass/blob/Electron-Test) to see how it would work trying to [manipulate the build](https://github.com/glenn2223/node-sass/blob/Electron-Test/scripts/build.js) rather than have a CI. Running an electron `prestart` script to build the new binding is getting this error. Any clues?

> gyp: name 'openssl_fips' is not defined while evaluating condition 'openssl_fips != ""' in binding.gyp while trying to load binding.gyp
> App threw an error during load
> Error: Command failed: "C:\Users\Glenn\Documents\GitHub\Sample-Electron\node_modules\electron\dist\electron.exe" C:\Users\Glenn\Documents\GitHub\Sample-Electron\node_modules\node-sass\scripts\build --directory="./node_modules/node-sass/"
> gyp: name 'openssl_fips' is not defined while evaluating condition 'openssl_fips != ""' in binding.gyp while trying to load binding.gyp
> 
>     at checkExecSyncError (child_process.js:636:11)
>     at execSync (child_process.js:673:15)
>     at electron/js2c/asar.js:761:23
>     at Object.<anonymous> (C:\Users\Glenn\Documents\GitHub\Sample-Electron\scripts\electron-build-node-sass.js:14:13)
>     at Module._compile (internal/modules/cjs/loader.js:839:30)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:851:10)
>     at Module.load (internal/modules/cjs/loader.js:701:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:633:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:625:3)
>     at loadApplicationPackage (C:\Users\Glenn\Documents\GitHub\Sample-Electron\node_modules\electron\dist\resources\default_app.asar\main.js:107:12)

*Function call*
```js
// if binding not found =>call
exec(`"${process.execPath}" ${Path.resolve("./node_modules/node-sass/scripts/build")} --directory="./node_modules/node-sass/"`);
```
_**Side note:** had to add [this function](https://github.com/glenn2223/node-sass/blob/Electron-Test/lib/extensions.js#L23-L37) to get the correct node module and replaced instances of `process.versions.modules` with it_

nschonni:
> If there's no objects from the team I'm keen to tighten the bolts and ship it asap. Thoughts? (Sorry can't @ on my phone)

@xzyfer there's an app now ;) https://github.com/mobile/

glenn2223:
> If there's no objects from the team I'm keen to tighten the bolts and ship it asap. Thoughts? (Sorry can't @ on my phone)

@xzyfer not that I'm part of the team, but if there's anything I can do to help nail down v5 then let me know. Do you have a milestone with required tasks for v5?

xzyfer:
@nschonni giving the app a whirl 🤞

@glenn2223 I think we've captured all the breaking changes we're expecting for v5. The last big outstanding change in N-API support but I'd prefer to push that to V6.

IMHO I'm happy to cut the release in the next couple days and roll forward with feature updates like this.

glenn2223:
Opened v5 PR, closing

PR #2866, opened 2020-03-18
update `mkdirp` to v1.0.3

SampsonCrowley:
fixes #2865

PR #2881, opened 2020-03-30
dep: remove mkdirp dependency

rharriso:
* dep: mkdirp hasn't been updated in 4 years, and node's fs library
has had a recursive option for `fs` since v10.12 (oldest active LTS
version). Additionally `mkdirp` depends on a version of library `minimist` that has a minor vulnerability associated.

Refs:
* https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback
* https://npmjs.com/advisories/1179


nschonni:
Thanks, this was updated in the v5 branch, but it can't be removed entirely yet as per https://github.com/sass/node-sass/pull/2866#issuecomment-603677083

rharriso:
@nschonni Sorry for not double checking for duplicate work.

nschonni:
No worries, thanks for taking the time. I think once we drop 10, this would be a welcome change

PR #2903, opened 2020-04-25
Assorted lint tweaks

XhmikosR:
* remove useless conditional in bin/node-sass; `sourceMapIsDirectory` always evaluates to true
* remove unneeded else blocks
* remove lonely if blocks
* remove useless escape characters from regular expressions
* disable use of Object.prototypes builtins directly
* throw a `TypeError` instead of a generic `Error` when the options isn't an Object
* specify explicitly 10 as the radix for `parseInt`
* remove the path segments from require statements
* remove spaces before function parentheses for consistency
* replace the `package` reserved keyword in test/api.js with `pkg`
* enforce end of file newlines
* disallow yoda expressions
* sort the ESLint rules by name so that it's easier to find a rule
* lint all the files for consistency

Let me know if you need me to make any changes 🙂

PS. I only tested this on GitHub Actions CI on my fork.
~~PPS. If Node.js 6.x is the minimum supported version, I guess we could use more ES6 features.~~ **EDIT**: Just saw the Travis matrix.


XhmikosR:
It seems tests fail on Node.js 0.10 and 0.12. I will try to pinpoint why and revert the related changes.

nschonni:
Please submit the changes separately

XhmikosR:
NVM then.

PR #2911, opened 2020-04-30
Add .cirrus.yml for FreeBSD (amd64, i386) CI/CD

saper:
This is an update to #2592

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -39,7 +39,7 @@ freebsd_build_task:
       if [ -n "${working_dir}" ]; then change_dir="cd ${working_dir} &&"; fi
       echo "$change_dir npm install --unsafe-perm --production" | $exec_prefix /bin/sh
       echo "Generated files:"
-      echo "$change_dir find vendor | $exec_prefix /bin/sh
+      echo "$change_dir find vendor" | $exec_prefix /bin/sh
 
 publish_task:
   only_if: $CIRRUS_TAG != ""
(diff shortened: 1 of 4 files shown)


scripts/configure_freebsd_ci_jail.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env sh
+
+set -o xtrace
```
nschonni, line 3: `set -e`?

.cirrus.yml
```diff
@@ -0,0 +1,56 @@
+freebsd_instance:
+  image_family: freebsd-11-3-snap
+
+freebsd_build_task:
+
+  env:
+    matrix:
+      - package: npm-node10
```
nschonni, line 8: How do we expand out the matrix to the other versions we currently list as supported for FreeBSD?

saper: You are right, I forgot it. Back to work :)

scripts/configure_freebsd_ci_jail.sh
saper: I planned to rewrite both of those shell scripts but forgot about them (didn't test upload either). Resetting to draft. Thanks!

xzyfer:
I think I've clicked all the buttons.

.cirrus.yml
xzyfer: We can limit this to v5 also to make our lives a bit easier.

saper: I've been only building binaries for node version currently in the FreeBSD ports tree, which are 10, 12 and 13 now. Pushed the commit to expand the matrix.

nschonni: We can drop 10 now, but it would be good to try and get 12, 14, 16, 17 to match the other platforms

PR #2917, opened 2020-05-09
Updated sass-graph due to security vulnerability

tabithahsia:


xzyfer:
This has already been addressed in the latest release

PR #2919, opened 2020-05-11
fix: Flatten sourceMap folder detection

nschonni:
CodeQL was flagging the second `else if`.
Looked at code and inlined the single use so it's not executed if
first condition is true.

### DIFF
--- a/bin/node-sass
+++ b/bin/node-sass
@@ -204,15 +204,15 @@ function getOptions(args, options) {
       options.sourceMapOriginal = options.sourceMap;
     }
 
-    // check if sourceMap path ends with .map to avoid isDirectory false-positive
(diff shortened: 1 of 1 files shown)


nschonni:
@xzyfer trying out CodeQL/LGTM.com, but this was the only thing it found right now. There are some things on the libsass side, but I'll open an issue over there

xzyfer:
Is the same possible without inlining the condition? The variable name was useful in conveying the intent. I think that's valuable given the messy state of the sourcemapa logic.

nschonni:
It is, but then it's evaluating the directory aspect even in the "happy path". Alternately, I could move it inside the final `else`

nschonni:
OK, moved it into the inner `else`. Does that look better?

Issue #2685, opened 2019-06-04
libsass 3.6.0


aldarund:
https://github.com/sass/libsass/releases/tag/3.6.0

ItsLeeOwen:
https://github.com/sass/libsass/releases/tag/3.6.1

I forked node-sass to integrate the latest libsass 3.6.1 commit #2b8a17a, mocha tests fail, but it does build successfully, and I'm able to use new libsass features like @content arguments. https://github.com/sass/node-sass/pull/2714

dylancopeland:
Is adding support for 3.6 anywhere on the roadmap?

xzyfer:
It's a decent amount of work so it'll happen when I have a couple free days
to build the binaries.

mbaluda:
libsass contains a HIGH severity vulnerability that is fixed in 3.6.0
https://nvd.nist.gov/vuln/detail/CVE-2018-19827
https://github.com/sass/libsass/commit/b21fb9f84096d9927780b86fa90629a096af358d

Can you provide a timeline for the update?

ZebraFlesh:
libsass 3.6.0 also contains a fantastic bug fix sass/libsass#2849 that fixes some Windows specific compilation errors.

eusonic:
And it implements passing arguments to content blocks, which is super useful. https://sass-lang.com/documentation/at-rules/mixin#passing-arguments-to-content-blocks

AdamNatale:
Is there any update on this?

bricss:
PR with upgrade is kinda waiting for several months, already ↪️ https://github.com/sass/node-sass/pull/2714

saper:
> PR with upgrade is kinda waiting for several months, already arrow_right_hook #2714

Does not seem that way to me, sorry!

saper:
@xzyfer What do you think about producing a 4.x release with libsass 3.6.x?

johnnyBira:
Is this happing any time soon? I'm looking forward to mixin content block arguemnts.
Thanks for making an maintaining this wonderful library <3

karthikrajthandapani:
![image](https://user-images.githubusercontent.com/57178800/67855318-36f0c580-fb38-11e9-8049-e0c5fd10f1f4.png)
As per Azure Dev Ops report, libsass contains a HIGH severity vulnerability that is fixed in 3.6.0
Please can you provide timeline for the update?

saper:
[...]

PR #2925, opened 2020-05-20
fix memory leak based from 2018 patch on parse_ie_keyword_arg()

coderlava:
Affected versions of this package are vulnerable to Out-of-bounds Read via the function Sass::Prelexer::exactly() which could be leveraged by an attacker to disclose information or manipulated to read from unmapped memory causing a denial of service.


nschonni:
Thanks, but we can't modify the libsass subtree directly. Follow https://github.com/sass/node-sass/issues/2685 for updates for bumping libsass version

PR #2944, opened 2020-07-17
improved language to be more sensitive to less privileged populations

rgeerts:
I believe I only changed how things were commented.  (blm)


nschonni:
Thanks! Can you remove the changes to the files under the `libsass` folder and submit them to https://github.com/sass/libsass, since that directory is vendored in from that repo

PR #2955, opened 2020-08-07
Update package.json

RamyaPayyavula:
older versions of lodas has injection vulnerability. An attacker can inject malicious code via `sourceURL` since it is not sanitized for the user-provided code that leads to the `eval()` function.


nschonni:
We don't lock the version of the dependency. Regenerating your local lock file should address this

Issue #2851, opened 2020-02-15
request has gone into maintenance mode. Maybe replace it.


Tchiller:
Dear team

thanks for your work!

I would like to report, that the package `request` which you are loading has went into maintenance mode https://github.com/request/request/issues/3142

Therefor, yarn is currently giving me 

> 
> warning node-sass > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
> warning node-sass > node-gyp > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

Maybe you find time (and a good replacement) to replace it for your upcoming major release. 
Thanks!

saper:
That is not a big issue. We could possibly drop the usage of request completely but this could mean some issues for users connecting via proxy.

cekvenich:
ftw I used node-fetch

nschonni:
As your message shows, "request" is also used by node-gyp, so removing it here wouldn't remove the transitive dependency or that second warning. Updating node-gyp is something we're not looking at till the next major version because of breaking changes

nschonni:
> ftw I used node-fetch

Doing a quick look shows it doesn't have any support for proxies, and wouldn't be a suitable replacement

cekvenich:
@nschonni Thanks for a response and a wonderful lib used by me and thousands. 

I do think it puts pressure on gyp if they are the only ones not solving it. However, I have no awareness of the effort, so no matter what, thank you for your efforts!

xzyfer:
We've looked at this a few times in the past. Previously we've been blocked on backwards compatibility since most request alternatives would require us to drop legacy Node support.

With v5 coming up we're in a position to make this breaking change. I've re-surveyed the http library landscape based the following resources:
- https://github.com/request/request/issues/3143
- https://www.npmjs.com/package/got#comparison

The most viable alternatives IMHO in order of preference are: 
- https://github.com/npm/make-fetch-happen
- https://github.com/axios/axios

There are some shinier, newer options but most refuse to support proxies out right. For this reason `make-fetch-happen` is the most appealing as it has robust proxy support as a result of being used within npm itself.

SampsonCrowley:
[...]

PR #2961, opened 2020-08-31
WIP: feat: Remove deprecated request package in favor of axios

dunklesToast:
As request was deprecated at the begin of this year I thought it would be good if it would get replaced.

The functionality is already ported to axios but I have some issues with the tests. Would be great if somebody could assist me there.


scripts/install.js
```diff
@@ -9,6 +9,7 @@ var fs = require('fs'),
   sass = require('../lib/extensions'),
   request = require('request'),
```
stephanvierkant, line 10: Should be removed?

stephanvierkant:
You've added yarn.lock (by accident?).

dunklesToast:
I added yarn.lock because I use yarn as a package manager. I will remove it once the PR is ready to merge.

scripts/install.js
dunklesToast: Yes. Missed that.

easingthemes:
Check https://github.com/sass/node-sass/issues/2851, `request` replacement. is in consideration, and final plugin is not decided yet.

dunklesToast:
Okay it seems that make-fetch-happen should be used. If that decision is final I'd like to create a PR for that.

Issue #2965, opened 2020-09-18
Node 15 Support


nschonni:
Version 15 is supposed to be released on October 20th, 2020. Few things that should probably happen to make it smoother going forward:

- [ ] Promote the fixups in the v5 branch
- [x] Drop testing on unsupported Node versions (<10)
- [ ] Allow people to try and build on old node version, but show warning that we won't support issues opened for those platforms
- [ ] Hold off 3.6 libsass upgrade for separate major release once there is a major version people can pin to

nschonni:
@xzyfer I believe the release is expected out today (the 20th). It may just be a matter of restarting the CIs on https://github.com/sass/node-sass/pull/2983 or we might have to wait for AppVeyor for the x86 build

xzyfer:
Traditionally we have been blocked on appveyor and Travis to add the new node versions.

xzyfer:
I can build the OSX binaries which will mean we're not blocked on Travis.

nschonni:
I think OSX/Linux should be fine on the GitHub Actions, since they should upload the binaries. I think the gaps will be:
- [ ] Windows x86 on AppVeyor
- [ ] Apline on GitHub Actions will require a new nodejs/docker-node PR and publish on official-images repo
- [ ] FreeBSD binraries from @saper unless we get the Cirrus CI figured out

Remembered that we're actually fully of Travis now, so we're not waiting for that part anymore

xzyfer:
I'm not sure I understand the alpine issue. Alpine images are usually
released alongside the Debian images. Are they not suitable?

nschonni:
We use the Node.js Alpine images from https://github.com/nodejs/docker-node, which will need to release a v15 version. Once a PR is merged there, it creates another PR to https://github.com/docker-library/official-images, which is then merged and built out on their CI to publish to the Docker Hub

xzyfer:
Sorry I think I'm just out of touch with how the new GitHub checks work.
I'll read up on it.

nschonni:
The Alpine stuff is here https://github.com/sass/node-sass/blob/master/.github/workflows/alpine.yml, it pulls and runs on the "official image" for Node.js. Rest of the Actions use the https://github.com/actions/setup-node/ to install a particular version on Node.js

xzyfer:
I quickly skimmed the action and it looks like it might be using an arch
Linux mode binary. I may still be missing something however

Typically we build the binaries on CentOS for better compatibility across
[...]

PR #2970, opened 2020-09-26
chore: Remove CI for unsupported Node versions

nschonni:
Was trying a larger branch/PR for #2965, but ended up having too much diff when I hit an issues. Will just submit these in smaller chunks for now and figure out what to cherry-pick from v5 separately

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -15,13 +15,13 @@ jobs:
         node:
           - 10
           - 12
-          - 14.5
+          - 14.6
         include:
           - node: 10
             alpine: "3.9"
           - node: 12
             alpine: "3.9"
-          - node: 14.5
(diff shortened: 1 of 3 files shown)


nschonni:
@xzyfer any clue why the sass.type tests would be failing on Node 14 only? I think this was the issue I ran into before on my v5-lite branch

nschonni:
Oh, right, I think I hit this and tried to fix this in the other branch. the `assert` behaviour was depricated in previous versions, and must be updated in 14

nschonni:
OK, not the `assert` behaviour changes. Looks like the `toString` is actually returning something different in the current v14 than previous node versions

nschonni:
Closing since landed version pinning to 14.5 for now. New PR should be sent to fix test and unpin

PR #2971, opened 2020-09-26
chore: Drop old versions from CI and Travis to GitHub Actions

nschonni:
Leaving #2970 for troubleshooting why Node 14.6+ is breaking the tests.

- Drop all EOL and old unstable releases from testing. Doesn't touch the package.json engines value or extensions.js as part of this PR
- Pin to 14.5 to avoid current `toString` issue
- Add Windows GitHub Actions, but can only currently cover x64 until https://github.com/actions/setup-node/issues/190
- Move JS Linting to separate job that only triggers for JS changes
- Move coverage job to separate job. May need some work for GitHub Action publishing
- Move Linux building from Travis to GitHub Actions, and keeps the GCC/G++ versions
- All GitHub actions include the build results as artifacts to enable releasing, but that would still be manual


xzyfer (approved):
🤲

PR #2978, opened 2020-09-28
chore: Update Istanbul to NYC and bump Mocha

nschonni:
- Update mocha to latest, which drops the the `npm audit` warnings
- Removes the depreciated istanbul package with nyc
- Drops Coveralls, since we can just use it only in CI now
- Coverage does decrease, since it now globs in more of the "scripts", not currently covered by tests


.nycrc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "all": true,
+  "include": [
+    "bin/*",
```
nschonni, line 4: Don't think I quite got this right, but I also don't think it was working before either

xzyfer (approved):

nschonni:
Oops, should have waited for CI to finish. I think Windows isn't exited properly now. Quick search sounds like it's related to uncleaned up references so I'll see what I can find

nschonni:
OK, narrowed it down to the CLI tests. I'll try troubleshooting it a bit before resorting to their hammer of `--exit` https://mochajs.org/#-exit

nschonni:
OK, found an "easy" fix, of updating cross-spawn. PR incoming

Issue #2972, opened 2020-09-27
v8 8.4 in Node 14.6+ changed the toString behaviour


nschonni:
- NPM version (`npm -v`): 6.14.8
- Node version (`node -v`): 14.12.0, but anything above 14.5.0 has this
- Node Process (`node -p process.versions`):
```json
{
  node: '14.12.0',
  v8: '8.4.371.19-node.16',
  uv: '1.39.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.16.0',
  modules: '83',
  nghttp2: '1.41.0',       
  napi: '7',
  llhttp: '2.1.2',
  openssl: '1.1.1g',       
  cldr: '37.0',
  icu: '67.1',
  tz: '2020a',
  unicode: '13.0'
}
- Node Platform (`node -p process.platform`): win32, but all in CI as well
- Node architecture (`node -p process.arch`): x64, but all in CI as well
- node-sass version (`node -p "require('node-sass').info"`):
- npm node-sass versions (`npm ls node-sass`): 4.14.1

<!--

When encountering a syntax, or compilation issue:

- Please note that we cannot backport fixes to old versions, so ensure that you are running the latest release https://github.com/sass/node-sass/releases
- Search for duplicate or closed issues https://github.com/sass/node-sass/issues?utf8=%E2%9C%93&q=is%3Aissue
- Validate with http://sassmeister.com/ that the code works with Ruby Sass, then open an issue on `LibSass` https://github.com/sass/LibSass/issues/new

Sorry you didn't have the experience you expected.

-->

Upstream node issue https://github.com/nodejs/node/issues/35365 was closed as a wont-fix, as the behaviour is a change in v8.
The solution seems to be set the @@toStringTag or https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag

**Comment by xzyfer** [comment]
Does this mean the latest release doesn't work on the latest node 14?

**Comment by nschonni** [comment]
It still runs, but the the `toString` in the following now just returns `[object Object]`, rather than `[object SassColor]` like it was previously, which is way CI was failing on 14.6 and above
```js
      var t = sass.types.Color();
      assert.equal(t.toString(), '[object SassColor]');
```
I don't think this makes a real functionally difference for the runtime, except maybe some extensions or advanced usage.

If we can't figure out the fix, then the tests could possibly be disabled, but I think trying to find a fix would be a good idea first

[...]

PR #2979, opened 2020-09-28
chore: Skip constructor tests on v14.6+

nschonni:
Related to tracking issue https://github.com/sass/node-sass/issues/2972


nschonni:
I've been trying to find the right way to fix these, but as a temporary option, this would let us get setup in the CITGM setup again to avoid future surprises like this

xzyfer (approved):
This will make node 15 support easier. Nice one.

PR #2982, opened 2020-09-29
chore: Prep README for v5

nschonni:
- Add an attempt at a support policy for Node versions at the top of the readme. I removed the old heading text, so some old links might break
- Replaced the Travis status badges with the new GitHub Actions ones
- Removed old status badges for services we don't really rely on

@xzyfer should the Slack on be removed as well? I don't know if it gets much attention by anyone


xzyfer:
Yeah lets remove the slack link

xzyfer (approved):

PR #3005, opened 2020-10-21
chore: Don't double build DependaBot PRs

nschonni:
These are submitted by PR, so branches don't need to build


xzyfer (approved):

nschonni:
@xzyfer AppVeyor already de-duplicates these builds, but the downside of Actions is that I don't think they have a native way of doing that right now, unless you know

xzyfer:
I haven't looked into actions at all. I know we had the same issue with Travis too if you pushed to an origin branch, instead of your fork.

PR #2983, opened 2020-09-29
chore: Add Node 15 support

nschonni:
This should fail till the actual v15 of Node comes out.
Platform compiler requirements based off of https://github.com/nodejs/build/issues/2423

PR #3012, opened 2020-10-27
feat: Update to 5.0.0

nschonni:
#2983 didn't include the update to the package.json, so the tag likely needs to be fixed too

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "node-sass",
-  "version": "4.14.1",
+  "version": "5.0.0",
   "libsass": "3.5.5",
   "description": "Wrapper around libsass",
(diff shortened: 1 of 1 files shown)


xzyfer:
I'll do this with the npm publish. The tag doesn't matter because publish
pushes a tarball to the registry.

nschonni:
OK, do you just want the binaries on the Release page even if the commit is off by one?

xzyfer:
Yeah as long as the binaries are on the release we're good. We only care
that the release tag and non version are the same. The commits don't matter.

nschonni:
OK, I'm uploading them now, so I'll close this off and ping you on the other PR once it's done

PR #3066, opened 2021-02-04
upgraded lodash version 4.17.15 to 4.17.20 to fix reported issue

Salinn:
What:

Looking to update lodash from version 4.17.15 to 4.17.20 to fix https://snyk.io/test/npm/lodash/4.17.15

Why:

Upgrade to a version without the know vulnerability


nschonni:
Thanks, we don't lock the dependency so the existing version range will already install this version. Rebuild your local lockfile if you're getting warnings with your security tool

Issue #3067, opened 2021-02-05
Security Vulnerability Issue [CVE-2020-24025]


LiuJinghao:
[https://nvd.nist.gov/vuln/detail/CVE-2020-24025](https://nvd.nist.gov/vuln/detail/CVE-2020-24025)

Certificate validation in node-sass 2.0.0 to 4.14.1 is disabled when requesting binaries even if the user is not specifying an alternative download path.

[https://github.com/sass/node-sass/pull/567#issuecomment-656609236](https://github.com/sass/node-sass/pull/567#issuecomment-656609236)

Version 5.0.0 was released in October, but through reading the source code of 5.0.0, we found that this issue is still unresolved.

Is there a plan to fix this issue?

zmariscal:
I noticed https://github.com/sass/node-sass/pull/3086 has been closed. Is there any plans to resolve this CVE or a documented workaround?

saper:
The workaround is to download or compile the binding.node file itself (https://github.com/sass/node-sass/blob/master/README.md#rebuilding-binaries) and provide that binary to your installations either via direct installation in the `vendor` subdirectory or by using the [environment variables](https://github.com/sass/node-sass/blob/master/README.md#binary-configuration-parameters).

xzyfer:
Given we plan to bump the major in the release I don't see why we can't
replace the hardcoded false value, and rely on the default true value which
can be overridden by an env supported by node.

AWare:
@zmariscal my sincere apologies for not following through. I found the process highly dispiriting, and the workaround we decided upon was to stop using node-sass and move to the dart implementation. 

If you (or anyone else really) are willing to set up an https interceptor, and do the requested edge-case testing; then I'd be happy for you to take over the PR- it's otherwise fine.

zmariscal:
@xzyfer is there any movement on this issue? I know @AWare was working on https://github.com/sass/node-sass/pull/3086. Should they open the PR back up?

LiuJinghao:
I noticed #3149 has been merged， and version [7.0.0](https://github.com/sass/node-sass/releases/tag/v7.0.0)  was released in November. It will set rejectUnauthorized to true by default In version 7.0.0.

v7.0.0 can fixed  this issue.

PR #3086, opened 2021-04-13
Remove rejectUnauthorized:false from download flags

AWare:
This should resolve https://github.com/sass/node-sass/issues/3067 `rejectUnauthorized: false` was added in https://github.com/sass/node-sass/pull/567 

It would seem more important to resolve the CVE-2020-24025 than to support misconfigured enterprise proxies.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -18,6 +18,13 @@ should always follow these steps before opening a new issue.
 
 ## Installation
 
+### Problems Downloading node-sass
+
+If node-sass is unable to download the sass binary, please download the correct build
+for your node version and platform from the [releases page](https://github.com/sass/node-sass/releases)
+and use the [binary configuration parameters](https://github.com/sass/node-sass/blob/master/README.md#binary-configuration-parameters)
+to manually use the downloaded binary. This may occur if your ISP/organisation is interfering with https connections.
+
 ### 404s
 
 If you see a 404 when trying to install node-sass, this indicates that you're trying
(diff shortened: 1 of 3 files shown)


nschonni:
I'd be OK with dropping this, but I think we need some sort of documentation or environment variable opt-in.
I'm not sure if https://stackoverflow.com/questions/35633829/node-js-error-process-env-node-tls-reject-unauthorized-what-does-this-mean covers off the same behaviour, or if a documented `NODE_SASS_REJECT_UNAUTHORIZED` makes more sense

AWare:
Thanks for looking at this. 

If the libsass opinion is to have it on a switch, I would like to respectfully disagree. HTTP support is absolutely the expected behaviour by end users, and this CVE has prompted us to reevaluate our use of node-sass. 

I would personally be in favour of suggesting users download libsass themselves if it cannot be done safely over https- I don't think the documentation would benefit from detail here. I would be more than happy to add something to this effect to TROUBLESHOOTING.md.

saper:
How to set one's own custom trusted root certificate instead for poor folks behind SSL intercepting corporate proxies?

AWare:
As described, they can provide the binary themselves.

xzyfer:
The linked SO post suggest `NODE_TLS_REJECT_UNAUTHORIZED = "0" npm install node-sass` would have the current behaviour. If this can be confirmed then we can ship this with that tip to the troubleshooting guide.

AWare:
image

I would prefer if that message were provided by the team, as I don't feel comfortable recommending that users disable HTTPS.

saper:
> I would prefer if that message were provided by the team, as I don't feel comfortable recommending that users disable HTTPS.

What *exactly* happens with this pull request if going through, say, SSL-intercepting proxy? What message the users will get? We have to work from there.

AWare:
Sorry but I have no way of testing this.

PR #3090, opened 2021-04-20
feat: Add Node 16 CI and drop Node 10

### LINKED ISSUE #3077: Node 16 support
Node 16 is currently scheduled for April 20 2021 https://github.com/nodejs/Release

- [ ] Update AppVeyor
- [ ] Update GitHub Actions
- [ ] Add module version 93 to https://github.com/sass/node-sass/blob/master/lib/extensions.js#L83
- [ ] Release v5.1.0

/cc @xzyfer @saper

### LINKED ISSUE #3094: Drop Node 10 when it hits EOL on 2021-04-30
It hits EOL on 2021-04-30.

- [ ] Update support table in readme
- [ ] Remove from CI

nschonni:
- Bump to 5.1.0
- Add CI builds for 16
- Add extension lookup for module 93

Fixes #3077
Fixes #3094


nschonni:
Node 16 alpine image isn't out yet, so I expect this to fail for now

pedrofurtado:
Nodejs 16 is out! 🎉🎉  Happy to see the related PR merged and released asap 🥇 🍻

nschonni:
@xzyfer looks like there is an internal compiler issue with some of the headers on macOS and Linux. Does this look like a Nan thing, or maybe a compiler version one. I tried bumping to GCC 8 since its' the new minimum, but I don't think it helped

saper:
* node 16 alpine containers do not seem to be published there yet
* FreeBSD did not merge in node 16 into the ports system

saper:
Experimenting with https://github.com/sass/node-sass/pull/3091

xzyfer:
Looks like the OSX failure might have a fix https://github.com/microsoft/node-pty/pull/433/files

saper:
#3091 seems to fix this, I wonder if we should be using the same compiler settings for all supported versions...

iamsivin:
Any news or date on the release of V 5.1.0 that has support for node V 16. Facing issue for M1 mac users.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "node-sass",
-  "version": "5.0.0",
+  "version": "5.1.0",
```
nschonni, line 3: @saper @xzyfer should we just bump this to 6 and also drop Node 10 with #3094? Would mean maybe waiting till the end of the week for cutting the release

saper: I'd like to have more research about actual compilers and C++ standards we need. Hope we do not need to use different once for different node versions?

I think one node version out, one node version in is a fair trade, we could do 6.x this way.

nschonni: I dropped the conditionals entirely here, and the CI seems OK, but that's no guarantee 😄

saper: We should with the same settings as node does. Maybe they are inherited from the node gyp files, we should check this. Long time since I've had look into this...

The c++0x workaround was for the old gcc on MacOS X I think...

xzyfer: I'm fine with bumping to 6 :)

nschonni:
@xzyfer @saper I added the dropping on Node 10, and bumped it to 6.0.
Do you want to still work on the compiler stuff, or try the trimmed down one and maybe add back if we get reports coming in?

xzyfer:
If we have 16 builds passing I'm happy enough to move forward and deal with
any issues as they come up.

xzyfer:
Because we haven't changed the binding code I typically copy over the
binaries from the previous releases, so I'd only be rebuilding the 16
binaries for release.

nschonni:
Alpine 16 seems to have some flakey CLI tests, but I think we say that in 15 for awhile too

nschonni:
Feel free to land this whenever you have time to cut the release. I bumped the package.json this time 😄

xzyfer:
My hero!

xzyfer:
Released v6.0.0 with Node 16 support.

stevecondylios:
Hi there. Thanks for the contributors' support on a tough issue under pressure. 

sass/node noobie here. Just wondering, it [this](https://stackoverflow.com/q/67241196/5783745) problem history now? i.e. is it okay to update to node 16? 

From what I read above, I assume it might be okay to go to node 16 now, (although I do see 1 failed check for build 16 (I'm not sure if that affects things): https://github.com/sass/node-sass/runs/2483549702)

I wrote [this](https://stackoverflow.com/a/67242989/5783745) quick suggestion for anyone not sure what to do, but if it's now all good to go to node 16, I will update that answer to avoid node 14 and go to 16 now :)

gotjoshua:
@stevecondylios 
> okay to update to node 16

not on mac osx its not.

asharafshahi:
How about now?  If I install the latest will it work with node 16?

PR #3091, opened 2021-04-21
[DNM] Bump C++ standard to C++14

saper:


binding.gyp
```diff
@@ -28,7 +28,7 @@
         }
       },
       'xcode_settings': {
-        'CLANG_CXX_LANGUAGE_STANDARD': 'c++11',
+        'CLANG_CXX_LANGUAGE_STANDARD': 'c++14',
```
nschonni, line 31: Should/do we need to condition this so it only applies to Node 16?

saper: Good question. They obviously build, but we should check if they work on old platforms.

nschonni: I think the other patch that @xzyfer pointed to in the other thread just dropped the line entirely. I wonder if that makes sense either.
I'm good with going with the 14 if it makes sense, since I think you've got more c++ experience than me 😄

xzyfer: I'd rather not change the things for previous versions because we know the resulting binaries are problem free and I have the rocker containers to build them ready

MaxWeisen: Our team is using node-sass v4.12.0 and have been unable to compile node-sass.

 We manually changed line 31 to `'CLANG_CXX_LANGUAGE_STANDARD': 'c++14'` and we were able to overcome the 'remove_cv_t' error with xcode.

saper: > We manually changed line 31 to `'CLANG_CXX_LANGUAGE_STANDARD': 'c++14'` and we were able to overcome the 'remove_cv_t' error with xcode.

Which node version were you on?

xzyfer: Looks like there is canonical answer for which cxx which should be targeting for each Node version at https://github.com/nodejs/node/blob/master/BUILDING.md#official-binary-platforms-and-toolchains

Lots of the decisions in our gyp configuration were made back in the node 0.11  / 0.12 days and cludged on since then to just make it work from them on.

MaxWeisen: > > We manually changed line 31 to `'CLANG_CXX_LANGUAGE_STANDARD': 'c++14'` and we were able to overcome the 'remove_cv_t' error with xcode.
> 
> Which node version were you on?

We are on Node v14.15.1. 

After reading the compatible version table on the readme, we deleted node_modules folder and package-lock.json, upgraded to node-sass 4.14 and still encountered the same error and fixed it with the same solution.

dcunited001:
if i simply pass `CLANG_CXX_LANGUAGE_STANDARD=c++14` to the build process is this sufficient? how do i do that?

polarbirke:
Is there still movement on this? The need for Node v16 compatibility is becoming more important now that it is the active LTS; Github Actions have updated a few days ago (actions/virtual-environments#4446), for example.

polarbirke:
> Is there still movement on this?

Never mind, https://github.com/sass/node-sass/releases/tag/v6.0.0 seems to have solved this.

PR #3096, opened 2021-04-25
fix(docs): corrects common typos in project documentation

plan-do-break-fix:
Scope of changes is restricted to markdown docs only. Changes are made in agreement with Standard American English.


nschonni:
Thanks, but all the changes in the `libsass` folder an vendor'd in so changes need to be made on https://github.com/sass/libsass

plan-do-break-fix:
Apologies I didn't catch that. Thanks for the information.

nschonni:
No problem, it's not obvious, since it's no longer a plain submodule

PR #3121, opened 2021-05-31
chore: Use default Apline version from docker-node

nschonni:
Matching the current defaults from https://github.com/nodejs/docker-node/blob/main/versions.json
A bunch of the current defaults are EOL versions of Alpine, so the tags are out of date

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -22,13 +22,13 @@ jobs:
 
         include:
           - node: 12
-            alpine: "3.9"
-          - node: 14
-            alpine: "3.10"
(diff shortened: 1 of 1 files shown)


xzyfer (approved):
Lgtm. Can we get by without even specifying the minor?

nschonni:
Yes, there are tags for just plain `16-alpine` that could be used. I'm not sure why I chose the previous ones before, but I was thinking it was probably the minimum supported version of Alpine for each Node version.
I'm fine with dropping the version, if this is just aligning it with the default version anyway.

xzyfer (approved):

PR #3129, opened 2021-06-11
Update README.md

hervegbedji:
Hello here! Just added exemple for usage of node-sass as a script in package.json


xzyfer:
Thank you but this is better suited for stackoverflow.

Issue #3106, opened 2021-05-12
Node 16 support with 5.x


maxburke:
Hi,

I noticed that Node 16 was released as version 6.0. Unfortunately a new major version seems to break all the other dependencies we have that require node-sass 4.x or 5.x. 

Is it possible to have this backported to be a patch on top of 5.x? Or released also as 5.1?

nschonni:
https://github.com/sass/node-sass#node-version-support-policy

maxburke:
@nschonni I see that but unfortunately that doesn't help address the knock-on ecosystem breakage brought on by a new major version! 

The original node-sass + Node 16 issue mentions a 5.1 version as well: https://github.com/sass/node-sass/issues/3077

maxburke:
For what it's worth the reason I am asking because node 15 isn't available through normal channels anymore with the release of node-16.1, and node 14 doesn't work on M1 Macbooks so downgrading to that version isn't possible.

cc @xzyfer who closed a similar issue.

xzyfer:
Odd numbered node versions are short live releases. You'll have better luck
with node 14 which is an LTS release.

maxburke:
@xzyfer But node 14 doesn't work with apple's m1 processor; node 15 was the first version to support it.

xzyfer:
That's an interesting call out. Looks like there are some ways to get node < 15 working via Rosetta bit there are quirks. I can take a look at what our options are here. 

Out of curiosity are you able to confrim that node-sass did previously work on Apple M1 with Node 15? It was our impression that node-sass wouldn't work on Apple M1.

maxburke:
@xzyfer Yup, it worked fine until I did a `brew upgrade` yesterday and had my node-15 replaced with node-16.

xzyfer:
Out of curiosity, are there any community packages that are being particularly problematic? I know gulp-sass doesn't support node-sass 6 yet for example.

maxburke:
@xzyfer The issue is that sass-loader 10 requires node-sass 4 or 5, and react-scripts/create-react-app right now depends on sass-loader 10. There's an issue open for react-scripts to support sass-loader but it's open (https://github.com/facebook/create-react-app/issues/10944) 

So the resolution I'm looking at in order to get our react app working for local development is either to eject the react webpack config and tweak versions myself, or build node-15 locally from source.

xzyfer:
Thanks. This isn't likely to happen for a couple days due to my schedule.
[...]

PR #3136, opened 2021-06-24
Attempt to backport Node 16 support to v5

xzyfer:
I'm not convinced this is going to be possible, or maintainable moving forward. 
This will be a one release until the ecosystem catches up.

See https://github.com/sass/node-sass/issues/3106


xzyfer:
Unfortunately this has proven to difficult to achieve in a sustainable way due to some of our dependencies dropping support for older Node versions whilst patching security issues.

PR #3149, opened 2021-07-16
Set rejectUnauthorized to true by default

scott-ut:
Resolve CVE-2020-240-25 by setting rejectUnauthorized to true by default.

Add configuration flag to override this to false if necessary.

Add doc option to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -595,12 +595,13 @@ When compiling a directory `--source-map` can either be a boolean value or a dir
 
 node-sass supports different configuration parameters to change settings related to the sass binary such as binary name, binary path or alternative download path. Following parameters are supported by node-sass:
 
-Variable name    | .npmrc parameter | Process argument   | Value
------------------|------------------|--------------------|------
-SASS_BINARY_NAME | sass_binary_name | --sass-binary-name | path
-SASS_BINARY_SITE | sass_binary_site | --sass-binary-site | URL
-SASS_BINARY_PATH | sass_binary_path | --sass-binary-path | path
-SASS_BINARY_DIR  | sass_binary_dir  | --sass-binary-dir  | path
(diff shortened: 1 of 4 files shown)


scott-ut:
This would resolve https://github.com/sass/node-sass/issues/3067

xzyfer:
Thanks! I've added the `Release - Next Major` flag since this is a breaking change.

scott-ut:
Added other options for changing the configuration setting and updated README. Have removed WIP tag from PR. A couple of build steps failed but I couldn't see why - is that expected @xzyfer ?

xzyfer:
Yes some of the Alpine build jobs fail. It's expected.

scott-ut:
@xzyfer - any update on when this will be merged?

JGJP:
@xzyfer
If this is a security fix then can we have it in a patch or minor update on 6.x.x? Upgrading to node-sass 7 is causing trouble for my peer dependencies that is not trivial to resolve. I suspect that other users will also avoid the upgrade.

xzyfer:
We put this into in v7 because it was technically a breaking change.

JGJP:
@xzyfer what part of it is breaking? I see the addition of a flag, but no change in the existing flags

xzyfer:
IIRC in order to address the security concern the default value for the
flag was reversed.

JGJP:
feels like a minor change to me, it doesn't break functionality right? None of the current binaries should be requiring that flag

JGJP:
@xzyfer Would appreciate a deeper look at the feasability of a minor update, I'm definitely not the only one that can't upgrade to 7 easily, and I'll probably ignore this update or patch it manually until I can do a broader tooling upgrade

PR #3041, opened 2020-12-11
chore: Windows x86 on GitHub Actions

nschonni:
setup-node 2.1.3 added support for x86 architecture so we can potentially drop AppVeyor

PR #3172, opened 2021-08-23
Remove Appveyor in favour of GitHub actions

xzyfer:
Before going too far down this rabbit hole, @nschonni do you think this makes sense?


nschonni:
I haven't looked in a bit, but Actions didn't have a way of using x86 Node.js before. That's the main reason it's still in the matrix. I think support landed for Azure DevOps, but not sure if Actions caught up yet

xzyfer:
Thanks, that's the context I need  :)

nschonni:
Looks like I started https://github.com/sass/node-sass/pull/3041, but didn't wrap it up. Not sure why anymore

xzyfer:
Might have just been blocked on me missing the review request. Looks like these Windows x86 builds ran which is encouraging.

nschonni:
OK, I'll close this in favour of #3041

PR #3193, opened 2021-10-13
Use make-fetch-happen instead of request

CamilleDrapier:
cf: https://github.com/sass/node-sass/issues/2851

Follows the implementation that was done here: https://github.com/sass/node-sass/pull/2961 ; also remove `npmlog` that seems unused from here on.

Because the latest node-gyp does not support python2 (https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#810-2021-05-28), use python3 for all nodes builds/tests. (Not sure if `src/libsass/script/ci-report-coverage:8` should be affected?)

### DIFF
--- a/test/downloadoptions.js
+++ b/test/downloadoptions.js
@@ -8,12 +8,10 @@ describe('util', function() {
     describe('without a proxy', function() {
       it('should look as we expect', function() {
         var expected = {
-          rejectUnauthorized: false,
           timeout: 60000,
           headers: {
             'User-Agent': ua(),
           },
-          encoding: null,
         };
 
         assert.deepStrictEqual(opts(), expected);
@@ -33,13 +31,11 @@ describe('util', function() {
 
       it('should look as we expect', function() {
         var expected = {
-          rejectUnauthorized: false,
           proxy: proxy,
           timeout: 60000,
           headers: {
             'User-Agent': ua(),
           },
-          encoding: null,
         };
 
         assert.deepStrictEqual(opts(), expected);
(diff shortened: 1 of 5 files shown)


CamilleDrapier:
Sorry for taking your time with this, I was merely trying to see if this would be reasonable enough.. while doing the changes locally it seemed okayish~~ but after creating the PR, I thought it might involve too many things that I did not understand enough.

Let's try to only replace the request/fetch, I missed seeing that the [strictSSL/rejectUnauthorized](https://github.com/zkat/make-fetch-happen#--optsca-optscert-optskey-optsstrictssl) option was supported by `make-fetch-happen`, but this rename probably impacts the PR you mentioned :bow:

test/downloadoptions.js
```diff
@@ -59,31 +57,11 @@ describe('util', function() {
...
-            'User-Agent': ua(),
-          },
-          encoding: null,
-        };
-
-        assert.deepStrictEqual(opts(), expected);
-      });
-    });
-
-    describe('with SASS_REJECT_UNAUTHORIZED set to false', function() {
```
xzyfer, line 74: If we restore this test I think we're ok to proceed.

xzyfer:
I've rebase this on the latest master

xzyfer (approved):
This seems reasonable to me. If @nschonni has no reservations I'm happy to roll give it a shot.

nschonni (approved):
I'm sure will get some weird edge cases not matter what library we use, so I'm OK with giving this a try 😄

xzyfer:
Thanks for your work @CamilleDrapier

xzyfer:
This was accidentally released in v7.0.2 and subsequently broke binary downloads. See #3297

CamilleDrapier:
Sorry about this! Not really sure what is the underlying problem, but please let me know if I can help :bow:

xzyfer:
No worries @CamilleDrapier, it was my mistake. My comment was intended for posterity. I've opened https://github.com/sass/node-sass/issues/3300 to track fixing it before releasing 8.0.0.

PR #3198, opened 2021-11-03
remove lodash dep

jimmywarting:
I don't think using just one lodash fn don't justify depending on it.


XhmikosR:
Hmm, perhaps it's not needed in https://github.com/dlmanning/gulp-sass either. I switched from lodash to lodash.clonedeep, but maybe it's not needed there either.

nschonni (approved):

jimmywarting:
@XhmikosR

> use of `lodash.clonedeep` and similar packages is discouraged and they will be removed in v5.

ref: https://lodash.com/per-method-packages

maybe a simple `JSON.parse(JSON.stringify(...))` is sufficient? 
[V8 aggressively optimized](https://v8.dev/blog/cost-of-javascript-2019#json) this pattern

ref: https://web.dev/structured-clone

XhmikosR:
Thanks for the reply.

I'm aware that the individual packages are sort of deprecated, but in this case I don't think it matters.

Either way, if you are up to it, happy to see one less dependency in gulp-sass too. It's already improved with the lodash -> lodash.clonedeep switch on main branch.

xzyfer:
Thanks but `Object.assign` is not equivalent to `clonedeep` and I'm not confident we have adequate test coverage to assert this is a safe behaviour change.

XhmikosR:
This needs to be rebased @jimmywarting but any chance you land this after this rebase @xzyfer?

Issue #3190, opened 2021-10-07
Vulnerability in node-sass > sass-graph > yargs > cliui > strip-ansi > ansi-regex


alexarsh:
- NPM version (`npm -v`): 6.14.15
- Node version (`node -v`): v14.18.0
- Node Process (`node -p process.versions`): 

{
  node: '14.18.0',
  v8: '8.4.371.23-node.84',
  uv: '1.42.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.17.2',
  modules: '83',
  nghttp2: '1.42.0',
  napi: '8',
  llhttp: '2.1.3',
  openssl: '1.1.1l',
  cldr: '39.0',
  icu: '69.1',
  tz: '2021a',
  unicode: '13.0'
}

- Node Platform (`node -p process.platform`): darwin
- Node architecture (`node -p process.arch`): x64
- node-sass version (`node -p "require('node-sass').info"`):

node-sass	6.0.1	(Wrapper)	[JavaScript]
libsass  	3.5.5	(Sass Compiler)	[C/C++]

- npm node-sass versions (`npm ls node-sass`): node-sass@6.0.1

There is the following dependencies tree:

─┬ node-sass@6.0.1
│ └─┬ sass-graph@2.2.5
│   └─┬ yargs@13.3.2
│     ├─┬ cliui@5.0.0
│     │ ├─┬ strip-ansi@5.2.0
│     │ │ └── ansi-regex@4.1.0

When ansi-regex@4.1.0 have the following vulnerability issues: 
https://snyk.io/vuln/npm:ansi-regex@4.1.0

Is there a chance that sass-graph@2.2.5 dependency can be updated in order to fix the issue?

matt-cote:
https://github.com/sass/node-sass/pull/3044

katannshaw:
@alexarsh I had the same issue today. After a lot of trial and error I finally noticed this error message:

```
npm WARN EBADENGINE Unsupported engine { package: 'node-sass@6.0.1',
npm WARN EBADENGINE   required: { node: '>=12' },
npm WARN EBADENGINE   current: { node: 'v10.22.1', npm: '7.21.1' } }
```

So I ran `nvm install v14.18.1` to update my version of node to 14 and it fixed the issue for me. Good luck to you.

driskell:
[...]

PR #3202, opened 2021-11-10
Absorb sass-graph 2.2.5 into node-sass

kiskoza:
Hi.

The current version of node-sass depends on [sass-graph v2.2.5](https://github.com/xzyfer/sass-graph/tree/v2.2.5). Unfortunately that package has a known vulnerability reported by various audit tools. I already tried to open a PR to fix it (see https://github.com/xzyfer/sass-graph/pull/117), but apart from an initial comment stating that the package is in maintenance mode, I haven't got too much response. There's an open issue for this in node-sass' repo too: https://github.com/sass/node-sass/issues/3190

Some of the comments are saying you should change node-sass to dart-sass, but other comments are saying that they have a performance issue with dart-sass, so it's not an option for them. [One of the comments suggested](https://github.com/sass/node-sass/issues/3190#issuecomment-962101527) we should merge sass-graph into node-sass and fix the problem in this repo. I tried to follow this path.

If you check the report closely, you can see that the reported vulnerability is related to sass-graph's command line binary, which was never used in node-sass. That means it should be enough to copy over sass-graph without the CLI tool.

I had to make these changes:
* `lib/sass-graph/*.js` - copy over sass-graph's core
* `lib/watcher.js` - include the new file instead of a package
* `package.json` - remove sass-graph and include `scss-tokenizer`. this one was the only relevant diff between sass-graph's dependencies and the packages already required for node-sass
* `test/sass-graph/*.*` - copy over sass-graph's test. I changed `require('chai').assert` to `require('assert').strict` and fixed the path to find the right sass-graph files

I hope this PR will help expanding this project's lifetime and reduce the cost of maintenance in the future. Is there anything else I need to do?


xzyfer:
Thanks but we maintain both packages so we're happy to keep them separate.

PR #3366, opened 2023-06-02
feat: drop Node 19

nschonni:
Node 19 hit EOL today (2023-06-01) so cleaning out the CI and support for the next release


nschonni:
Yeah, I think it's even OK to keep coping the bindings to the new releases after this, but it's more of marking that we won't look at any issues raised now that it's EOL

